PR #11, opened 2017-01-12
Implement Routing

cpx86:
Implementation of router actor + broadcast, random, consistent-hash and round-robin router configs.

Some thoughts about the current code:
- A lot is basically directly copied from the Go implementation. Some things could be refactored into more idiomatic C#.
- I couldn't find any hash ring implementation for .NET, so I wrote a minimalist implementation. Looks correct to me, but I may well have missed something :)
- - I haven't made any changes to the structure of the project, so Routing.cs has grown by something like 400 lines. The code should perhaps be restructured into a more idiomatic C# style, using more granular namespaces, smaller files, etc, but that would be best to do in a separate PR. It might also be a good time to start introducing some unit tests, to make refactoring safe.

### DIFF
--- a/ProtoActor.sln
+++ b/ProtoActor.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26014.0
+VisualStudioVersion = 15.0.26020.0
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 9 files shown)


rogeralsing:
Looks good!
We are currently restructuring a lot of the router code in Go, but this can be replicated later for C#
I'll merge and dive into the details.
Many thanks!

cpx86:
Yeah, just after I pushed submit I noticed there had just been a bunch of restructuring of the Go code :) But yes, a good opportunity to bring them both in line!

PR #12, opened 2017-01-16
Implemented BoundedMailbox and UnboundedMailbox

### LINKED ISSUE #5: Port BoundedMailbox to C#
The bounded mailbox has some nice properties in terms of low GC overhead.
It would be good to have the same for C# too.

cpx86:
BoundedMailbox uses a lock-free, bounded MPMC queue, shamelessly copied from https://github.com/alexandrnikitin/MPMCQueue.NET and slightly tweaked so Enqueue/Dequeue will keep retrying until they succeed (this is how the ring buffer in protoactor-go works)

UnboundedMailbox uses the .NET ConcurrentQueue that was in place before.

Added a simple benchmark for comparing the two - it spawns an actor, sends 10M messages to it, and measures how many long it takes to send all messages and receive all messages. Nothing fancy but it does seem to reveal some difference in memory allocation between the two choices.

(Fixes #5)


rogeralsing:
Awesome!
There is an issue in .NET core to use Server mode GC, so performance will be a lot worse than it could be.
Once that has been addressed it will be easier to see the performance differences.
https://github.com/dotnet/coreclr/issues/4118

PR #13, opened 2017-01-16
Fix Props.Copy implementation

cpx86:
The usage of default parameters in Props.Copy does not quite work. The previous implementation basically did `_foo = foo ?? this._foo` which obviously won't work if you want to set the value to `null` (as in the case of RouterConfig - although perhaps the choice to use null as an indicator of a non-router Actor could be reconsidered).

This PR reimplements the Copy method to make a full copy first, then apply a mutator Action in order to change the desired property to it's new value.


rogeralsing:
Good catch :-)

PR #16, opened 2017-01-17
Receive pipeline

cpx86:
Not a lot of important stuff in this PR. Main change was to properly hook up the receivers in Props to the Context. Also removed the duplicate Receive delegate


rogeralsing:
Looks good!

PR #17, opened 2017-01-19
Get structure in line with Go

cpx86:
Most of the actual code is the same as before. Changes:
- Moved all routing code to separate project
- Removed router code from Props, replaced with Spawner abstraction
- Renamed Routing to Router
- Renamed Remoting to Remote


rogeralsing:
Thanks!

PR #29, opened 2017-01-22
Use bounded queues for EndpointWriterMailbox.

### LINKED ISSUE #23: EndpointWriterMailbox should use a bounded mailbox for user messages
EndpointWriterMailbox should use a big bounded mailbox for user messages.
This is what we do in Go, in C# we still use a concurrent queue here.

As there will not be a lot of endpoint writers, it's ok to have a bigger queue.

cpx86:
Fixes #23


cpx86:
Closing this for now, because of problems with the BoundedMailboxQueue.

PR #30, opened 2017-01-24
Implemented mailbox stats

cpx86:
Also reverted back to unbounded queues.

### DIFF
--- a/src/Proto.Actor/IMailbox.cs
+++ b/src/Proto.Actor/IMailbox.cs
@@ -19,8 +19,9 @@ namespace Proto
     public interface IMailbox
     {
         void PostUserMessage(object msg);
-        void PostSystemMessage(SystemMessage sys);
+        void PostSystemMessage(SystemMessage msg);
         void RegisterHandlers(IMessageInvoker invoker, IDispatcher dispatcher);
+        void Start();
     }
 
     public interface IMailboxQueue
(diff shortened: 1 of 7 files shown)


src/Proto.Actor/IMailbox.cs
```diff
@@ -77,27 +78,31 @@ public class DefaultMailbox : IMailbox
...
 
         private int _status = MailboxStatus.Idle;
         private bool _suspended;
 
-        public DefaultMailbox(IMailboxQueue systemMessages, IMailboxQueue userMailbox)
+        public DefaultMailbox(IMailboxQueue systemMessages, IMailboxQueue userMailbox, params IMailboxStatistics[] stats)
         {
             _systemMessages = systemMessages;
             _userMailbox = userMailbox;
+            _stats = stats;
```
rogeralsing: Either we should have something like this:
`_stats= stats ?? Array.Empty<IMailboxStatistics>()`
Or, we should default empty array to null.

Im not sure how foreach behaves over empty arr, if that still results in some compiler magic and a null check might be faster (?)
Optimize for default path.

Thoughts?

cpx86: foreach doesn't get much magic compiler love AFAIK. I think ensuring it is always non-null and readonly might be good - the JITter should be able to get rid of the internal null dereference check, right? And switching to a for-loop instead of a foreach-loop is always faster. Will update.

rogeralsing:
👍

PR #31, opened 2017-01-25
Ported receive timeout

cpx86:


src/Proto.Actor/Context.cs
```diff
@@ -212,6 +218,23 @@ public void Unwatch(PID who)
...
+            var influenceTimeout = true;
+            if (ReceiveTimeout > TimeSpan.Zero)
+            {
+                var notInfluenceTimeout = msg is INotInfluenceReceiveTimeout;
+                influenceTimeout = !notInfluenceTimeout;
+                if (influenceTimeout)
+                    StopReceiveTimeout();
+            }
+
+            await ProcessMessageAsync(msg);
```
cpx86: There's a bit of work to be done to get the message invocation in line with Go. Parts of it are at least extracted to a ProcessMessageAsync method, but there are some major differences left, e.g. the receive pipeline.

rogeralsing:
This looks great, we do need to stop the receive timeout inside `Stopped()` and inside `Restart` so that the receive timer is not running for a dead or restarted actor.

I saw that this is missing in Go too.

cpx86:
Done. It was already done in Go, but in the `tryRestartOrTerminate` function: https://github.com/AsynkronIT/protoactor-go/blob/dev/actor/local_context.go#L259

rogeralsing:
:+1:

PR #35, opened 2017-01-29
Separate mailbox project

cpx86:
Just an initial stab at it, splitting off the mailbox, queues and dispatcher into a separate lib. If we want to make this into something more generic, we should think about how we want the API surface of Proto.Mailbox to look. E.g. currently we have a concept of user vs. system messages which makes sense in an actor context, but might not in other contexts.


rogeralsing:
This needs rebasing

cpx86:
Yep. Will fix the conflicts asap.

cpx86:
There we go :)

rogeralsing:
whoop whoop

PR #49, opened 2017-02-12
Added some unit tests and added PID.RequestAsync timeout feature in the batch

jrouaix:


src/Proto.Actor/Futures.cs
```diff
@@ -1,33 +1,39 @@
...
+        {
+            _tcs = tcs;
+            _cancellationToken = cancellationToken;
+        }
+
+        public override void SendUserMessage(PID pid, object message, PID sender)
+        {
+            if (message is T)
+            {
+                if (_cancellationToken.IsCancellationRequested) return;
```
cpx86: The PID still needs to be stopped in this case.

tests/Proto.Actor.Tests/ActorTests.cs
```diff
@@ -0,0 +1,97 @@
...
+            var messages = new Queue<object>();
+
+            var pid = SpawnActorFromFunc(ctx =>
+            {
+                messages.Enqueue(ctx.Message);
+                are.Set();
+                return Actor.Done;
+            });
+
+            pid.Request("hello", null);
```
cpx86: Small thing, but I think you could use `Tell` here instead since you're not passing a sender.

tests/Proto.Actor.Tests/ActorTests.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            var reply = await pid.RequestAsync<object>("hello", TimeSpan.FromMilliseconds(100));
+
+            Assert.Equal("hey", reply);
+        }
+
+        [Fact]
+        public void ActorLifeCycle()
+        {
+            var are = new AutoResetEvent(false);
```
cpx86: Is this used to ensure messages are processed before the test continues? In that case you could use the TestMailbox instead - it processes messages synchronously.

src/Proto.Actor/Futures.cs
jrouaix: If cancellation is requested, Stop() is already called by the RequestAsync timeout continuation Task.
Could we call it twice ?
By the way, i'm not a big fan of this intricated code I made : Once stop() is called by the upper caller, once it's called by the FutureProcess itself, if you think of better design : what I need is a way to timeout IN the future.

Perhaps It's time to code an AlarmClock actor somewhere to send DelayMessages to any actor, with this the code of this timeout story could be nicer.

tests/Proto.Actor.Tests/ActorTests.cs
jrouaix: Ok, i'll give a try :-)

tests/Proto.Actor.Tests/ActorTests.cs
jrouaix: yep, my bad

src/Proto.Actor/Futures.cs
cpx86: How about if we move the CancellationTokenSource and the logic related to it into the FutureProcess. Then we don't clutter the PID with all of the timeout/cancellation logic. The only difference in PID would be that we optionally pass a CancellationTokenSource to the FutureProcess, and we would return FutureProcess.Task instead of TaskCompletionSource.Task. Does that make sense? :)

jrouaix: Well the timeout task has to know the PID value in order to stop() it, and currently the future process knows only about it when SendUserMessage is called ...
Could the FutureProcess do the PID/ProcessRegistry registration itself in its ctor ?

jrouaix: Tried it this way, working.

src/Proto.Actor/PID.cs
```diff
@@ -40,44 +40,18 @@ public void Request(object message, PID sender)
...
-            var name = ProcessRegistry.Instance.NextId();
-            var (pid, absent) = ProcessRegistry.Instance.TryAdd(name, reff);
-            if (!absent)
-            {
-                throw new ProcessNameExistException(name);
-            }
-            Request(message, pid);
-
-
-            if (cts == null) return tsc.Task;
```
cpx86: I think it would be better to keep everything above this line in PID, and only move the Task mojo into FutureProcess. Like this:
```csharp
var tsc = new TaskCompletionSource<T>();
var reff = (cts == null) ? new FutureProcess<T>(tsc) : new FutureProcess<T>(tsc, cts);		
var name = ProcessRegistry.Instance.NextId();		
var (pid, absent) = ProcessRegistry.Instance.TryAdd(name, reff);		
if (!absent)		
{		
    throw new ProcessNameExistException(name);		
}		
Request(message, pid);
return reff.Task;
```

jrouaix: The very problem here is the CancellationTask need the PID, and the PID is known only after the FuturProcess instance is created. As I try not to implement a sequential coupling anti pattern I'd rather not call a SetPID() method on the FuturProcess after registered it.

For the TaskCompletionSource I tought we could move the creation of this instance in the FutureProcess ctor as we probably won't create an other class inherited from TaskCompletionSource or send null to the constructor so ... could be simpler. But perhaps you have more to come on FuturProcess that will need to send different kind of "ReceivedResponseAction".

So tell me ?
- SetPID ?
- Something else ? (can a process instance can retrieve the associated PID after the registration ?)

jrouaix:
Thanks, to be continued ...

PR #51, opened 2017-02-13
Do not use await in Mailbox

cpx86:
Most of this has been reviewed already, but found some additional things that were missing.

So to summarize:
- If a task is complete, invoke message statistics and continue
- If a task is faulted, escalate the failure and continue.
- If a task is non-complete, keep the mailbox in busy state but do not continue. When the task completes, handle any fault, invoke message statistics, and resume the processing.

Also added a bunch of unit tests for the mailbox :)


rogeralsing:
Looks great. but we need to drop the goto part in order to free up some execution time for other actors. see inline comment.

cpx86:
I don't see any inline comment about that :)

rogeralsing:
These bits:

There should be a await Task.Yield(); before the goto.
otherwise the mailbox will hog the current thread until it is empty.
awaiting yield will do the same as runtime.Schedule in Go. it will drop out of the flow and allow other tasks to get some time to execute
---
And when I wrote this, I realized this will be hard in a no async mailbox :P
---
it might be a better idea to drop the goto and just call schedule() instead, that will have the desired effect.

cpx86:
Done. One thing I started thinking while adding some comments - we really have a third state of the mailbox now "halted". Maybe that can be made more explicit by adding it to `MailboxStatus`? So instead of returning `false` from `ProcessMessages` we'd set status to `MailboxStatus.Halted`. When the task completes we set it back to `Busy` - and we would only set it to `Idle` if it was previously `Busy`.

rogeralsing:
Merging, this yields ~10% perf so its all good :-)

PR #53, opened 2017-02-13
Refactor pid process ref

### LINKED ISSUE #52: Assign PID.Process the same way Go does it
The current implementation of assigning the cached Process in the PID, is in the ProcessRegistry.
https://github.com/AsynkronIT/protoactor-dotnet/blob/master/src/Proto.Actor/ProcessRegistry.cs#L43

This is not how it works in Go.
The Go impl, will check if the cached process is nil, and if so, it will resolve the process and assign to the cache field.
https://github.com/AsynkronIT/protoactor-go/blob/dev/actor/pid.go#L11

The benefit of doing this is that when a process dies, the cache can be invalidated for each PID instance pointing to that process.

>Why can there be multiple PID instances pointing to the same process?

The PIDs might come from deserialized messages, and thus they will all contain the same values, but be different instances.

cpx86:
Fixes #52


rogeralsing:
Looks good, just the few interlocked reads/writes that needs fixing :)

cpx86:
Done

rogeralsing:
👍

PR #59, opened 2017-02-16
Persistence fixes

cpx86:
The replay command is sent as a user message when the actor is started, but since we don't block on the started message, the replay command is not guaranteed to be the first in the user mailbox. In the persistence example, it always ends up at the end of the user mailbox, meaning the Persistence property is not initialized when the other messages are processed.


cpx86:
Also made all Persistence methods async.

PR #60, opened 2017-02-18
Support port 0 in remoting

### LINKED ISSUE #48: Make port 0 mean any free port
Currently, when using port 0, gRPC will use a free port as expected. but, the system does not know what port this is and thus PID lookups fail.
We need to either extract the port from gRPC, or find a free port if port 0 is used.

cpx86:
Fixes #48


rogeralsing:
👍 Can you get rid of the random port assigment in this example too? 
https://github.com/AsynkronIT/protoactor-dotnet/blob/master/examples/Chat/Client/Program.cs#L11

cpx86:
Sure thing!

rogeralsing:
Ty :-)

PR #62, opened 2017-02-22
Added supervision methods to ISupervisor

### LINKED ISSUE #55: Port Go Supervisor
There is a new interface in Go:
https://github.com/AsynkronIT/protoactor-go/blob/dev/actor/supervision.go#L14

The implementation is here:
https://github.com/AsynkronIT/protoactor-go/blob/dev/actor/local_context.go#L395
and here for the root supervisor
https://github.com/AsynkronIT/protoactor-go/blob/dev/actor/root_supervisor.go

This interface is now used in the SupervisionStrategy, like so:
https://github.com/AsynkronIT/protoactor-go/blob/dev/actor/strategy_one_for_one.go#L23

This means that the "Supervisor" is responsible for how to Stop, Restart and Escalate failures.
It's mostly an abstraction to not have to deal with system messages inside the supervisor strategy.
But it can also make testing SupervisorStrategies easier.

cpx86:
Fixes #55 

Will add tests too, but ended up refactoring a bunch of stuff, so will send separate PR for that.

### DIFF
--- a/src/Proto.Actor/LocalContext.cs
+++ b/src/Proto.Actor/LocalContext.cs
@@ -267,6 +267,30 @@ namespace Proto
             Parent.SendSystemMessage(new Failure(who, reason, _restartStatistics));
         }
 
+        public void RestartChildren(params PID[] pids)
+        {
+            for (int i = 0; i < pids.Length; i++)
+            {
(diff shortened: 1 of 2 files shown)


rogeralsing:
👍

PR #63, opened 2017-02-24
change wait() to await.

alexpantyukhin:


src/Proto.Persistence/Persistence.cs
```diff
@@ -28,7 +28,7 @@ public void Init(IProvider provider, IContext context)
             if (t != null)
             {
                 EventIndex = t.Item2;
-                Context.ReceiveAsync(t.Item1).Wait();
+                await Context.ReceiveAsync(t.Item1);
```
rogeralsing: The method is a void method so that would not compile.
Are you missing any additional changes in the commit?

alexpantyukhin: Interesting.. there is the same: https://github.com/AsynkronIT/protoactor-dotnet/blob/master/src/Proto.Persistence/Persistence.cs#L48

rogeralsing: No that one is "async Task PersistReceiveAsync" signature, this one is "void Init"

alexpantyukhin: Hm.. but I see 'async Task InitAsync' https://github.com/AsynkronIT/protoactor-dotnet/blob/master/src/Proto.Persistence/Persistence.cs#L21

rogeralsing: hmm, this is what is in the PR:

change

Oh well, Ill merge, if it turns out strange we can always fix it :-)

alexpantyukhin: hm... don't know how is it possible) and it doesn't show that signature was changed.

PR #65, opened 2017-02-26
Tests, mailbox fixes, etc

cpx86:
Quite a lot going on in this PR 😅 To summarize:
- Tests for the OneForOneStrategy
- Re-added exception handling in mailbox, for cases where an actor throws an exception instead of returning a failed task
- Send system messages to mailbox statistics
- Moved SystemMessage into mailbox so Suspend/ResumeMailbox can inherit from it
- Moved all test fixtures into a shared project


tests/Proto.Actor.Tests/SupervisionTests.cs
```diff
@@ -0,0 +1,121 @@
...
+            var childProps = Actor.FromProducer(() => new ChildActor())
+                .WithMailbox(() => UnboundedMailbox.Create(childMailboxStats));
+            var parentProps = Actor.FromProducer(() => new ParentActor(childProps))
+                .WithSupervisor(strategy);
+            var parent = Actor.Spawn(parentProps);
+
+            parent.Tell("hello");
+
+            childMailboxStats.Reset.Wait(1000);
+            Assert.Contains(Restart.Instance, childMailboxStats.Posted);
```
rogeralsing: Maybe we can create some sort of abstraction for this, I have an half-assed attempt for receive handling in Go here:
https://github.com/AsynkronIT/protoactor-go/blob/dev/actor/supervision_test.go#L87

that would allow us to get rid of resetevents in the tests too.
We could have a timeout on the Expect methods

rogeralsing:
I'll merge this, but please see the inline comment on message expect abstraction.
Maybe we could create something similar to Akka testkit, but not tied to any specific test framework, only helper abstractions for the built in primitives.

tests/Proto.Actor.Tests/SupervisionTests.cs
cpx86: Yes, that would be very nice, especially for users if we can make it into a testing package.

PR #69, opened 2017-03-04
Alarm clock > TO CLOSE

jrouaix:
Main feature of this PR is an alarm clock implementation to be able to 
context.SelfDelayMessage(msg, delay); from the actor receive method.

It's implemented as an extentions, I didn't change the main classes, could be in another project...

Also some vscode stuff build & test tasks


rogeralsing:
Regarding message scheduling. 
I would like to see some sort of scheduler interface that we can inject into the actors to allow transient scheduling, persistent scheduling (e.g. quartz) 
Maybe that could also be used in conjunction with the above extensions you provided (?)

jrouaix:
Right, I totally agree about that, indeed I strongly need this kind of persistent scheduling in order to use proto.actor in production. This was just a first tiny step.

I heard about quartz in MassTransit but yet i didn't use it : http://masstransit-project.com/MassTransit/usage/scheduling/ perhaps it's time for me to step in.

Do you plan to have a Proto.Scheduling assembly ? any tought about a base interface to begin with ?
Do we open an issue to discuss about that ?

adamhathcock:
Just butting in: for scheduling, Hangfire would fit I think.  I'd be interested adding that when scheduling is defined.

cpx86:
@rogeralsing Are you thinking something similar to how we inject persistence?

rogeralsing:
@cpx86 IMO it doesn't even need to be a plugin, it could just be vanilla ctor arguments.
we could pass two different impls to the same actor if we want.
e.g. `Ctor(IScheduler transientScheduler, IScheduler persistentScheduler)`

I'd like to see this like any other dependency

src/Proto.Actor/AlarmClock.cs
```diff
@@ -0,0 +1,67 @@
...
+                case DelayMessage delayMsg:
+                    _messages.Add(delayMsg);
+                    context.Self.Tell(_tickMessage);
+                    break;
+
+                case _tickMessage:
+                    if (_messages.Count == 0) break;
+
+                    var now = DateTime.UtcNow;
+                    var msgsToSend = _messages.Where(msg => msg.Timeout < now).ToArray();
```
rogeralsing: If the set is sorted, it should be possible to use `TakeWhile`, no?

jrouaix: yes sure ! my bad

jrouaix:
By the way, I didn't update this PR much but i'd like to improve it by :
- adding a Actor.Scheduling project (ISheduled definition, and in memory implementation recycling what is in this PR)
then adding a Acotr.Scheduling.Hangfire or  Acotr.Scheduling.Quartz project for other implementation

I just have 0 time in those days :-(

cpx86:
@rogeralsing When you say ctor arguments, do you mean on the actor? In that case, how do you jack it up in a nice way? Feels like we're starting to move into DI territory here...

jrouaix:
@cpx86 I think we can just propose some scheduler implementation and let the user bootstrap the factories/implementation he wants.

It is indeed dependency injection because the actor ctor will receive interfaces

But then the DI configuration is done at startup, in factories definitions, so no need any tool to that. Then the final users can use any DI framework if they need/want to do that.

cpx86:
Yeah I don't see any problem with using a DI framework - for most realistic scenarios you would do that and hook up the container's resolve method as the actor producer. Just thinking out loud here :)

So the scheduler interface we provide would essentially just provide a few methods that are tailored to scheduling proto messages? Something like `IScheduler.ScheduleMessage(IContext context, PID pid, object message, DateTime timestamp)`?

jrouaix:
Guys, about this api : 
IScheduler.ScheduleMessage(IContext context, **PID pid,** object message, DateTime timestamp)

I don't see a use case where we should use the target pid. In my opinion the target pid should always be the sender pid because, again in my opinion, only the Actor should maintain it's own scheduling behaviours. And while it's technically the same, I don't think we should provide space for bad practices ...

IScheduler.ScheduleMessage(IContext context, object message, DateTime timestamp)

What you think about ?

cpx86:
One thing I started thinking is if we should have a generic scheduler interface at all. If that should support multiple schedulers we risk ending up with either the least common denominator of those, or an overly complex interface that works partially for each scheduler. So perhaps a better approach would be to directly build e.g. a Quartz addon without having a pluggable interface in between. What do you think?

jrouaix:
In fact I just try to reproduce a simple but very usefull feature here :
- Ability to recall an actor in the time.

In Azure Service Fabric (the one we are currently using and want to get rid off) there is 2 "delay" features
-> timers : in memory callbacks, they don't influence the receive timeout and are not persisted
-> reminders : they are persisted callbacks registered and unregistered by a key identifier

I think this stupid simple interface is enough to reproduce this two use case :
-> timers : using a non persisted scheduling provider, and send SystemMessages (in order not to interfere with Timeout right ?). A periodic timer produced by resending the same message after or even before treated it.
-> reminders : using a persistent sheduling provider (will also need to recreate actors when they have been timedout, i'm supposing here it'll be possible to plug it to the activation system you guys will ship soon)

And for more complex / framework specific bound scenarios, yes you're right, it won't be enough, better for those use case to stick direct to the scheduling framework.

adamhathcock:
You can implement this without an interface, right?  Just as extension methods off of `IContext`. 

I can see the need for a simple delay but I definitely understand the apprehension for declaring an official interface that would change.

A quartz scheduler could also just be extension methods off of the context too, probably.

jrouaix:
@adamhathcock 
Actually an extension method on IContext was my first draft, but it's also true that in order to test the functionnal behavior of the actor, you'll want to mock the scheduler in unit tests and then it's so much nicer to have an interface to inject rather than having to setup globally some messaging system in static classes.

adamhathcock:
I agree about the injection but it often feels like actor systems aren't made with DI in mind :/

cpx86:
Actually I found a way to inject into the pipeline using only extension methods and middleware. Not completely sold on the approach yet myself but it's an option. Will provide a code sample soon to discuss around.

cpx86:
OK so this is the API, a `WithScheduler()` extension on Props and a `Schedule()` extension on IContext.
```csharp
var props = Actor.FromFunc(ctx =>
            {
                var msg = ctx.Message;
                if (msg is Hello r)
                {
                    ctx.Schedule($"Goodbye {r.Who}", TimeSpan.FromSeconds(1));
                    Console.WriteLine($"Hello {r.Who}");
                }
                if (msg is string s)
                {
                    Console.WriteLine(s);
                }
                return Actor.Done;
            })
            .WithScheduler();
        var pid = Actor.Spawn(props);
        pid.Tell(new Hello
        {
            Who = "Christian"
        });
```

The implementation simple uses a custom message type and a sender middleware to intercept the message and delay it:
```csharp
public static class Scheduler
{
    public static Props WithScheduler(this Props props)
    {
        return props.WithSenderMiddleware(
            next => async (context, target, message) =>
            {
                if (message.Message is DelayMessage dm)
                    Task.Run(async () =>
                    {
                        await Task.Delay(dm.Delay);
                        context.Tell(context.Self, dm.Message);
                    });
                else
                    await next(context, target, message);
            });
    }

    public static void Schedule(this IContext context, object message, TimeSpan delay)
    {
        context.Tell(context.Self, new DelayMessage {Delay = delay, Message = message});
    }

    class DelayMessage
    {
        public object Message { get; set; }
        public TimeSpan Delay { get; set; }
    }
}
```

Thoughts? Is this an OK use for middleware? Ping @rogeralsing :)

rogeralsing:
The suggested middleware will be racy.
if we spin off an async operation inside that middleware, the current message headers of the context can change during this time as the actor might start processing another message.

```csharp
                    context.Tell(context.Self, dm.Message); //<-- the actor could have processed 10 new messages when we get here
                });
```

cpx86:
Damn you @rogeralsing ! ;-)

Aside from the potentially racy stuff though - sender middleware seems to be a pretty powerful option for adding these types of extensions, as long as we can provide a good API that doesn't create the risk of race conditions and whatnot.

jrouaix:
Ok, i'll middle wire.

cpx86:
@rogeralsing how can we avoid the race condition though? i can't quite see where it would happen.

jrouaix:
@cpx86 @rogeralsing I'll use the ISheduler interface already proposed in the Sender so the Delay stuff is externalised to other concerns

adamhathcock:
I have found I need a simple in-memory scheduler.  Looking forward to this.

rogeralsing:
I've added some concerns we should resolve first before implementing anything see @97

jrouaix:
Closing the PR, i'll repost something more according to https://github.com/AsynkronIT/protoactor-dotnet/issues/97 & https://github.com/AsynkronIT/protoactor-dotnet/issues/99 when we'll have a conclusion

PR #70, opened 2017-03-05
More expressive methods for supervision

### LINKED ISSUE #56: Port Go RestartStatistics
The RestartStatistics in Go have changed slightly to carry intent rather than just expose state.
https://github.com/AsynkronIT/protoactor-go/blob/dev/actor/child_restart_stats.go
It now has a few methods to show what you can do with the stats.

(It can still be improved, e.g. not exposing mutable fields in the way it does now)

cpx86:
Fixes #56


rogeralsing:
Looks great!

Issue #18, opened 2017-01-20
Pick a logging abstraction


rogeralsing:
My impression is that the https://www.nuget.org/packages/microsoft.extensions.logging.abstractions/ is what will become the de-facto standard for logging in .NET.
After all, this is what is baked into Asp.NET Core.

Thoughts, should we aim for this?

cpx86:
Seems to be the standard for typical logging scenarios - see comment in the aspnet/Logging repo here: https://github.com/aspnet/Logging/issues/332#issuecomment-187836070

Another option that could be interesting is EventSource/ETW logging. E.g. if we want to log diagnostics.

adamhathcock:
I'm interested in taking a stab at this.  How should this be done?  
For internal logging only?
Added to the context like Akka.net? (I guess this is correct?)

I'm used to dependency injection so if something needs logging it just asks for it.

adamhathcock:
Giving this a go and running into chicken-and-egg problems.

I can't just replace all usages of Console with logger because things like `EventStream` are always static.  There's a lot of this.

I want to solve this by changing `Props` to really be `IServiceProvider` and use IoC to create objects.  I'm guessing this is probably an unwelcome change.

adamhathcock:
Any thoughts about using IoC to resolve/create the proto actor primatives to inject logging?  I would use the interfaces and stuff from the Microsoft.Extensions like logging.

I can try my best with a static implementation if that's desired instead.

@rogerasling @cpx86

cpx86:
I'm a bit undecided at the moment, to be honest. I'd like to avoid having any container abstraction or service locator within the library. One question is for which scenarios we need logging. E.g. for general purpose actor logging, it can be injected via the actor producer. To log e.g. all received messages, it can be added as a middleware. If we want to log mailbox events, that can be done with the mailbox statistics extension. If there are more use cases, perhaps we can provide similar extension points for those?

adamhathcock:
Actor logging can be done themselves via the producer.  I was going to optionally allow usage from the context because I figured the library itself would use that.

However I don't know how a logger factory can be shared among the primatives.  Maybe nothing is needed.

rogeralsing:
In Go, we have a special EventStream where we push a LogMessage onto.
That way, anyone can subscribe to that and apply whatever log we want.
I think that would be the cleanest here too.

e.g.
`Proto.Log.Stream.Subscribe( e => { myLogger.Warn(..) })`

[...]

PR #81, opened 2017-03-14
Logging

adamhathcock:
Use a static log object.  Instrument some code and use an example.

I recognize this isn't what is discussed in: https://github.com/AsynkronIT/protoactor-dotnet/issues/18

Essentially, if I try to put an EventStream in I'll have to fan-in from all usages of `ILogger` (or some other interface) to the stream and put data in a custom `LogMessage`.  Then fan-out again to `ILogger` to allow usage of `ILoggerFactory` and all the logging implementations that use that.

Ideally, you want to use `ILogger` or `ILoggerFactory` directly in the code to be instrumented.  The problem was always how to get one of those objects to the code.

An event stream doesn't feel like a fit because the `LoggerFactory` itself is a pub/sub type construct in which all the implementations are subbing to the factory and the code is pubbing.

If this doesn't fit the vision at all, that's cool.  I guess I just don't see how log messages should be involved in an actual event stream.


adamhathcock:
Closing in favor of https://github.com/AsynkronIT/protoactor-dotnet/pull/86

adamhathcock:
@cpx86 reopened this.  

I have a static logger factory so we don't need to use DI.

I was wondering if the log should be moved to Proto.Mailbox or it's own project as Proto.Actor isn't the lowest level library.

cpx86:
I think Proto.Actor is the best fit. The reason Mailbox is separate is because we want to make it possible to use Mailboxes for other purposes than Actors. But Proto.Actor is really the core library. Having a separate Proto.Logging feels a bit overkill since it probably wouldn't be usable outside of Proto.Actor.

adamhathcock:
No problem.  Then I guess this is ready unless the style isn't liked.

PR #84, opened 2017-03-14
Add CurrentSynchronizationContextDispatcher

adamhathcock:
I can make a separate Proto.Extensions or Proto.Mailbox.Extensions if desired.

Or just have my own library if extensions aren't desired.


rogeralsing:
Looks great!

PR #85, opened 2017-03-15
Throw exception when a future gets a message it does not expect.

adamhathcock:
I keep screwing up responses.  This at least tells me when I send a bad message and the future isn't frozen anymore.

I'm sure there's maybe a nicer way to do this.

I wish I could also tell then a future process's message is processed but no reply was sent at all.  This is also a problem I have when developing.

### DIFF
--- a/src/Proto.Actor/Futures.cs
+++ b/src/Proto.Actor/Futures.cs
@@ -59,7 +59,12 @@ namespace Proto
 
                 _tcs.TrySetResult((T)message);
                 pid.Stop();
+            }            
+            else
+            {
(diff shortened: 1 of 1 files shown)


rogeralsing:
Looks good.

> wish I could also tell then a future process's message is processed but no reply was sent at all. This is also a problem I have when developing.

I'm not quite following here, could you expand on this?

adamhathcock:
If I forget to respond to a message the caller is sat waiting for a reply that won't return.  I was hoping to also get an exception after a message is processed but no reply was sent.  Not sure it's possible but wanted to ask.

cpx86:
@adamhathcock If you use RequestAsync, you can set a timeout after which the associated Task will fail. There's isn't really any way that the framework could determine that an actor should have responded.

PR #86, opened 2017-03-16
Use LibLog copy.

adamhathcock:
LibLog is still converting to be a package: https://github.com/damianh/LibLog/pull/125 so I just copy/pasted.

It will compile in anyway.


adamhathcock:
Nevermind, we actually do want to use M.E.Logging https://github.com/AsynkronIT/protoactor-dotnet/pull/81

PR #88, opened 2017-03-17
Implemented remote deployment

rogeralsing:
This PR implements the remote deploy functionality.
The main aspect to review is how to group the Remote API.
Currently we have `RemotingSystem` as the API surface for all actions.
First, the package has been renamed from Remoting to Remote, so the name makes no sense anymore.


cpx86:
Looks nice and clean to me! Two small things:
1. Agree on the naming. We should at least rename it to RemoteSystem to make it more consistent. Not sure about the System suffix either (sounds a bit like actor system, which we don't have) but unless we think of something else that's fine.
2. Would be nice to have an example showing how this works! :)

rogeralsing:
Added RemoteActivate example and renamed RemotingSystem to just Remote.
Potential issue `Proto.Remote.Remote` but it does look OK from the current code.

cpx86:
Awesome!

PR #89, opened 2017-03-17
new persistence api

raskolnikoov:
Rewrite of persistence api and added ravendb persistence plugin. Maybe ravendb/couchebase should be moved outside ofta the core framework into its own repos?


raskolnikoov:
Im not 100% sure I got the versioning right. Byt maybe that is not set in the projects.

rogeralsing:
Can you describe the changes to start a discussion about it?
How is snapshotting handled now?

raskolnikoov:
I have removed the "GetSnapshotInterval" method and property from Persistence. I also removed the bool property "Recovering". I think snapshots should be triggered by the actor itself instead when snapshots are needed.

I've introduced the following messages to be send to the actor on recovery/persisting:

- RecoveryStarted
- RecoveryCompleted
- RecoverSnapshot *contains persisted snapshot object
- RecoverEvent *contains persisted event object
- PersistedEvent *contains the persisted event object

raskolnikoov:
"IProviderState" has been expanded with methods to handle "DeleteEventsAsync" and "DeleteSnapshotsAsync" so the actor could clean up stored events/snapshots that are not needed anymore.

cpx86:
Letting the actor control snapshotting makes sense IMO. It is more related to the needs of the actor than the needs of the persistence.

About versioning, don't worry. It's currently set by the build script and we only bump it after merging to master.

raskolnikoov:
Anyone time ro review my changes?

rogeralsing:
The change looks good to me.
cc @cpx86

src/Proto.Persistence/InMemoryProviderState.cs
```diff
@@ -60,5 +51,15 @@ public Task PersistSnapshotAsync(string actorName, ulong eventIndex, object snap
             _snapshots[actorName] = Tuple.Create((object) snapshot, eventIndex);
             return Task.FromResult(0);
         }
+
+        public Task DeleteEventsAsync(string actorName, ulong fromEventIndex)
+        {
+            throw new NotImplementedException();
```
cpx86: We should just return `Task.FromResult(0)` here.

src/Proto.Persistence/InMemoryProviderState.cs
```diff
@@ -60,5 +51,15 @@ public Task PersistSnapshotAsync(string actorName, ulong eventIndex, object snap
...
         }
+
+        public Task DeleteEventsAsync(string actorName, ulong fromEventIndex)
+        {
+            throw new NotImplementedException();
+        }
+
+        public Task DeleteSnapshotsAsync(string actorName, ulong fromEventIndex)
+        {
+            throw new NotImplementedException();
```
cpx86: These should probably be implemented as well. If we want to keep this class - I'm not sure if it's needed for anything except testing (in which case it should be moved)

cpx86:
Looks good! I like the introduction of the recovery messages, so the actor can handle those differently if necessary :)

cpx86:
Just fix the merge conflict and I'll merge 👍

src/Proto.Persistence/InMemoryProviderState.cs
raskolnikoov: I think we should remove InMemoryProvider. Dont see the use case for it really. If we need a default provider maybe we should persist om disk in a File.

cpx86: Do it then :)

rogeralsing:
Merged, thanks!

PR #91, opened 2017-03-22
Cluster

rogeralsing:
Ready for merge


examples/ClusterHelloWorld/Node2/Program.cs
```diff
@@ -0,0 +1,45 @@
...
+                            Message = "Hello from node 2"
+                        });
+                        break;
+                    default:
+                        break;
+                }
+                return Actor.Done;
+            });
+
+            Remote.RegisterKnownKind("HelloKind", props);
```
rogeralsing: Register a known kind "HelloKind"

examples/ClusterHelloWorld/Node2/Program.cs
```diff
@@ -0,0 +1,45 @@
...
+            Remote.Start("127.0.0.1", 12000);
+            Cluster.Start("MyCluster",null);
```
rogeralsing: Boot up the cluster. we don't have a consul provider impl yet. thus null (which would fail)

src/Proto.Actor/EventStream.cs
```diff
@@ -42,6 +42,20 @@ public Subscription<T> Subscribe(Action<T> action)
             return new Subscription<T>(sub, this);
         }
 
+        public Subscription<T> Subscribe<TMsg>(Action<TMsg> action) where TMsg : T
```
rogeralsing: Utility impl for eventstreams, make it possible to subscribe to a given type

src/Proto.Actor/Extensions.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Proto
+{
+    public static class Extensions
+    {
+        public static void Deconstruct<TKey,TValue>(this KeyValuePair<TKey, TValue> self,out TKey key, out TValue value)
```
rogeralsing: Syntax sugar, pretty foreach over dict

src/Proto.Actor/LocalContext.cs
```diff
@@ -525,5 +528,16 @@ private void ReceiveTimeoutCallback(object state)
         {
             Self.Request(Proto.ReceiveTimeout.Instance, null);
         }
+
+        public void ReenterAfter<T>(Task<T> target, Func<Task<T>,Task> action)
```
rogeralsing: New context func to allow actor to reenter processing after a task is completed

raskolnikoov:
Does Proto.Actor Cluster work like Akka.NET Cluster with seed nodes etc? Or how does it work.

rogeralsing:
@raskolnikoov no, Proto.Actor cluster is more like Microsoft Orleans.
You do not have to deal with nodes yourself, you simply tell the cluster that you want to talk to an actor of a given type with a given name and the cluster makes sure such actor exists.

The whole thing is backed by Consul.IO, but the provider is pluggable.

src/Proto.Cluster/HashRing.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace Proto.Cluster
+{
+    public class HashRing
+    {
+        private const int ReplicaCount = 100;
```
rikbosch: Should this value be documented?

src/Proto.Cluster/PidCache.cs
```diff
@@ -0,0 +1,111 @@
...
+
+namespace Proto.Cluster
+{
+    public static class PidCache
+    {
+        public static PID PID { get; private set; }
+
+        public static void Spawn()
+        {
+            var props = Router.Router.NewConsistentHashPool(Actor.FromProducer(() => new PidCachePartitionActor()), 128);
```
rikbosch: This number should be documented I think?

src/Proto.Actor/LocalContext.cs
rogeralsing: It is the same as we introduced in Go which was called AwaitFuture there, it creates a continuation and runs the continuation within the actor concurrency constraint when the task has completed.
This allows the actor to process other messages while the task is running

src/Proto.Actor/LocalContext.cs
```diff
@@ -286,25 +289,25 @@ public void EscalateFailure(PID who, Exception reason)
 
         public void RestartChildren(params PID[] pids)
         {
-            for (int i = 0; i < pids.Length; i++)
+            foreach (var pid in pids)
```
adamhathcock: foreach on a generic list is a struct allocation I believe.

src/ClusterProviders/Proto.Cluster.Consul/ConsulProvider.cs
```diff
@@ -0,0 +1,134 @@
...
+        public async Task RegisterMemberAsync(string clusterName, string host, int port, string[] kinds)
+        {
+            _id = $"{clusterName}@{host}:{port}";
+            _clusterName = clusterName;
+            _index = 0;
+            _ttl = TimeSpan.FromSeconds(3);
+            _refreshTtl = TimeSpan.FromSeconds(1);
+            _deregisterCritical= TimeSpan.FromSeconds(10);
+            _blockingWaitTime = TimeSpan.FromSeconds(20);
+
```
rogeralsing: Yes, the Consul settings should also be configurable.
I suppose the new .NET core `IOption<T>` is the modern way to go about handling options.

src/ClusterProviders/Proto.Cluster.Consul/ConsulProvider.cs
```diff
@@ -0,0 +1,134 @@
...
+            _client = new ConsulClient(c =>
+            {
+               //TODO: config
```
rogeralsing: Good idea. we could let the user pass in the func to be used there.

src/ClusterProviders/Proto.Cluster.Consul/ConsulProvider.cs
```diff
@@ -0,0 +1,134 @@
...
+                            TTL = _ttl
+                        }
+            };
+            await _client.Agent.ServiceRegister(s);
+
+
+            //register a unique ID for the current process
+            //similar to UID for Akka ActorSystem
+            //TODO: Orleans just use an int32 for the unique id called Generation.
+            var kvKey = $"{_clusterName}/{host}:{port}"; //slash should be present
```
rogeralsing: Will do, currently we just use some semi random/unique string. but this could be something simpler like a long value with the current tick.
But I will address the comment in the code

src/Proto.Actor/LocalContext.cs
rogeralsing: Cosmetic reasons only, this is a non-critical path in terms of performance.
it will occur occasionally only.

src/Proto.Cluster/HashRing.cs
```diff
@@ -0,0 +1,50 @@
...
+                    hashKey = i + n,
+                    node = n
+                }))
+                .Select(a => Tuple.Create(Hash(a.hashKey), a.node))
+                .OrderBy(t => t.Item1)
+                .ToList();
+        }
+
+        public string GetNode(string key)
+        {
```
rogeralsing: I'd vote for using the C# hashring as the base and port it over to Go.
The current Go impl is less effective

src/Proto.Actor/LocalContext.cs
rogeralsing: I have added doc comments for the Interface method

rogeralsing:
cc @AsynkronIT/developers comments have been added.
Anything else we need to address for now?

cpx86:
@rogeralsing Nope, LGTM! Shall we merge it?

cpx86:
One thing though, it seems your Git email has changed so the CLA bot doesn't recognize you.

rogeralsing:
@cpx86 yes I saw that the commit history sees me as two users, not sure why, my settings are correct.

It's ready for merge otherwise. 
Not sure if there is an easy way to correct that messed up commit with the wrong email? (its only one in the middle of the history)

cpx86:
@rogeralsing Should be quite easy using an interactive rebase - see http://stackoverflow.com/questions/3042437/change-commit-author-at-one-specific-commit

rogeralsing:
I've managed to kill my own git history a few times now. pulling in and maybe we can fix up any history issues later

PR #93, opened 2017-03-22
marten persistence

jeffdoolittle:


rogeralsing:
Looks good to me.
cc @cpx86 one thing we should consider is how to maintain these kind of providers.
Should we keep only one of them as "official" and others as "contributions" in order to lower support burden?

jeffdoolittle:
I'm doing a bit more testing on this and having an issue with mapping a uint with postgres:

> The CLR type System.UInt64 isn't supported by Npgsql or your PostgreSQL. If you wish to map it to a PostgreSQL composite type you need to register it before usage, please refer to the documentation.

jeffdoolittle:
Temporary hack is to cast all `ulong` to long in the `Envelope` and `MartenProviderState` classes.

I'm curious the need for 18,446,744,073,709,551,615 values as opposed to the 9,223,372,036,854,775,807 available for a `long` that starts at zero? I don't mean this to sound critical, especially since this is my first foray into contributing to this project. Just trying to understand.

cpx86:
@rogeralsing Added some thoughts about this in #99.

cpx86:
@jeffdoolittle There isn't really any reason other than that it makes sense that it's unsigned (since you can't have a negative number of events) and I see no reason for arbitrarily limiting it to a 32-bit number. On the other hand, SQL databases generally don't support unsigned integers, so changing it to a signed long could make sense.

jeffdoolittle:
If you guys are okay with it I can alter the PR to use `long` instead of `ulong` in the persistence abstractions and the existing persistence store implementations.

cpx86:
Go ahead :)

jeffdoolittle:
Rebased, squished and ready to go. I temporarily replaced the `Persistence` example locally to use Marten instead of CouchDB and it works great.

cpx86:
Ping @raskolnikoov. Any input on the above question?

raskolnikoov:
@jeffdoolittle I changed ravendb to separate snapshots and events because I like that style instead of a "flat" database table that has everything. Every implementation of persistence can choose how they thinks the stored data should look like. :)

jeffdoolittle:
Since Marten and RavenDB have a lot in common it might be nice to have their persistence mechanisms function in a similar way as well. I can leave this as is, or change it depending on what you all think would be best. Let me know and thanks for engaging on this with me.

raskolnikoov:
@jeffdoolittle I would vote for the ravendb implementation style just because thats how I like it. And as you said, marten and ravendb is similar how they work.

raskolnikoov:
I wrote a akka persistence marten plugin a couple months ago and then I separated snapshots and events.

https://github.com/raskolnikoov/Akka.Persistence.Marten

jeffdoolittle:
@raskolnikoov great, I can make that change and resubmit the PR. thanks.

jeffdoolittle:
Let me know if this looks good and I can squish and push up the (final?) commit. Thanks.

raskolnikoov:
Looks good!

rogeralsing:
Thank you for the contribution!

jeffdoolittle:
Glad to help. Looking forward to other ways I can contribute to this project in the future.

jeffdoolittle:
Now that this is merged in, it occurs to me that this may be better implemented using the Marten EventStream functionality rather than using documents. This would allow for hooking in to the projections API for inline or async daemon projection generation. Thoughts? If there's interest I can make the change and submit a new PR.

raskolnikoov:
Yeah that sounds like a better Choice.

PR #95, opened 2017-03-23
added Scheduler api for IContext and added example application

raskolnikoov:
Implemented Scheduler api as an extension for IContext on the actor.

It can be used in two ways.

**ScheduleTellOnce - fire and forget with delay**
```
context.ScheduleTellOnce(TimeSpan.FromSeconds(1), context.Self, new Hello());
```

**ScheduleTellRepeatedly - fire with delay, interval and cancellation**
```
var timer = context.ScheduleTellRepeatedly(TimeSpan.FromSeconds(5), TimeSpan.FromMilliseconds(250), context.Self, new Hello());

timer.Cancel();
```

### DIFF
--- a/src/Proto.Actor/Scheduler.cs
+++ b/src/Proto.Actor/Scheduler.cs
@@ -1,0 +1,46 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Proto
+{
+    public static class Scheduler
+    {
+        public static void ScheduleTellOnce(this IContext context, TimeSpan delay, PID target, object message)
+        {
+            Task.Run(async () =>
+            {
+                await Task.Delay(delay);
+
+                context.Tell(target, message);
+            });
+        }
+
+        public static CancellationTokenSource ScheduleTellRepeatedly(this IContext context, TimeSpan delay, TimeSpan interval, PID target, object message)
+        {
+            CancellationTokenSource cts = new CancellationTokenSource();
+
+            var task = Task.Run(async () =>
+            {
+                await Task.Delay(delay);
+
+                async void Trigger(object _message)
+                {
+                    if (cts.IsCancellationRequested)
+                        return;
+
+                    context.Tell(target, message);
+
+                    await Task.Delay(interval);
+
+                    Trigger(_message);
+                }
+
+                Trigger(message);
+
+            }, cts.Token);
+
+            return cts;
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


raskolnikoov:
@rogeralsing @cpx86

src/Proto.Actor/Scheduler.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    public static class Scheduler
+    {
+        public static void ScheduleTellOnce(this IContext context, TimeSpan delay, PID target, object message)
+        {
+            Task.Run(async () =>
+            {
+                await Task.Delay(delay);
+
+                context.Tell(target, message);
```
rogeralsing: This will be the subject of race conditions.
the context is never intended to be passed off to other async operations t must adhere to the actor concurrency constraint.

What could happen here is that the actor can continue to process messages, update it's current message and message headers.
So when `context.Tell` is invoked, the headers have been changed, making the outbound middleware to be non deterministic in behavior (as you dont know what headers is active when Tell is invoked)

rogeralsing: Another question that we need to decide on, should scheduled messages go through the current actors middleware?
(I would say yes, but there might be more to this that needs to be explored)

raskolnikoov: Would it be better to pass on the actors PID instead and do the tell on that?

cpx86: Yes, what about context.Self.Tell? That should be safe, right @rogeralsing?

rogeralsing: That would be safe, _but_ what is the semantics we want here?
Should scheduled messages use the outbound middleware?
e.g. if you have a tracing middleware, should the trace continue when scheduling?
(imo yes, and if so, we need to deal with this in this code)

jrouaix:
Guys, this is two pull requests about Scheduling
https://github.com/AsynkronIT/protoactor-dotnet/pull/69 
Don't you want to merge this one or give me a feedback so I can make it good enough to integrate ?

cpx86:
@jrouaix Hmm yes we need to decide on how this should look. What's the main difference between this and #69? Can we get them aligned into a single PR? Ping @raskolnikoov

rogeralsing:
We should ignore specific technical impls for now IMO and discuss the semantics we want to achieve around middleware

raskolnikoov:
Im closing this PR and instead focusing on the new one i've created at https://github.com/AsynkronIT/protoactor-dotnet/pull/119

PR #98, opened 2017-03-23
Fix LocalContext.PopBehavior not restoring previously pushed behavior…

jokokko:
… (unifies the implementation with the Go version). Created unit test to exercise the behavior.


rogeralsing:
Great catch!
Thanks for your contribution

PR #103, opened 2017-03-27
Children should be initialized to empty to avoid null children on context

adamhathcock:
Other style clean up

### DIFF
--- a/src/Proto.Actor/LocalContext.cs
+++ b/src/Proto.Actor/LocalContext.cs
@@ -16,7 +16,7 @@ namespace Proto
 {
     public class Context : IMessageInvoker, IContext, ISupervisor
     {
-        private static readonly IReadOnlyCollection<PID> Empty = new List<PID>();
+        public static readonly IReadOnlyCollection<PID> EmptyChildren = new List<PID>();
 
         private readonly Stack<Receive> _behavior;
         private readonly Receive _receiveMiddleware;
@@ -58,7 +58,7 @@ namespace Proto
             }
         }
 
-        public IReadOnlyCollection<PID> Children => _children?.ToList() ?? Empty;
+        public IReadOnlyCollection<PID> Children => _children?.ToList() ?? EmptyChildren;
 
         public IActor Actor { get; private set; }
         public PID Parent { get; }
--- a/src/Proto.Actor/LocalContext.cs
+++ b/src/Proto.Actor/LocalContext.cs
@@ -21,7 +21,7 @@ namespace Proto
         private readonly Sender _senderMiddleware;
         private readonly Func<IActor> _producer;
         private readonly ISupervisorStrategy _supervisorStrategy;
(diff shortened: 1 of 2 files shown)


cpx86:
There is a benefit to lazily assigning the children field, which is that we avoid an unnecessary allocation for actors that don't have any children. It's possible that this could be something of a premature optimization - it all depends on what the most common use case is - if most actors have children then it makes sense to preallocate it to avoid the null-checks, but if most don't then it makes sense to lazily allocate it. @rogeralsing What are your thoughts?

adamhathcock:
Well, I could change it to also lazy initialize the collection when the `Children` property is accessed.

I just hate having a null collection and it burned me today.

cpx86:
How did it affect you? NullReferenceException?

rogeralsing:
I'm also curious on when and how this happened?

Actor "weight" is super important, this will directly affect the cost of spawning actors.
Having a lazy eval property might work, assuming we do not prematurely instance the list/set when checking length. e.g. `if (Children.Any())`

adamhathcock:
yeah, I got a null exception.  I was accessing the `Children` collection before any were created.  Just didn't expect it to be null.

I can ensure if there's any length or Any checking that it also does a null check to avoid initializing if possible.

adamhathcock:
Even better, static empty collection.

src/Proto.Actor/LocalContext.cs
```diff
@@ -31,7 +33,7 @@ public class Context : IMessageInvoker, IContext, ISupervisor
         private bool _stopping;
         private HashSet<PID> _watchers;
         private HashSet<PID> _watching;
-        private readonly ILogger logger = Log.CreateLogger<Context>();
+        private readonly ILogger _logger = Log.CreateLogger<Context>();
```
rogeralsing: Can we share the logger between contexts?
The current design also comes with a weight per actor.

src/Proto.Actor/LocalContext.cs
```diff
@@ -56,7 +58,8 @@ public Context(Func<IActor> producer, ISupervisorStrategy supervisorStrategy, Re
             }
         }
 
-        public IReadOnlyCollection<PID> Children => _children?.ToList();
+        public IReadOnlyCollection<PID> Children => _children?.ToList() ?? EmptyChildren;
```
rogeralsing: 👍

src/Proto.Actor/LocalContext.cs
adamhathcock: The logger isn't thread-safe.  The logger class should be per instance.

If no logging isn't turned on, then no objects are created though.  So if people are really perf concerned, then don't use logging.

adamhathcock: Of course, if logging statements create strings by concatenation or whatever then those are still evaluated unless logging level is checked.

PR #106, opened 2017-03-27
As we’ve stopped children, forget them.

adamhathcock:
This is needed because of the `if (_children?.Count > 0)` check in `TryRestartOrTerminateAsync`

### DIFF
--- a/src/Proto.Actor/LocalContext.cs
+++ b/src/Proto.Actor/LocalContext.cs
@@ -451,6 +451,7 @@ namespace Proto
                 {
                     child.Stop();
                 }
+                _children = null;
             }
             await TryRestartOrTerminateAsync();
         }


src/Proto.Actor/LocalContext.cs
```diff
@@ -451,6 +451,7 @@ private void HandleRootFailure(Failure failure)
                 {
                     child.Stop();
                 }
+                _children = null;
```
rogeralsing: This is not the right way to do it, we need to wait for each child to be completely stopped, which is signaled by a `Terminate` message.

The right place to null the _child collection would be here once the child count is 0:
https://github.com/adamhathcock/protoactor-dotnet/blob/33a6e8120f49a1f8f1c725b5733e8b5388dd1de9/src/Proto.Actor/LocalContext.cs#L431

adamhathcock:
Not a good solution

Issue #92, opened 2017-03-22
Notify watchers on actor stopped


cpx86:
When an actor is stopped we don't notify the watchers.

Go implements this: https://github.com/AsynkronIT/protoactor-go/blob/dev/actor/local_context.go#L361

PR #107, opened 2017-03-28
Notify watchers when actor is stopped/terminated

adamhathcock:
Fixes https://github.com/AsynkronIT/protoactor-dotnet/issues/92
Supercedes https://github.com/AsynkronIT/protoactor-dotnet/pull/106

### DIFF
--- a/examples/Supervision/Program.cs
+++ b/examples/Supervision/Program.cs
@@ -33,6 +33,7 @@ class Program
         actor.Tell(new Recoverable());
         actor.Tell(new Fatal());
 
+        actor.Stop();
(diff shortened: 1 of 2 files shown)


cpx86:
Nice! Just one small part missing. If you try to Watch an actor that is stopping, we should reply with a Terminated message. See Go impl here: https://github.com/AsynkronIT/protoactor-go/blob/dev/actor/local_context.go#L264

adamhathcock:
Just fixed.  I saw that then promptly forgot to add it :(

cpx86:
Great work - thanks!

PR #111, opened 2017-03-30
Persistence updates

raskolnikoov:
This merge has the following changes:

- Updated couchbase provider to work with the new persistence api
- Updated demo app for persistence to show event/snapshot recovering
- Updated persistence api recovery logic to only recover the events that has newer indexes than the last snapshot. This change after discussion on gitter on how the recovery could work better.
- Removed index counter update on snapshot


src/Proto.Persistence/Persistence.cs
```diff
@@ -28,39 +27,39 @@ public class Persistence
...
         public async Task PersistSnapshotAsync(object data)
         {
-            var persistSnapshotIndex = SnapshotIndex;
+            var index = Index;
 
-            await State.PersistSnapshotAsync(Name, persistSnapshotIndex, data);
+            await State.PersistSnapshotAsync(Name, index, data);
 
-            SnapshotIndex++;
+            Index++;
```
tomliversidge: I've not tested yet, but I don't think saving a snapshot should increment the event index. For example, I take a snapshot at event index 10, then 5 minutes pass and I take another snapshot, but I received no new events during this time, the event index should be the same, 10. 

Another example, if I take 20 snapshots at event index 10, they are all the same capture of the state of the actor at that index. The event index shouldn't now be 30 as no new events have been received.

tomliversidge: I've not tested yet, but I don't think saving a snapshot should increment the event index. For example, I take a snapshot at event index 10, then 5 minutes pass and I take another snapshot, but I received no new events during this time, the event index should be the same, 10.

Another example, if I take 20 snapshots at event index 10, they are all the same capture of the state of the actor at that index. The event index shouldn't now be 30 as no new events have been received.

src/Proto.Persistence/Persistence.cs
```diff
@@ -28,39 +27,39 @@ public class Persistence
 
             if (t != null)
             {
-                SnapshotIndex = t.Item2;
+                Index = t.Item2;
                 await Context.ReceiveAsync(new RecoverSnapshot(t.Item1));
             }
 
-            await State.GetEventsAsync(Name, EventIndex, e =>
+            await State.GetEventsAsync(Name, Index, e =>
```
tomliversidge: 👍 👍 👍 index is what was returned from snapshot

src/Proto.Persistence/Persistence.cs
raskolnikoov: Yes u are correct, I will change this.

raskolnikoov: @tomliversidge this is now updated

jrouaix:
GG guys, totally aggree with that @tomliversidge (https://gitter.im/AsynkronIT/protoactor?at=58dd5956b402a53211cbdf25)

rogeralsing:
LGTM, ready for merge? anyone else have any opinions?

raskolnikoov:
It's friday, so ready!!!

cpx86:
LGTM! 😄

Issue #108, opened 2017-03-30
Make Context._logger lazy


rogeralsing:
Having the eager init of the logger adds a few % overhead to actor creation.
By making it lazy we can shave off that overhead.

With the eager logger spawn benchmark: 1.1 sec
With lazy logger spawn benchmark: 0.97 sec

raskolnikoov:
Yes I agree, lazy is king!

adamhathcock:
So the assigning of the singleton instance of the null logger causes a difference of ~0.1 sec?

Did you just remove the log statement when you ran it otherwise?

adamhathcock:
You might get more perf by doing the _behavior stack lazily.  If there's no extra receives then why do it?

I guess what I'm getting at is what is enough perf without making things too complex?

cpx86:
@adamhathcock Good idea - in many cases you'll only have one default behavior so it might be more optimal to null-check the behavior stack on each receive and only use it when you have >1 behaviors (default + 1 alternative)

adamhathcock:
I just went ahead and did it then :p

cpx86:
Fixed by #112

PR #112, opened 2017-03-31
Make logging and behavior stack creation lazy.

adamhathcock:
Covers https://github.com/AsynkronIT/protoactor-dotnet/issues/108

### DIFF
--- a/src/Proto.Actor/LocalContext.cs
+++ b/src/Proto.Actor/LocalContext.cs
@@ -18,7 +18,7 @@ namespace Proto
     {
         public static readonly IReadOnlyCollection<PID> EmptyChildren = new List<PID>();
 
-        private readonly Stack<Receive> _behavior;
+        private Stack<Receive> _behavior;
         private readonly Receive _receiveMiddleware;
         private readonly Sender _senderMiddleware;
         private readonly Func<IActor> _producer;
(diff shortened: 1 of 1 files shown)


src/Proto.Actor/LocalContext.cs
```diff
@@ -124,18 +134,26 @@ public PID SpawnNamed(Props props, string name)
 
         public void SetBehavior(Receive receive)
         {
-            _behavior.Clear();            
+            _behavior = null;
```
adamhathcock: Not sure if clear is better than setting it to null.  I guess I should just clear it if it's already made.

rogeralsing: I think this is good, if we can clear out objects that are no longer used, we should do so 👍

rogeralsing:
Looks good!

src/Proto.Actor/LocalContext.cs
```diff
@@ -124,18 +134,26 @@ public PID SpawnNamed(Props props, string name)
...
+            }
             _behavior.Push(_receive);
             _receive = receive;
         }
 
         public void PopBehavior()
         {
+            if(_behavior == null)
+            {
+                return;
```
rogeralsing: Thinking about it, we should probably throw this ´throw new Exception("Can not unbecome actor base behavior");` when _behavior is null, as the meaning is the same as the later statements.

PR #113, opened 2017-03-31
Throw exception when behavior is null as it means the same as zero behaviors added.

adamhathcock:
From comment on last PR.


rogeralsing:
👍

Issue #87, opened 2017-03-17
Strongly-typed handlers


adamhathcock:
I modeled this after MediatR.  I basically want the same behavior.  Invoke a message and have it routed to an Actor

Declaration
```
    public class TestActor : IActor,
        IActorRequestHandler<TestRequest1>,
        IActorRequestHandler<TestRequest2, string>
    {
        public Task Handle(TestRequest1 request, IContext context)
        {
            Console.WriteLine("Got test 1");
            return Task.CompletedTask;
        }

        public Task<string> Handle(TestRequest2 request, IContext context)
        {
            Console.WriteLine("Got test 2");
            return Task.FromResult("test");
        }
        //default
        public Task ReceiveAsync(IContext context)
        {
            return Task.CompletedTask;
        }
    }
```

Test usage
```
  static void Main(string[] args)
        {
            ActorRegistry.Scan(Assembly.GetEntryAssembly());

            var serviceCollection = new ServiceCollection();
            serviceCollection.AddProtoActor();
            var sp = serviceCollection.BuildServiceProvider();

            ActorHandle handle = sp.GetRequiredService<IActorFactory>()
                .GetActor<TestActor>();
            
            handle.Publish(new TestRequest1());
            handle.Publish(new TestRequest1());
            handle.Publish(new TestRequest1());

            var response = handle.Send(new TestRequest2()).Result;
            Console.WriteLine(response);

            response = handle.Send(new TestRequest2()).Result;
            Console.WriteLine(response);

            Console.ReadLine();
        }
```

adamhathcock:
This is all done with an extension library and using the receive middleware to intercept the message.

It's done with some reflection that could be cleaned up.

adamhathcock:
I think I'm gonna dump ActorHandle and just have extension methods off of PID.  While I hate the name PID, ActorHandle doesn't add anything to it.

rogeralsing:
As discussed in various places, let's keep the API surface area as small as possible.
Users can add this themselves if needed

PR #118, opened 2017-04-01
Adding new ReceiveActor class that process messages using strongly ty…

Armando-GarzaTech:
@jakob-garzatech and I have a proposed solution to issue #87 that we would like to submit for your consideration.

### DIFF
--- a/src/Proto.Actor/ReceiveActor.cs
+++ b/src/Proto.Actor/ReceiveActor.cs
@@ -1,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Proto
+{
+    public class ReceiveActor : IActor
+    {
+        private readonly Dictionary<Type, Func<IContext,Task>> _handlers
+            = new Dictionary<Type, Func<IContext, Task>>();
+
+        protected void RegisterHandler<T>(Func<T, IContext, Task> handler)
+        {
+            _handlers.Add(typeof(T), context => handler.Invoke((T) context.Message, context));
+        }
+
+        public Task ReceiveAsync(IContext context)
+        {
+            var messageType = context.Message.GetType();
+
+            return _handlers.ContainsKey(messageType) 
+                ? _handlers[messageType](context) 
+                : OnUnknownMessage(context);
+        }
+
+        protected virtual Task OnUnknownMessage(IContext context)
+        {
+            return Actor.Done;
+        }
+    }
+}
--- a/tests/Proto.Actor.Tests/ReceiveActorTests.cs
+++ b/tests/Proto.Actor.Tests/ReceiveActorTests.cs
@@ -1,0 +1,46 @@
+﻿using System.Threading.Tasks;
+using Xunit;
+
+namespace Proto.Tests
+{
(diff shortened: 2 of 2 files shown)


rogeralsing:
I'm all for having this kind of functionallity.
_But_, I would rather see it as a contrib extension.
The thing is that once you start adding inheritence to the mix, you get into all sorts of issues in terms of API evolution.

in Akka.NET, we had ReceiveActors, PersistentReceiveActors FsmActors, and a whole bunch of permutations of these combinations.
I would very much like to avoid this kind of scenario repeating in ProtoActor.

cc @AsynkronIT/developers

raskolnikoov:
@rogeralsing I agree, keep it simple and let the developers add stuff if needed. :)

Armando-GarzaTech:
I agree with no creating an extended API to handle our concern. I will close this pull request. 
However, we would be interested in starting a contrib extension. Is there a standard way to accomplish this or are we free to create something through Nuget?

Thanks - Armando

adamhathcock:
I'm still not sure I even like mine.  The C# 7 pattern matching makes things nicer.  Though, I still would like strongly typed replies.  But that's my only real issue at the moment.

Armando-GarzaTech:
@adamhathcock we decided to go with an implementation that didn't rely on inheritance and implemented the message routing in a helper class (MessageRouter). This way we could ensure that messages are typed when passed to a handler as well as providing multiple message routers that we can use to change actor behavior via a message. We felt that a fluent interface made the declaration of the handlers clear and easier to read.


    public class MessageRouter
    {
        private Func<IContext, Task> _otherwise;
        private readonly Dictionary<Type, Func<IContext, Task>> _whenHandlers
            = new Dictionary<Type, Func<IContext, Task>>();

        public MessageRouter When<T>(Func<T, IContext, Task> handler)
        {
            _whenHandlers.Add(typeof(T), context => handler((T)context.Message, context));
            return this;
        }

        public void Otherwise(Func<IContext, Task> handler)
        {
            _otherwise = handler;
        }

        public Task Handle(IContext context)
        {
            var messageType = context.Message.GetType();
            return _whenHandlers.ContainsKey(messageType)
                ? _whenHandlers[messageType](context)
                : _otherwise == null
                    ? Actor.Done
                    :_otherwise(context);
        }
    }

**SAMPLE ACTOR:**

    class MyTestActor : IActor
    {
        private readonly MessageRouter _router = new MessageRouter();

        public MyTestActor()
        {
	     _router
            .When<string>((msg, context) =>
            {
                context.Respond("string handled");
                return Actor.Done;
            })
            .When<int>(HandleInt)
            .Otherwise(context =>
            {
                context.Respond("otherwise handled");
                return Actor.Done;
            });
        }

        private Task HandleInt(int value, IContext context)
        {
            context.Respond(41 + value);
            return Actor.Done;
        }

        public Task ReceiveAsync(IContext context)
        {
            return _router.Handle(context);
        }
    }

adamhathcock:
Thanks for the info.

However, that looks much more complex than the C# 7 switch case pattern matching.  Doing that is less characters, still strongly typed, and likely much more performant than your solution.  I would have done that before C# 7 but not now.

Issue #97, opened 2017-03-23
Scheduling messages - Semantics and middleware


rogeralsing:
We now have two different PR's for scheduling messages.
Before picking an implementation, we should think about middleware semantics.

e.g.
Let's say we have a tracing middleware.
And now someone tries to schedule a message in 10 seconds.
Should that message go through the middleware?
Should the trace be continued from the original point forward to the receiver of the scheduled message?

This relates to #95  #69

rogeralsing:
cc @jrouaix @raskolnikoov @cpx86

rogeralsing:
Another thing to consider.
If we want to scheduler to use the middleware.

Lets say an actor schedules a message in 10 seconds, then the actor is stopped.
Should the message still be sent?
Should the message still go through the now dead middleware?

jrouaix:
yep so, after comments from @cpx86 (https://github.com/AsynkronIT/protoactor-dotnet/pull/69#issuecomment-286994606) the implementation I proposed is using the SenderMiddleware.

Basically some extensions to the IContext (https://github.com/Bee-Htcpcp/protoactor-dotnet/blob/c94d5dea71e639e6902e10271cfa56f1adec5514/src/Proto.Scheduling/SchedulingExtensions.cs) wrapping the message to a "DelayMessage" instance, and tell to self this message. 
Then the middleware implementation catching this DelayMessage and use the IScheduler implementation given to Schedule the message.

The IScheduler interface is stupid simple : https://github.com/Bee-Htcpcp/protoactor-dotnet/blob/c94d5dea71e639e6902e10271cfa56f1adec5514/src/Proto.Scheduling/IScheduler.cs and the extensions methods are just proposing to delay a message to "Self". Because _in my opinion_ no one but the actor should delay something to it. (https://github.com/AsynkronIT/protoactor-dotnet/pull/69#issuecomment-286750465)

The activation is not yet covered, right now, a message delayed to a stopped actor will end in the dead letters (tested here : https://github.com/Bee-Htcpcp/protoactor-dotnet/blob/c94d5dea71e639e6902e10271cfa56f1adec5514/tests/Proto.Scheduling.Tests/InMemorySchedulerTests.cs#L59)

I proposed to have 2 pipes : one Transcient & one Persisted, right now they are implemented just the same but we could introduce slight changes.

I don't know yet how this will integrate with "named actors" "cluster activation" in my opinion :
[...]

PR #119, opened 2017-04-02
Simple scheduler with example

raskolnikoov:
New schedule structure after discussions on PR https://github.com/AsynkronIT/protoactor-dotnet/pull/95


raskolnikoov:
SimpleSchedule works as an extension to IContext.

rogeralsing:
I like this one, simple, just an extension, no fancy structures to hold the scheduler data. 
👍

raskolnikoov:
@rogeralsing I changed my imlementation to not use this as an extension. So no more depencies for IContext, instead you pass the target (and if needed sender) PID's.

tomliversidge:
Looks like this might need rebasing (and squashing?) - seems to have become a bit corrupted at some point as it is showing tests that are already in dev as being new - maybe from when the Proto.Router.Tests project was created at the root of the repo?

raskolnikoov:
@tomliversidge yeah I fixed this a moment ago. Have a look at it again and then I'll merge.

tomliversidge:
1. Tests? :) Would be interesting to try and test schedulers. Need to take away the time part somehow. 
2. This pull request has 13 commits - would look better in the history as 1 - "Add simple scheduler". I personally prefer to rebase and squash 
3. What else would implement ISimpleScheduler? I'm wondering if you need an interface for this.

What about the discussions in https://github.com/AsynkronIT/protoactor-dotnet/issues/97 around middleware and what to do when actors are stopped?

raskolnikoov:
ping @tomliversidge 

1. Yeah this we need in the future, but for now I dont have any.
2. Yes I got my git skills poorly on this one. :)
3. I always try to implement to interfaces when coding. So for now no one else would implement to ISimpleScheduler I think. But it's nice to have.

raskolnikoov:
Ready to merge if its okay by you other? ping @rogeralsing  @cpx86

cpx86:
I'm a strong proponent of "merge early - change later". It's clean and provides value so let's not bikeshed it any more :)

PR #120, opened 2017-04-03
Start of Proto.Router.Tests

tomliversidge:
Tests for RoundRobin and GroupBroadcast routers. (others to be added soon!)

NOTE: I had to change the Router.Spawner to pass through the Mailbox from props arguments in order to give the router a TestMailbox


Proto.Router.Tests/BroadcastGroupTests.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            var routees = await router.RequestAsync<Routees>(new RouterGetRoutees());
+            Assert.Contains(routee1, routees.PIDs);
+            Assert.Contains(routee2, routees.PIDs);
+            Assert.Contains(routee3, routees.PIDs);
+            Assert.Contains(routee4, routees.PIDs);
+        }
+
+        [Fact]
+        public async void RoundRobinGroupRouter_RemovedRouteesNoLongerReceiveMessages()
```
cpx86: Nit-pick, but these should be in a separate file RoundRobinGroupTests :)

cpx86: Wrong test name? It's for Broadcast, not RoundRobin right?

tomliversidge: Ahh yes good spot

cpx86:
Nice work! Just a small nit-pick about a test name ;)

tomliversidge:
Ok, I fixed the test name (I squashed the commits into 1 so the history doesn't have "fixed test name" in it :) )

cpx86:
LGTM! Thanks a bunch! :)

PR #121, opened 2017-04-03
ConsistentHashGroupRouter tests

tomliversidge:
Adding tests for ConsistentHashGroup router. 

(I think I accidentally included this file in the csproj in the last commit)

These were slightly more complicated than the other tests as I ran into problems running them in isolation vs with all other tests - I think the hash algorithm changed depending on how they were ran. To fix, I now dynamically find which routee processed messages so it shouldn't matter which one the hash ring finds.

The tests basically prove that if the message hash is the same, the same routee processes the messages. If it is different, the messages are spread amongst routees.

### DIFF
--- a/Proto.Router.Tests/ConsistentHashGroupTests.cs
+++ b/Proto.Router.Tests/ConsistentHashGroupTests.cs
@@ -1,0 +1,227 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


tomliversidge:
Will move the project first then re-open this

cpx86:
The hashring is a bit messy to predict what it'll do, e.g. each node in the ring is replicated 100 times in order to improve the distribution between nodes.

One way to fix it which could also be a nice feature is to make the hash ring pluggable so you can make it deterministic for tests.

PR #123, opened 2017-04-03
Fast set for PID's.

rogeralsing:
Performance optimization for watchers and watchees.
Only allocate a full hashset when there are more than one PID in the set.
Leads to less allocations for actors with no or one children.


cpx86:
👍 

Any perf numbers?

rogeralsing:
we are at 780-980 ms for skynet right now.
Erlang 715-900 ms.

So we are in really good shape right now, considering Erlang have been optimizing this for 20 years.

PR #125, opened 2017-04-03
tests for ConsistentHashGroupRouter

tomliversidge:
To allow deterministic testing, you can now pass in a function to compute the hash in the HashRing

Tests use a super-simple deterministic hashing function to determine which messages go to which routee.


cpx86:
Nice! 👍

PR #126, opened 2017-04-04
tests for RandomGroupRouter and Pool routers

tomliversidge:
Added ability to pass in a seed to the random function so that tests can be deterministic. 

Not sure if there is anything else to test for the pool routers - I'm just testing that they create a pool of routees.


rogeralsing:
👍

PR #127, opened 2017-04-04
Redesigning process

rogeralsing:
Redesigning `Process`, removed Sender from API.
We already have `MessageEnvelope` which carries sender information,
previously the message envelope was created inside the local process.
Now it is created at the call site.
This is done to enable passing message headers together with the message and to eventually complete the outbound middleware.

In my machine the Router tests break now, I cant fint out why. 
Any router based example works fine, so I need to locate the issue before merging.

### DIFF
--- a/src/Proto.Actor/DeadLetter.cs
+++ b/src/Proto.Actor/DeadLetter.cs
@@ -26,14 +26,8 @@ namespace Proto
 
         public override void SendUserMessage(PID pid, object message)
         {
-            if (message is MessageEnvelope envelope)
-            {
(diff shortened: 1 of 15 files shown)


rogeralsing:
Or is there an issue with the router tests that have been fixed recently?
The tests seems to fail silently here on the CI too.

cc @AsynkronIT/developers

rogeralsing:
Any way to re-run tests?
It works locally now

rogeralsing:
Ready for merge cc @AsynkronIT/developers 
Failed test is due to problem in Proto.Mailbox which haven't been changed in this PR.

raskolnikoov:
:+1:

cpx86:
LGTM :)

PR #131, opened 2017-04-05
Small refactoring to Persistence API

tomliversidge:
A few small changes around the persistence API. If you prefer, please merge, if not, just close this :)

- Encapsulate the IProviderState, Name, and IContext - they are now private
- Persistence.Index is now private-settable
- Add DeleteSnapshotAsync method to Persistence as IProviderState is no longer exposed
- Consistent naming around Events and Snapshots - rather than Data, which is very generic, I've renamed to use EventSourcing nomenclature
- Updated providers to use new naming
- Rename fromIndex on IProviderState delete methods. It is actually the opposite of from, as these methods delete all events/snapshots upto and including the passed in index. I can't think of a good name for this parameter though - inclusiveToIndex is pretty rubbish. Some alternatives considered were "eventsLessThanOrEqualToIndex" and "upUntilAndIncludingIndex" - better?


src/Proto.Persistence/Persistence.cs
```diff
@@ -11,51 +11,53 @@ namespace Proto.Persistence
...
         {
             var index = Index;
+            await _state.PersistSnapshotAsync(ActorId, index, snapshot);
+            await _context.ReceiveAsync(new PersistedSnapshot(index, snapshot));
+        }
 
-            await State.PersistSnapshotAsync(Name, index, data);
-            
-            await Context.ReceiveAsync(new PersistedSnapshot(index, data));
+        public Task DeleteSnapshotsAsync(long index)
```
raskolnikoov: Should be async  and awaited?

raskolnikoov: I would also like to have the DeleteEventsAsync method in here.

tomliversidge:
@raskolnikoov - i've added the DeleteEventsAsync method and made things async-await

raskolnikoov:
Great, ill update my PR with these updates.

PR #133, opened 2017-04-05
added persistence sqlserver

raskolnikoov:


jeffdoolittle:
I always liked how the NEventStore project generalized SQL persistence to make it easier to maintain projects for multiple RDBMSs.

https://github.com/NEventStore/NEventStore.Persistence.SQL/tree/master/src

Might be worth doing something similar for protoactor to make it easy to support other RDBMS persistence stores.

PR #135, opened 2017-04-06
Tests for EventStream

tomliversidge:
- made EventStream ctor public. We do not want to use the static EventStream.Instance for tests as they would all interfere with each other anytime a message type was sent that another test subscribed to.

### DIFF
--- a/tests/Proto.Actor.Tests/EventStreamTests.cs
+++ b/tests/Proto.Actor.Tests/EventStreamTests.cs
@@ -1,0 +1,52 @@
+﻿using System.Collections.Generic;
+using Xunit;
+
+namespace Proto.Tests
+{
+    public class EventStreamTests
+    {
+        [Fact]
+        public void EventStream_CanSubscribeToSpecificEventTypes()
+        {
+            var received = "";
+            var eventStream = new EventStream();
+            eventStream.Subscribe<string>(theString => received = theString);
+            eventStream.Publish("hello");
+            Assert.Equal("hello", received);
+        }
+
+        [Fact]
+        public void EventStream_CanSubscribeToAllEventTypes()
+        { 
+            var receivedEvents = new List<object>();
+            var eventStream = new EventStream();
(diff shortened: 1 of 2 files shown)


tests/Proto.Actor.Tests/EventStreamTests.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            var logger = new MockLogger();
+            var pid = new PID("test", "test");
+            var eventStream = new EventStream(logger);
+            eventStream.Publish(new DeadLetterEvent(pid, "dead", null));
+            eventStream.Publish("not a dead letter event");
+            Assert.Equal(1, logger.Events.Count);
+        }
+
+        private class MockLogger : ILogger {
```
adamhathcock: Is mocking and testing the logs a good idea?  I've always found it to be a pain.

tomliversidge:
I guess that depends if anyone cares about DeadLetterEvents being logged :) ?

tomliversidge:
I think you're right @adamhathcock - I'm going to delete the test that something is logged

tests/Proto.Actor.Tests/Proto.Actor.Tests.csproj
```diff
@@ -13,6 +13,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="1.1.1" />
```
adamhathcock: I think you can remove this and the `using` statement

tomliversidge: ahh yes, good catch

tomliversidge: @adamhathcock - OK, removed

rogeralsing:
Awesome!

PR #136, opened 2017-04-06
TO DISCUSS - Allow separate Event and Snapshot persistence to be used

tomliversidge:
Provided as an example - **please do not merge (yet).**

The rational behind this is that I might want to use a different persistence provider for snapshots vs events. For example, something like redis for snapshots, eventstore for events.

This is moving further away from the Go API though


tomliversidge:
Issue to discuss persistence API here https://github.com/AsynkronIT/protoactor-dotnet/issues/137

raskolnikoov:
Is this something we still want to implement now when we have rebuilt our persistence structure?

tomliversidge:
The new persistence API still only supports a single store as it hasn't changed how iprovider or iproviderstore is implemented.

tomliversidge:
I'm going to close this issue as there is an easier, more transparent way to do this based on recent refactorings

Issue #102, opened 2017-03-27
Error handling in EventStream subscriptions


rogeralsing:
We need proper error handling in the event stream handlers.
If the code below fails, that might kill the application

```csharp
        public void Publish(T msg)
        {
            foreach (var sub in _subscriptions)
            {
                sub.Value(msg);
            }
        }
```

tomliversidge:
How should errors here be handled? Also, is it right that it is synchronously going through each subscription? Any reason it's not asynchronous, fire and forget?

rogeralsing:
Fixed in #150

PR #140, opened 2017-04-06
handle exceptions when EventStream.Publish

alexpantyukhin:
#102 
For now I think that it would be okay. Also a test when https://github.com/AsynkronIT/protoactor-dotnet/pull/135 will be merged.

### DIFF
--- a/src/Proto.Actor/EventStream.cs
+++ b/src/Proto.Actor/EventStream.cs
@@ -33,6 +33,8 @@ namespace Proto
         private readonly ConcurrentDictionary<Guid, Action<T>> _subscriptions =
             new ConcurrentDictionary<Guid, Action<T>>();
 
+        private readonly ILogger _logger = Log.CreateLogger<EventStream<T>>();
+
         public Subscription<T> Subscribe(Action<T> action)
         {
             var sub = Guid.NewGuid();
@@ -58,7 +60,14 @@ namespace Proto
         {
             foreach (var sub in _subscriptions)
             {
-                sub.Value(msg);
+                try
+                {
+                    sub.Value(msg);
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogWarning(0, ex, "Exception has occurred when publish msg.");
+                }
             }
         }
 


src/Proto.Actor/EventStream.cs
```diff
@@ -56,9 +56,18 @@ public Subscription<T> Subscribe(Action<T> action)
 
         public void Publish(T msg)
         {
+            var logger = Log.CreateLogger<EventStream>();
```
adamhathcock: The logger should be a member of the class, not local to the method.

alexpantyukhin: ooh. I see. Does it mean that there is the same problem: https://github.com/AsynkronIT/protoactor-dotnet/blob/dev/src/Proto.Actor/EventStream.cs#L19 ?

adamhathcock: It's not the same thing as the logger in the constructor gets hoisted to a member variable.  I probably should have just made the logger a member variable anyway.

alexpantyukhin: Actually EventStream inherits EventStream<T>. And maybe it would be reasonable to have protected logger which was iniated in the constructor of the EventStream<T>. But I think that if you as user inherits EventStream and see protected logger it's a little bit strange..  My suggestion is to have different loggers in EventStream and EventStream<T>. What do you think?

adamhathcock: Loggers shouldn't be inherited.  They're unique to the object and type.  Honestly, there shouldn't be a base class with a logger in them at all.

alexpantyukhin: Got your point. Thx!

alexpantyukhin: @adamhathcock fixed.

adamhathcock:
👍

rogeralsing:
@adamhathcock I remember you saying that the logger is not threadsafe?
EventStream can be invoked from any thread. thus making the logger used from multiple threads if they both get the same issue when pubishing.

Is this correct?

adamhathcock:
@rogeralsing I did say that.  The more I think about it, I must have lied.

The logger is just a lightweight shim to actual implementations and those implementations would have to be threadsafe to accept log events from anywhere.

I do this all the time in my code with situations like this.  It should be fine.

In any case, calling that get logger over and over doesn't actually create a new logger each time.  But it does do some work to get the correct one that is cached.

PR #143, opened 2017-04-07
Dependency Injection Extensions with ASP.NET Core Hosting sample

adamhathcock:
I have `ActorFactory` will get actors in a virtual sense.  It is injected with `IActorFactory`

`ActorPropsRegistry` is used when adding Proto.Actor to `IServiceCollection` to allow lazy creation of actors with specific props.

In the example, `ActorManager` is just supposed to be some random domain class that uses actors.  It uses a named and non-named of the same type to show usage.  

Also, `EventStream<>` is registered to DI as a singleton so it can share it with DI.  This is in the demo too.


adamhathcock:
@AsynkronIT/developers any one else for usage with ASP.NET Core style DI?

adamhathcock:
🎉

PR #144, opened 2017-04-09
Remoting test

tomliversidge:
The PR shows and example of how we might run a remote test. It is based on the RemoteBenchmark example code. 

PR adds:

1. Proto.Remote.Tests.Messages - .proto messages for the tests
2. Proto.Remote.Tests.Node - a console app that acts as a remote node
3. Proto.Remote.Tests - the xunit test project

The test works by launching the Proto.Remote.Tests.Node console app via System.Diagnostics.Process using "dotnet <path to dll>" and then sends a message and awaits a reply. This is used in an xunit fixture so shared amongst tests (note this might have to change if we add tests that start killing the remote process).

I've also added an xunit Trait (Category=Remote) so it can be run separately from the other unit tests as it is more of a system level test - probably best to have CI run unit tests followed by Remote tests

Tested xplatform on win10 and mac, but it would be good to have other people checkout and verify it works.
Requires Proto.Remote.Tests.Node to be built first (solution should do this)

What are people's thoughts on this way of testing Proto.Remote? Any better ideas? 

@cpx86 - what tests need writing for remote?


rogeralsing:
This is brilliant! love it

PR #146, opened 2017-04-09
Hopefully fixing build

tomliversidge:
- using conditional build symbols to determine path for Proto.Remote.Tests.Node dll

- the project is already outputting RELEASE symbols so can use these


cpx86:
:+1:

PR #148, opened 2017-04-09
remove not used dictionary

alexpantyukhin:
It seems that this dictionary is used just for add and remove elements. I don't see reason to support it.


rogeralsing:
Good catch, I saw we have the same in Go too

PR #149, opened 2017-04-09
substitute Dictionary with ConcurrentDictionary

alexpantyukhin:
I believe it should be like this. If it could called with the same `address` 2 times concurent it will raise an exception. This variant is more concurent stable I think.

### DIFF
--- a/src/Proto.Remote/EndpointManager.cs
+++ b/src/Proto.Remote/EndpointManager.cs
@@ -5,7 +5,7 @@
 // -----------------------------------------------------------------------
 
 using System;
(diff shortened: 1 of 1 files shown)


rogeralsing:
There is no need for any concurrent dictionaries here, the code is guarded by the actors concurrency constraint.
There can never be more than one thread mutating the state.

It might however be good to add a similar test to see if the address is already in the dict similar to what you do here.

alexpantyukhin:
Oh, got it, thx. I will close the PR.

PR #150, opened 2017-04-10
Test for EventStream exception handling

### LINKED ISSUE #102: Error handling in EventStream subscriptions

tomliversidge:
Change logging of exceptions from Warning level to Error level
Closes #102


tomliversidge:
ConcurrentDictionary doesnt have order so non-exception handelr might run first

PR #155, opened 2017-04-10
Test for spawning remote actors

tomliversidge:


tests/Proto.Remote.Tests/RemoteTests.cs
```diff
@@ -38,9 +38,25 @@ public class RemoteTests
...
+        [Fact]
+        public async void CanSpawnRemoteActor()
+        {
+            var props = Actor.FromProducer(() => new LocalActor());
+            var pid = Actor.Spawn(props);
+            var remoteActorName = Guid.NewGuid().ToString();
+            var remoteActor = await Remote.SpawnNamedAsync("127.0.0.1:12000", remoteActorName, "remote", TimeSpan.FromSeconds(5));
+            await remoteActor.RequestAsync<Start>(new StartRemote { Sender = pid }, TimeSpan.FromMilliseconds(2500));
+            remoteActor.Tell(new Ping());
+            await Task.Delay(TimeSpan.FromSeconds(2));
```
rogeralsing: why do we need the delay?

rogeralsing:
Great!

tests/Proto.Remote.Tests/RemoteTests.cs
tomliversidge: It's to wait for the remoteActor.Tell to have time to complete. I guess a better test would be to ReceiveAsync on the remote actor...

rogeralsing: Agree, making it use ReceiveAsync would be better and will fail less often on the build server

tomliversidge: https://github.com/AsynkronIT/protoactor-dotnet/pull/158

PR #157, opened 2017-04-10
Allow async subscriptions to EventStream with dispatchers.

adamhathcock:
Added a null dispatcher

### DIFF
--- a/src/Proto.Mailbox/Dispatcher.cs
+++ b/src/Proto.Mailbox/Dispatcher.cs
@@ -27,10 +27,10 @@ namespace Proto.Mailbox
     public static class Dispatchers
     {
         public static ThreadPoolDispatcher DefaultDispatcher { get; } = new ThreadPoolDispatcher();
-        public static NullDispatcher NullDispatcher { get; } = new NullDispatcher();
+        public static SynchronousDispatcher SynchronousDispatcher { get; } = new SynchronousDispatcher();
     }
 
-    public sealed class NullDispatcher : IDispatcher
+    public sealed class SynchronousDispatcher : IDispatcher
     {
         public int Throughput => 300;
 
--- a/src/Proto.Mailbox/Dispatcher.cs
+++ b/src/Proto.Mailbox/Dispatcher.cs
@@ -27,6 +27,17 @@ namespace Proto.Mailbox
     public static class Dispatchers
     {
         public static ThreadPoolDispatcher DefaultDispatcher { get; } = new ThreadPoolDispatcher();
+        public static NullDispatcher NullDispatcher { get; } = new NullDispatcher();
+    }
+
+    public sealed class NullDispatcher : IDispatcher
+    {
+        public int Throughput => 300;
+
+        public void Schedule(Func<Task> runner)
+        {
+            runner().Wait();
+        }
     }
 
     public sealed class ThreadPoolDispatcher : IDispatcher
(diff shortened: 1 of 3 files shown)


src/Proto.Mailbox/Dispatcher.cs
```diff
@@ -27,6 +27,17 @@ public interface IDispatcher
...
+        public static NullDispatcher NullDispatcher { get; } = new NullDispatcher();
+    }
+
+    public sealed class NullDispatcher : IDispatcher
+    {
+        public int Throughput => 300;
+
+        public void Schedule(Func<Task> runner)
+        {
+            runner().Wait();
```
adamhathcock: Not sure I like this but seems like the only thing to do.

adamhathcock:
Maybe using `ThreadPoolDispatcher` as the default is better than the "null" one.

tests/Proto.Actor.Tests/EventStreamTests.cs
```diff
@@ -48,5 +49,18 @@ public void EventStream_OnlyReceiveSubscribedToEventTypes()
...
+
+        [Fact]
+        public void EventStream_CanSubscribeToSpecificEventTypes_Async()
+        {
+            var received = "";
+            var eventStream = new EventStream();
+            eventStream.Subscribe<string>(theString =>
+            {
+                received = theString;
+                Assert.Equal("hello", received);
```
tomliversidge: I don't think this assertion ever gets called - try changing it to "some random string" and it still passes

tomliversidge: Passing in NullDispatcher will fix it as it runs synchronously

tomliversidge:
Would NullDispatcher be better of named something like ImmediateSynchronousDispatcher or BlockingDispatcher? It's not really null

tomliversidge:
Also, what happens now when a subscriber errors / throws an exception?

tomliversidge:
Like the idea of being able to pass in a dispatcher though! 👍

adamhathcock:
`PassThroughDispatcher` ?

Error handling does need to be thought about, especially now as tasks are being dispatched.  I can do that when people give more thumbs up to this direction.

tests/Proto.Actor.Tests/EventStreamTests.cs
adamhathcock: It does get called, well it did when I was debugging.

Obviously, the test will usually complete before as it's non-deterministic.  Wasn't sure how else to demo things.  By default, it does always use the `NullDispatcher`

tomliversidge: public static ThreadPoolDispatcher DefaultDispatcher { get; } = new ThreadPoolDispatcher();

- uses ThreadPoolDispatcher by default, hence the test assertion doesn't get called, as the test completes before the assertion.

tomliversidge:
Hmmm.. what is being passed through, and to where? :) The behaviour is "call the function that returns the task , then wait for the task to complete" - 

I find it confusing that the function that returns the task is called "runner" as isn't it the dispatcher that is doing the running? :) - one dispatcher runs things synchronously, the other runs things asynchronously on the threadpool. alternatively you could argue that the threadpool is the runner :) 🥇 

i guess we have "tasks", that get "dispatched" somewhere to be "run"?

adamhathcock:
`SynchronousDispatcher` ?  Naming is hard :)

adamhathcock:
Renamed and added back try/catch in publish.  I hadn't meant to remove it.

rogeralsing:
My main concern for this PR, is that events could be processed out of order as there is no guarantee that tasks that are scheduled on the threadpool run in order.
This is likely fine for many cases, but we might need to add some sort of heads up docs for it

adamhathcock:
I think the best part is that each subscriber can choose.  

I can add to docs...if I knew how :)

PR #159, opened 2017-04-10
fix bug where _children might be null when handling a terminated message

tomliversidge:
In the included test, _children is null, so the _children.Remove call threw an exception. This PR just checks for null first.


rogeralsing:
👍

PR #160, opened 2017-04-11
add check if address already in the _connections.

alexpantyukhin:
@rogeralsing  . The check that element is already was added into the dictionary. I believe that if element already was added into the _connections then it should be returned. Forgot to add it earlier. 
https://github.com/AsynkronIT/protoactor-dotnet/pull/149

### DIFF
--- a/src/Proto.Remote/EndpointManager.cs
+++ b/src/Proto.Remote/EndpointManager.cs
@@ -87,7 +87,16 @@ namespace Proto.Remote
                 var watcher = SpawnWatcher(address, context);
 
                 endpoint = new Endpoint(writer, watcher);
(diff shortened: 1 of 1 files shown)


rogeralsing:
After looking at the code, this can never happen.
we have this guard block:

```csharp
 var ok = _connections.TryGetValue(address, out var endpoint);
 if (!ok)
```
So there is no chance the same address can be added twice.

Closing

alexpantyukhin:
It's not a lock section. Actually 2 processes can go into the "!ok" block. But it has very small possibility.

rogeralsing:
No, it's inside an actor so that is impossible

PR #161, opened 2017-04-11
Remote watch test

tomliversidge:
The current code in RemoteProcess.Send causes an infinite loop when watching a remote actor. Before removing the code in this PR, I was seeing 10,000s of RemoteWatch messages until the test timed out - RemoteProcess was passing the RemoteWatch message back to the EndpointManager, which then passed it to the RemoteProcess, which then passed it back to the EndpointManager...... and on.

### DIFF
--- a/src/Proto.Remote/EndpointWatcher.cs
+++ b/src/Proto.Remote/EndpointWatcher.cs
@@ -57,8 +57,7 @@ namespace Proto.Remote
                     _watched[msg.Watcher.Id] = null;
 
                     var w = new Unwatch(msg.Watcher);
-                    msg.Watchee.SendSystemMessage(w);
-
+                    RemoteProcess.SendRemoteMessage(msg.Watchee, w);
                     break;
                 }
                 case RemoteWatch msg:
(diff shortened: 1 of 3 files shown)


tomliversidge:
img_2339

tomliversidge:
Looks like the same code is in Go version too. https://github.com/AsynkronIT/protoactor-go/blob/dev/remote/remote_process.go#L42

I'm not sure if just deleting this code is the correct fix :) - presumably it was there for a reason?

rogeralsing:
The code is there to make sure watch and unwatch is proxied through the EndpointWatcher, so that the EndpointWatcher knows what remote actors are being watched.
So if the endpoint dies, the endpoint watcher can notify the watchers that the watched actors terminated.

tomliversidge:
I think the EndpointWatcher still gets sent the RemoteWatch message with this RemoteProcess code deleted - it happens at the start of what was the infinite loop...

tomliversidge:
Also in this test the Terminated message is received twice in the local actor

tomliversidge:
@rogeralsing I've updated the code to send a remote message after handling the RemoteWatch/RemoteUnwatch events

src/Proto.Remote/EndpointWatcher.cs
```diff
@@ -57,17 +57,15 @@ public Task ReceiveAsync(IContext context)
                     _watched[msg.Watcher.Id] = null;
 
                     var w = new Unwatch(msg.Watcher);
-                    msg.Watchee.SendSystemMessage(w);
-
+                    RemoteProcess.SendRemoteMessage(msg.Watchee, w);
```
rogeralsing: 👍

PR #165, opened 2017-04-12
Refactor sender middleware

rogeralsing:
This PR fixes some issues in the outbound middleware design.
Now it is possible for sender middleware to read current message, current message headers and modify outbound message headers.
This is one step closer to full tracing support.

This is related to #127


examples/Middleware/Program.cs
```diff
@@ -5,47 +5,49 @@
...
+                             {
+                                 Console.WriteLine($"middleware 2 enter {c.Message.GetType()}:{c.Message}");
+                                 await next(c);
+                                 Console.WriteLine($"middleware 2 exit {c.Message.GetType()}:{c.Message}");
+                             });
 
         var pid = Actor.Spawn(actor);
-        pid.Tell("hello");
+
+        var root = new RootContext(new MessageHeader(), next => async (c, target, envelope) =>
```
rogeralsing: Outbound middleware is set on the root context, so when this example Tells the actor, the middleware is activated for the message.
Inbound middleware in the actor then reacts to the incoming message.

src/Proto.Actor/Actor.cs
```diff
@@ -11,7 +11,7 @@ namespace Proto
 {
     public delegate Task Receive(IContext context);
 
-    public delegate Task Sender(IContext ctx, PID target, MessageEnvelope envelope);
+    public delegate Task Sender(IOutboundContext ctx, PID target, MessageEnvelope envelope);
```
rogeralsing: Sender now have access to an IOutboundContext, this is because in the case of a RootContext, there is no full actor context, there is only a partial context.

You only have access to Message, MessageHeader here.

adamhathcock:
Seems awkward.  So this puts the burden on each caller of the actor to setup the sender middleware instead of having it part of the props on the actor?

rogeralsing:
No, the actor can also have it, but that is for messages going out of the actor.

If you have tracing, the caller owns the context.
Eg your rest endpoint need to extract tracing info and pass forward

src/Proto.Actor/MessageHeader.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Collections.Generic;
+
+namespace Proto
+{
+    public class MessageHeader : Dictionary<string, string>
+    {
+        internal static readonly MessageHeader EmotyHeader = new MessageHeader();
```
tomliversidge: Typo? EmotyHeader

adamhathcock:
I guess, thinking about it more, all you need to do is have an extension method that creates the root context and overload `Tell` to take the sender middleware 👍

rogeralsing:
"extension method" on what?

We are going from the outside world into the actor world, and we want to propagate contextual information, e.g. tracing values.

```
[Console app]  --- tell --> [Actor]
[Root Context] -----------> [Actor Context]
[Sender Middleware] ------> [Receive Middleware]
[Message Headers Out] ----> [Message Header In]
```

rogeralsing:
I've applied some refactoring on the API level.

`RootContext` -> `ActorClient`
`Context.MessageHeader` -> `Context.Headers`

RootContext felt a bit confusion and didn't quite communicate what it was.
ActorClient is what it sounds like, it is a client to interact with Actors.
The reason those exists, is if you want to apply middleware and headers between your non-actor code into the world of actors.

adamhathcock:
I mean, couldn't you have an extension method off of `PID` that looked something like: `public static void Tell(this PID pid, object message, Func<Sender, Sender> middleware)` and inside that method create the `ActorClient` and send?

i guess you'd usually want to reuse Actor clients.  Can I abuse message headers to have an expected reply message type? :)

rogeralsing:
>Can I abuse message headers to have an expected reply message type?

Headers are much like HTTP headers, just KVP of strings. I guess you _could_ pass the expected response type name that way. 
But then again, that would probably just be easier to just do that in the message itself.

rogeralsing:
btw. is there anything that needs to be done here before merging?

Issue #162, opened 2017-04-11
EndpointManager does not subscribe to EndpointTerminatedEvent


rogeralsing:
see https://github.com/AsynkronIT/protoactor-go/blob/dev/remote/endpoint_manager.go#L19
this part is responsible for notifying the endpointmanager when a node is no longer available.

tomliversidge:
I need to fix this as part of a test I'm writing for killing a remote node so I'll take this

PR #166, opened 2017-04-12
add test for notification on remote node killed

tomliversidge:
fixes https://github.com/AsynkronIT/protoactor-dotnet/issues/162

refactor xunit fixture RemoteManager to allow management of multiple remote nodes

refactor remote node console app to accept host and port as parameters to allow launching on different endpoints


src/Proto.Remote/Remote.cs
```diff
@@ -72,11 +72,9 @@ private static void SpawnEndpointManager(RemoteConfig config)
         {
             var props = Actor.FromProducer(() => new EndpointManager(config));
             EndpointManagerPid = Actor.Spawn(props);
+            EventStream.Instance.Subscribe<EndpointTerminatedEvent>(EndpointManagerPid.Tell);
         }
```
tomliversidge: Wasn't sure if this subscription needs unsubscribing somewhere.....

rogeralsing: In a real world scenario it is unlikely that unsubscribing is needed.
In a test scenario where we run many tests, it might be, maybe we should add a Remote.Shutdown() just to deal with this kind of scenarios (?)

tomliversidge: Lets add it when there becomes a need?

rogeralsing: Sure

Issue #167, opened 2017-04-13
Replace Console.Writeline to Logger


rogeralsing:
There are still a few places in the code that uses Console.WriteLine, we should replace this with proper logging code.

alexpantyukhin:
Some of them:

https://github.com/AsynkronIT/protoactor-dotnet/blob/0321d6acb066f1ffca439db78bec71f7e338da0a/src/Proto.Remote/EndpointWriter.cs#L101
https://github.com/AsynkronIT/protoactor-dotnet/blob/0321d6acb066f1ffca439db78bec71f7e338da0a/src/Proto.Remote/EndpointWriter.cs#L118
https://github.com/AsynkronIT/protoactor-dotnet/blob/577ddff5fb33bd7b441bffdb50ace916923b1281/src/Proto.Cluster/Cluster.cs#L30
https://github.com/AsynkronIT/protoactor-dotnet/blob/0321d6acb066f1ffca439db78bec71f7e338da0a/src/Proto.Remote/EndpointManager.cs#L41
https://github.com/AsynkronIT/protoactor-dotnet/blob/577ddff5fb33bd7b441bffdb50ace916923b1281/src/Proto.Cluster/MemberListActor.cs#L24
https://github.com/AsynkronIT/protoactor-dotnet/blob/cb680772049d850f14abeb2fc7a0d25eff34e2ad/src/Proto.Remote/Remote.cs#L62
https://github.com/AsynkronIT/protoactor-dotnet/blob/577ddff5fb33bd7b441bffdb50ace916923b1281/src/Proto.Cluster/Partition.cs#L72

adamhathcock:
Looking into this as I'm avoiding some work.

Should some of the statements just be deleted?  Like the ones that are just "X Started" ?

rogeralsing:
I still think they add some value on start up of cluster and remoting, as they make it easy to see where in the start up flow something broke. maybe just do them as debug

rogeralsing:
And should we just have a shared logger per module?
e.g.
Proto.Actor.Logger
Proto.Cluster.Logger

instead of creating different ones all over the place?

adamhathcock:
Each logger is contextual to the class using it.  Basically, it uses the class name as part of the context of the log event.  Loggers themselves are really lightweight but there is one made per instance (unless no logging is set).

The Cluster and Remote static classes are getting a static logger each though.

tomliversidge:
[...]

PR #168, opened 2017-04-13
Replace console.writeline with loggers naively.

adamhathcock:
This for https://github.com/AsynkronIT/protoactor-dotnet/issues/167

### DIFF
--- a/src/Proto.Cluster/PidCache.cs
+++ b/src/Proto.Cluster/PidCache.cs
@@ -7,6 +7,7 @@
 using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
 using Proto.Remote;
 
 namespace Proto.Cluster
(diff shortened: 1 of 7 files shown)


src/Proto.Cluster/PidCache.cs
```diff
@@ -53,7 +55,7 @@ public Task ReceiveAsync(IContext context)
             switch (context.Message)
             {
                 case Started _:
-                    //  Console.WriteLine("Started PidCacheActor");
+                    _logger.LogDebug("Started PidCacheActor");
```
rogeralsing: This one is used behind a router, so we get 100 of these, which might be a bit too much :-)

src/Proto.Remote/EndpointManager.cs
```diff
@@ -38,7 +41,7 @@ public Task ReceiveAsync(IContext context)
             {
                 case Started _:
                 {
-                    Console.WriteLine("[REMOTING] Started EndpointManager");
+                    _logger.LogDebug("Started EndpointManager");
                     return Actor.Done;
```
tomliversidge: I actually liked this [REMOTING] tag, as it is useful in tests to see where a log was being written. i.e. in tests where remote nodes are provisioned (even better would be some more info about _which_ remote node)

rogeralsing: Let's address cosmetics as separate PR's

PR #169, opened 2017-04-14
Remote test additions

tomliversidge:
- add test for unwatching remote actor
- add test for watching multiple remote actors
- add test for multiple local actors watch one remote
- add assertion on killing remote node that only a single Terminated message is received. 
- add tests for registration of remote kinds

Refactorings:

- removed duplication around starting local and remote actors
- local actor now keeps a record of all Terminated messages it has received and use this to satisfy "did the actor receive a Terminated message for this remote actor" queries 
- better console logging. This is primarily because there are a few delays in the code to allow propagation of messages. Adding logging around this helps to understand what's going on when reading the test output
- added a polling mechanism for asserting on conditions where we cannot just using RecevieAsync. This should result in faster and more robust test execution

Test output should look like:

capture


rogeralsing:
Awesome work!

PR #173, opened 2017-04-15
fix for perstistence context change

raskolnikoov:


tomliversidge:
In other words, all PersistedEvent messages (aka state changes) need to be guaranteed to be applied before the next message is processed

raskolnikoov:
ah I see.. so how can we achive that?

raskolnikoov:
is it possible to set priority in the mailbox so these messages will be processed first?

raskolnikoov:
or will this PR https://github.com/AsynkronIT/protoactor-dotnet/pull/172 fix this issue so the persistence api dosent need any changes?

tomliversidge:
I wouldn't even go through the mailbox - if you see the option I provided in the issue it is just calling a new method I added to the IPersistantActor interface. The advantage is all state changes are processed synchronously from within the context of the received message.

rogeralsing:
Agree with @tomliversidge, going through the mailbox is begging for issues.
e.g. if the replay fails, restarts and now some messages are left in the mailbox.
as we support async await we can safely do the init process as a sync flow

Issue #171, opened 2017-04-14
Context changes during message processing after calling Persistence API


tomliversidge:
For example:

```
public async Task ReceiveAsync(IContext context)
        {
            switch (context.Message)
            {
                case Multiply msg:
                    // context.Message is Multiply
                    await Persistence.PersistEventAsync(new Multiplied { Amount = msg.Amount });
                    // context.Message is now PersistedEvent
                    break;
            }
        }
```

It has also been reported that context.Sender changes too.

tomliversidge:
Presumably https://github.com/AsynkronIT/protoactor-dotnet/blob/dev/src/Proto.Persistence/Persistence.cs#L48 is what causes the context to change

raskolnikoov:
Same goes for PersistSnapshotAsync too I guess?

rogeralsing:
In Go, the persistence plugin does not go through the ReceiveAsync method, it simply picks the current behavior and calls that. meaning the message being processed will not be the same as the Message property. (which is not completely great either)

What we could do, is to let the persistence provider read the current (raw) message, and then re-set it after the persist* call have completed.

I also saw we have a small oddity in ReceiveAsync where it sets the current Message through the setter of that property, while the getter property unwraps potential MessageEnvelopes.

Imo, we should get rid of the setter and use the raw _message field instead as that is agnostic to raw or envelope wrapped messages.

tomliversidge:
How about adding an UpdateState method onto the IPersistentActor interface, then just calling this:

https://github.com/AsynkronIT/protoactor-dotnet/compare/dev...tomliversidge:persistence_refactor?expand=1

Keeps it simple :)

raskolnikoov:
I'll have a look at this in the weekend :)

raskolnikoov:
Now we have two things to solve this issue https://github.com/AsynkronIT/protoactor-dotnet/pull/175

Issue #137, opened 2017-04-06
Persistence differences between Go and .net APIs plus other things


tomliversidge:
The Go persistence API and the .net persistence API are different. 

The Go API looks like

```
        Restart()
	GetSnapshotInterval() int
	GetSnapshot(actorName string) (snapshot interface{}, eventIndex int, ok bool)
	GetEvents(actorName string, eventIndexStart int, callback func(e interface{}))
	PersistEvent(actorName string, eventIndex int, event proto.Message)
	PersistSnapshot(actorName string, eventIndex int, snapshot proto.Message)
```

The .net one:

```
        Task<(object Snapshot, long Index)> GetSnapshotAsync(string actorName);
        Task PersistSnapshotAsync(string actorName, long index, object snapshot);
        Task DeleteSnapshotsAsync(string actorName, long inclusiveToIndex);
        Task GetEventsAsync(string actorName, long indexStart, Action<object> callback);
        Task PersistEventAsync(string actorName, long index, object @event);
        Task DeleteEventsAsync(string actorName, long inclusiveToIndex);
```

Obvious differences are
1. You can delete events and snapshots in .net API
2. You can restart in Go API
3. You can get the Snapshot interval in Go API

There are a couple of other things I'd like to see in both:

1. allow replaying up to a certain event index
2. allow replaying up to a certain point in time
3. separate out the stores for snapshots and events see https://github.com/AsynkronIT/protoactor-dotnet/pull/136 for example

What are people's thoughts on this?

rogeralsing:
As there are a lot of traction in the .NET side of things now, I think we could drive the API design for persistence here and then backport it to Go

raskolnikoov:
I think that if you want to mix Redis and EventStore for snapshots/events just build a plugin that do that in the same provider instead of splitting them into two. Other way would be to introduce another persistence api, SnapshotPersistence/EventPersistence api. So If you don't wanna store the data in the same storage you go for two middleware providers to your actor. Don't really have an idea how to implement this in the actor then, since when using Persistence api you use a IPersistenceActor... And do we really wanna introduce more "IActors"?

tomliversidge:
I'd like it to work where you just say 'I want to use provider1 for events and provider2 for snapshots'  where provider1 and provider2 are the existing implementations. There's no way of doing that with the current API as it doesn't distinguish between the two things. I don't think a new persistence API is necessary either.

[...]

PR #174, opened 2017-04-16
Persistence refactor

tomliversidge:
Fixes https://github.com/AsynkronIT/protoactor-dotnet/issues/171

- Add UpdateState method to IPersistentActor
- Added base types for Event and Snapshot so that if you don't care about PersistedEvent vs RecoverEvent or PersistedSnapshot vs RecoveredSnapshot you don't have to - you can just look for Event and Snapshot 
- Updated example and tests

NOTE: I considered having the IPersistentActor interface look like

```
public interface IPersistentActor : IActor
     {		     
         {
             Persistence Persistence { get; set; }		         
             void Apply(Event @event)
             void Apply(Snapshot snapshot);
         }		     
     }
```

but didn't want to force consumers to implement Apply(Snapshot snapshot) as snapshots are not always used.

### DIFF
--- a/examples/Persistence/Persistence/Program.cs
+++ b/examples/Persistence/Persistence/Program.cs
@@ -10,6 +10,8 @@ using Messages;
 using Proto;
 using Proto.Persistence;
 using Proto.Persistence.Sqlite;
+using Event = Proto.Persistence.Event;
+using Snapshot = Proto.Persistence.Snapshot;
 
 class Program
 {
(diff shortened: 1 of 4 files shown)


rogeralsing:
I like it.
Having the UpdateState clearly separates the state mutations from the actor receive behavior.

PS. there is a valid concern on deltas and snapshots here:
https://github.com/AsynkronIT/protoactor-dotnet/issues/137#issuecomment-294317314

But I do like the general direction here.

tomliversidge:
Yeah there is another PR that actually separates snapshots and events https://github.com/AsynkronIT/protoactor-dotnet/pull/136

raskolnikoov:
Haha I worked on the same issue this morning too.. Here is my PR https://github.com/AsynkronIT/protoactor-dotnet/pull/175 . It works kinda the same way, but uses eventhandlers instead.

PR #175, opened 2017-04-16
persistence event handlers

raskolnikoov:
- Add EventHandlers to Persistence
- Updated example and tests

### DIFF
--- a/examples/Persistence/Persistence/Program.cs
+++ b/examples/Persistence/Persistence/Program.cs
@@ -41,12 +41,16 @@ class Program
             {
                 case Started msg:
 
+                    RegisterHandlers();
+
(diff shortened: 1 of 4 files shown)


tomliversidge:
I have a biased opinion, but I think the approach in https://github.com/AsynkronIT/protoactor-dotnet/pull/174 is simpler - there are no events, tasks and delegates to hook up, it's just a synchronous method call - I think this makes it easier for someone to implement.

rogeralsing:
I give my vote for @tomliversidge solution, simplicity is key for long term maintenance.
Are there any clear benefits to the other solution that would be hard or impossible to do with the simpler approach that makes it favorable?

raskolnikoov:
Yes PR #174 is simpler. I'll close this in favor for the other.

PR #177, opened 2017-04-16
fixed persistence example

raskolnikoov:
fixes the example


src/Proto.Persistence/Persistence.cs
```diff
@@ -54,6 +54,7 @@ public class Persistence
         {
             var index = Index;
             await _state.PersistSnapshotAsync(ActorId, index, snapshot);
+            _actor.UpdateState(new PersistedSnapshot(snapshot, index));
         }
```
tomliversidge: This shouldn't be called here - you're taking a snapshot, not applying it.

PR #178, opened 2017-04-16
removed RecoveryStarted and Completed messages due to un-needed

raskolnikoov:
Due to new changes in the api (updatestate) we don't need to send messages for recoverystarted and recoverycompleted anymore.

### DIFF
--- a/examples/Persistence/Persistence/Program.cs
+++ b/examples/Persistence/Persistence/Program.cs
@@ -12,6 +12,7 @@ using Proto.Persistence;
 using Proto.Persistence.Sqlite;
 using Event = Proto.Persistence.Event;
 using Snapshot = Proto.Persistence.Snapshot;
+using System.Text;
 
 class Program
 {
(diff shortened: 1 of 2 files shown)


raskolnikoov:
Forgot to remove the messagetypes and cleanup example. Im gonna fix that when im home again.

tomliversidge:
Do we not need them now? Can an actor receive new messages during recovery?

raskolnikoov:
No the actor gets its first message (Started) after the recovery is finished.

raskolnikoov:
Updated PR with some more missing cleanups in first commit.

tomliversidge:
I'm a bit confused as to why the recent change has meant these events now need deleting - is it just that they are not useful / needed? The recent API change was from await ReceiveAsync to _actor.UpdateState, both being an essentially synchronous workflow, so I'm confused why this has affected RecoveryStarted and RecoveryCompleted

raskolnikoov:
That's because RecoveryStarted/RecoveryCompleted used ReceiveAsync method and that messed up the context so the actor never got the "Started" message upon startup. They aren't needed since when the recover starts UpdateState is called with recovery data and awaited, then Startup gets passed to the actor and it should then have started with the "last state" it had upon shutdown/crash/earhquake.

tomliversidge:
Ahh.. ok thanks for the explanation!

PR #181, opened 2017-04-16
Remove call to _actor.UpdateState on taking a snapshot and update exa…

tomliversidge:
…mple app

- we shouldn't be calling _actor.UpdateState(new PersistedSnapshot(snapshot, index)) after taking a snapshot, as there is nothing to update

- I updated the example app to write to the console after taking a snapshot to preserve existing behaviour

### DIFF
--- a/examples/Persistence/Persistence/Program.cs
+++ b/examples/Persistence/Persistence/Program.cs
@@ -70,9 +70,6 @@ class Program
                         Console.WriteLine("MyPersistenceActor - RecoverSnapshot = Snapshot.Index = {0}, Snapshot.State = {1}", Persistence.Index, ss.Name);
                     }
                     break;
-                case PersistedSnapshot msg:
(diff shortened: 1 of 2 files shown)


rogeralsing:
👍

raskolnikoov:
the PersistedSnapshot message can be used to aknowledge that the snapshot passed. Why can't it be there?

tomliversidge:
Well it doesn't really make sense with the UpdateState function. You're saying "save the current state of the actor" followed by "now update the state of the actor with the same state we just saved" ... but it hasn't changed - it's the same state, so there's nothing to update, hence why call the function? If you need acknowledgement that the operation succeeded you can always wrap `await _persistence.PersistSnapshotAsync` in a try-catch, or use a continuation, as if it fails the provider should throw an exception. 

The code inside of UpdateState should be only concerned with updating the state of the actor - it should _never_ have other side effects.

raskolnikoov:
Yes u are right

PR #185, opened 2017-04-20
Extract behavior

rogeralsing:
This PR extracts the become/unbecome functionality into a separate object.


src/Proto.Actor/Behavior.cs
```diff
@@ -0,0 +1,45 @@
...
+//       Copyright (C) 2015-2017 Asynkron HB All rights reserved
+//   </copyright>
+// -----------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Proto
+{
+    public class Behavior
```
rogeralsing: The new behavior class

src/Proto.Actor/Behavior.cs
```diff
@@ -0,0 +1,45 @@
...
+            Become(receive);
+        }
+
+        public void Become(Receive receive)
+        {
+            _behaviors.Clear();
+            _behaviors.Push(receive);
+        }
+
+        public void BecomeStacked(Receive receive)
```
rogeralsing: Has both Become and BecomeStacked methods

tests/Proto.Actor.Tests/PopPushBehaviorTests.cs
```diff
@@ -13,21 +10,23 @@ public class PopPushBehaviorTests
         [Fact]
         public async Task PopBehaviorShouldRestorePushedBehavior()
         {
-            PID pid = SpawnActorFromFunc(ctx =>
+            var behavior = new Behavior();
+            behavior.Become(ctx =>
```
rogeralsing: Updated test to use the new Behavior class

tomliversidge:
Looks good to me!

Issue #176, opened 2017-04-16
Reducing API surface area


rogeralsing:
I have been thinking about the API.
We do have some features that were just copied from Akka.

e.g.
Stashing and Stackable behaviors.

Do we really need those to be in the core API or should we turn those into pluggable receive handlers?
The LocalContext is somewhat huge and reducing complexity would be nice.

adamhathcock:
I would say remove them.  

I don't get the stackable behavior thing.  Seems overly complex to reuse the same actor.  Besides using something like Stateless is nicer (https://github.com/dotnet-state-machine/stateless) to control behavior imo.

Stashing: dump it too.

tomliversidge:
Funny, Become / UnBecome is one of my favourite features of Akka :) How would this work as a receive handler?

If something like Stashing is taken out and implemented as receive middleware, would it work the same as Persistence? i.e. an Actor would implement something like IHaveAStash and the have a Stash property or something? I like this plug-in approach.

rogeralsing:
@tomliversidge the simplest way would be to just inject a `Behavior` instance to the actor:

```csharp
public class MyActor : IActor
{
     private readonly Behavior _behavior;
     public MyActor(Behavior behavior)
     {
          _behavior = behavior;
          behavior.Become(Behavior1);
     }
     public Task ReceiveAsync(IContext context)
     {
           return _behavior.ReceiveAsync(context);
     }

     public Task Behavior1(IContext context)
     {
         ....
     }

     public Task Behavior2(IContext context)
     {
           ....
     }
}
```

This way, behaviors would just become an injectable dependency

rogeralsing:
cc @AsynkronIT/developers are we OK with the change to behaviors?
Extracting become/unbecome into a separate class that can be used from within the actor itself as shown above.
It has a very nice effect on spawning performance and reduces the amount of random stuff the Context needs to care about.

rogeralsing:
I would also like to extract Stashing and ReceiveTimeouts from the context and make them optional.

tomliversidge:
My only concern is how to use things together I.e. If you want both behaviour and stack. Is it easy to combine?

rogeralsing:
@tomliversidge https://github.com/AsynkronIT/protoactor-dotnet/pull/185
This covers the extraction of the behavior support.
Few actors use it, so it makes sense to make it opt-in so not everyone have to carry the extra weight.

[...]

PR #191, opened 2017-04-21
TO DISCUSS (https://github.com/AsynkronIT/protoactor-dotnet/issues/176) removing Persistence as middleware

tomliversidge:
Shows how life will look if Persistence is not middleware... This came up in https://github.com/AsynkronIT/protoactor-dotnet/issues/176 

NOTE the first commit reproduces existing behaviour, the second is an alternative option where I split UpdateState into Apply(Event event) and Apply(Snapshot snapshot). I _think_ I prefer this as it means your code isn't trying to recover from a snapshot if you aren't using them..


adamhathcock:
Could persistence be a feature of the props system?  

While I like the idea of removing features to make the core simple, if everything is loose then there's no guidance on how things like persistence ought to be.  

Basically, here it looks like you can do anything and there's little guidance as you have to implement the persistence messages yourself on each actor.  That kind of sucks.  

I want to fit together the blocks of things I need and only those, but I don't want to have to reimplement the features each time especially not on each actor.

tomliversidge:
I think the differences per actor are just that instead of inheriting from an interface (IPersistentActor) you have to:
1. inject IProvider in the ctor 
2. call _persistence.InitAsync when the Started message is received

1. is no different from having to say UsingMiddleware(Persistence.Using(provider))
2. could be better as I imagine people will forget. 

> you have to implement the persistence messages yourself on each actor
@adamhathcock which messages?

adamhathcock:
I guess I misread about the messages.  Sorry.

So as an implementor: I have to inject the persistence.  Initialize it.  And call when I want to persist for events and snapshots?  I'm just stating this to see how it compares with other systems.

Maybe I'm just worrying over nothing.  I just don't want to go overboard with refactoring features so that users have to do a ton of stuff to wire things in.

tomliversidge:
An alternative could be to go back to using interfaces to guide the user to implement the correct behavior, where we could also have separate IEvents, ISnapshots interfaces that enforce Apply(Event @event) and Apply(Snapshot snapshot) methods which are then called inside the Persistence class

Or, could split up InitAsync into separate methods for events and snapshots

```
case Started _:
                    _persistence.Init(context);
                    await _persistence.RecoverFromSnapshotAsync(Apply); // if using snapshots
                    await _persistence.RecoverFromEventsAsync(Apply); // if using event sourcing
                    break;
```

tomliversidge:
> So as an implementor: I have to inject the persistence. Initialize it. 

Yes, in the "not middleware" model. If done as middleware, this is taken care of for you.

> And call when I want to persist for events and snapshots? 

Yes, but that's how it was working since before i got involved! :) I raised a separate issue to add in support for persistence strategies.

rogeralsing:
Do the persistence layer really need the icontext ref now.
Quick look and IT seems se dont use it any more(?)
IF that is the case, we can init in ctor.

rogeralsing:
Sorry for weird caps, iPhone autocomplete

tomliversidge:
@rogeralsing it is only used to get the id of the actor (_context.Self.Id). This could always be passed in to the various methods on Persistence, i.e. SaveEventAsync(**id**, event).

src/Proto.Persistence/Persistence.cs
```diff
@@ -11,38 +11,51 @@ namespace Proto.Persistence
...
     {
-        private IProviderState _state;
-        private IPersistentActor _actor;
+        private readonly IProviderState _state;
+        private Action<Event> _applyEvent;
+        private Action<Snapshot> _applySnapshot;
+
         public long Index { get; private set; }
         private IContext _context;
         private string ActorId => _context.Self.Id;
```
tomliversidge: @rogeralsing here

raskolnikoov:
I like this idea of removing it from the middleware. Id of the actor could be passed on the Init method. Dont need to ve callend on each save of event/snapshot.

rogeralsing:
Would it be safe to assume that persistence id's should be explicit and never relate to pid?

tomliversidge:
Hmmm good point, yes it should :) The ID needs to be something unique for each actor that survives restarts. It should also be unique across clusters. Basically a GUID. Then the first time the actor is created we should create the ID, any subsequent time it is loaded it should be loaded by this ID

cpx86:
Whichever option we choose, I think it might make sense to go all the way. Currently we're in a bit of a mixed situation where some concerns are handled implicitly by Proto.Persistence, but some concerns need to be handled explicitly by each actor. Both approaches have their pros and cons:
* Middleware + property injection - Pros: The actor itself doesn't need to contain implementation details relating to the persistence. Persistence details can be made declarative so the actor only contains core logic. Cons: A declarative style risks being hard to follow and the user loses control.
* No middleware + ctor injection - Pros: Persistence is made an explicit concern and the user is in full control. No magic interface is required to be implemented by the actor. Also there's no risk of unexpected behaviors due to out-of-order middlewares. Cons: Persistent actors will contain lots of details related to persistence, increasing the risk that the user messes something up, and also making the core logic of the actor less visible. (One could argue though that actors shouldn't contain application logic, but only be used as a mechanism for improving concurrency/eliminating contention)

In the end it feels like it's the old question of either making things convenient for the user, but with the risk of making things hard to understand, vs giving full control to the user, but with the risk of making things complex to implement.

eddier1289:
Hey guys, I just started using Proto.Actor for a new project at work and I'm really liking it so far. One of the things I like about it is the lack of "magic" in Proto.Actor as compared to Akka.net.

I would personally prefer the **no middleware + ctor injection** approach.

So that's my 2 cents for whatever it's worth.

tomliversidge:
@raskolnikoov 

> Id of the actor could be passed on the Init method. Dont need to ve callend on each save of event/snapshot.

I only mentioned this option as @rogeralsing asked if we could call Init in the ctor, rather than waiting for the Started message and then calling Init. It's my understanding that the Id of the actor as it is currently implemented (i.e. PID) is not available until we receive the Started message, so we cannot call Init (as it is currently written) in the ctor.

rogeralsing:
>currently implemented (i.e. PID) is not available until we receive the Started message, so we cannot call Init (as it is currently written) in the ctor.

I think we should get rid of the PID for the persistence ID.
It makes no sense to use the PID as the ID would be dependant on what parent you have or what auto ID you get.
it would make more sense if you have to explicitly set the persistence ID per actor.

tomliversidge:
I just added a commit that initialises the actor state in the ctor, but I think it would be better done when the Started message is received as this method is async

rogeralsing:
>I just added a commit that initialises the actor state in the ctor, but I think it would be better done when the Started message is received as this method is async

Agree, we need to separate the init flow from the creation of the actor instance. otherwise it will block the parent actor that spawns it

tomliversidge:
OK, I've created a couple of other PRs with 2 options for removing Persistence as middleware.... we should just pick one and merge it and delete the other - they're pretty similar, mainly differing in how you create the Persistence class and use it.

I'll close this one

PR #194, opened 2017-04-22
add simple fsm example

alexpantyukhin:
Just small example for demonstrating fsm.

### DIFF
--- a/ProtoActor.sln
+++ b/ProtoActor.sln
@@ -140,6 +140,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Contrib", "Contrib", "{F28F
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DependencyInjection", "examples\DependencyInjection\DependencyInjection.csproj", "{506C2213-CAF6-443E-88C1-B49374AFEE20}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "FSM", "examples\FSM\FSM.csproj", "{AA1AB7FD-79E0-4F17-BD17-1EE81158B288}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
--- a/examples/FSM/Event.cs
+++ b/examples/FSM/Event.cs
@@ -1,0 +1,31 @@
+namespace FSMExample
+{
+    public class Event<TD>
+    {
+        public Event(object fsmEvent, TD stateData)
+        {
+            StateData = stateData;
+            FsmEvent = fsmEvent;
+        }
+
+        public object FsmEvent { get; private set; }
+
+        public TD StateData { get; private set; }
+    }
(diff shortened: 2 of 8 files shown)


cpx86:
Cool stuff! I wonder though if maybe we should separate this example from the others. The examples we have so far are mainly intended as an executable documentation of protoactor features, whereas this is more of an example of how an FSM pattern can be implemented using actors.

E.g. perhaps the `examples` directory could be split into something like:
`examples/features`
`examples/benchmarks`
`examples/patterns`

@AsynkronIT/developers thoughts?

alexpantyukhin:
@cpx86 I agree with your suggestion about splitting examples types.

alexpantyukhin:
The example is based on the akka.net implementation and it is simplfied a little. I think I can simplify it more. But now it is working prototype :)

cpx86:
Would be nice to split the file as well - it was a bit hard to read the diff tbh 😄

alexpantyukhin:
done :) cleaned up a little.

alexpantyukhin:
@cpx86 I thought a little and got that benchmarks actually are not examples.
What if
examples/features
   Chat
   ClusterHelloWorld
   DependencyInjection
   Futures
   HelloWorld
   Labb
   LifecycleEvents
   Middleware
   Persistence
   ReceiveTimeout
   RemoteActivate
   Router
   Schedulers/SimpleScheduler
   Supervision

benchmarks
   InprocessBenchmark
   MailboxBenchmark
   RemoteBenchmark
   SpawnBenchmark

examples/patterns
   FSM

rogeralsing:
While I do think FSM DSL's can be useful, I would like to bring this one up on the design table and see if we can make this more C# 7 idiomatic and how it fits with the ideas around Proto.Actor.

Thoughts @AsynkronIT/developers ?

adamhathcock:
I would rather see how something like Stateless would be used instead of making something new.  I've liked using it.  Maybe a tighter integration could be done.

https://github.com/dotnet-state-machine/stateless/

rogeralsing:
Does Stateless allow for event triggers, e.g. timeouts or how can we integrate something like this in the actor world?

But I agree, that is exactly the Proto.Actor way to go about things. re-use proven building blocks

adamhathcock:
You can definitely trigger events on state transitions. Not sure what you mean about timeouts.

I might get some time tomorrow to implement a FSM example.

rogeralsing:
In Akka, you can set timeouts for states, so that if you have been in state X for Y amount of time, then you should revert to some other state.

adamhathcock:
Actually, when attempting to use Stateless, I'm basically having to recreate what's going on in this PR anyway.

Stateless's main usefulness, it seems, is to just have a graph of validate states so the current state is tracked and allows a transition to happen or not.

Most of the code in this PR is doing actor specific things.  Basically, it wasn't adding much as the state handlers could not be attached and then you add the burden of the user having to declare what state transitions are allowed (well this could be optional).

So, in conclusion.  Users might want to layer Stateless on top of this FSM for validation purposes but it doesn't add anything that I can see.

alexpantyukhin:
Need to close this PR?

PR #200, opened 2017-04-26
Remove Persistence as middleware option 1

tomliversidge:
As discussed in https://github.com/AsynkronIT/protoactor-dotnet/pull/191 this is an option for removing Persistence as middleware. There is another option here https://github.com/AsynkronIT/protoactor-dotnet/pull/199 - PLEASE REVIEW AND MERGE ONLY ONE! :)

Advantages:

- Static creation methods on Persistence enforce using either event sourcing, snapshotting or both
- ApplySnapshot and ApplyEvent are cached in Persistence
- Avoids unneccessary calls to db
- Removes coupling to PID - now need to pass in a persistenceId

Disadvantages:

- Need to call RecoverStateAsync on Started message
- More work to do to create Persistence (i.e. have to pass in IProvider and persistenceId in actor ctor)


raskolnikoov:
Great! Ill have a look at both tomrorrow!

rogeralsing:
In this example, where is the RecoverAsync called?

tomliversidge:
It would need calling when you receive the Started message

rogeralsing:
Agree with @raskolnikoov , I really like this version too.
Clean and explicit on what is going on. 👍

tomliversidge:
Thanks guys!

cpx86:
Nice! Looks a lot nicer than before :)

PR #206, opened 2017-04-29
Snapshot strategies

tomliversidge:
Adds functionality for automatically saving snapshots according to a strategy. This takes the burden away from the actor for having to save snapshots. Supported strategies are

- Interval strategy: save snapshot every N events, i.e. every 100 events
- Time strategy: save snapshot based upon a timeframe, i.e. every day
- EventType strategy: save snapshot when a certain event type is seen i.e. ReallyImportantEvent

Snapshot strategies are an optional feature available when using Event sourcing and Snapshotting together only (it makes no sense to use them in the other supported modes)


rogeralsing:
LGTM 👍

PR #208, opened 2017-05-01
add AllForOneStrategy for supervision

tomliversidge:
AllForOneStrategy ported as requested in https://github.com/AsynkronIT/protoactor-dotnet/issues/197

Issue #197, opened 2017-04-23
Port all supervisor strategies from Go to C#


rogeralsing:
AllForOne and ExponentialBackoff supervisor strategies needs to be ported over.

edblackburn:
Do you think you'll take a dependency on Polly? Or will you write your own? If the later is there scope for injecting the policy as a delegate?

rogeralsing:
We can probably have some contrib/extension package for Polly, but I would like to avoid it out of the box. 
The default supervision strategies are simple enough to just implement directly. 10-20 lines each.

tomliversidge:
Does it matter if we just use milliseconds for the c# version rather than nanoseconds in the ExponentialBackoff strategy?

```
public async void HandleFailure(ISupervisor supervisor, PID child, RestartStatistics rs, Exception reason)
        {
            SetFailureCount(rs);
            var backoff = rs.FailureCount * _initialBackoff.TotalMilliseconds;
            var noise = _random.Next(500);
            var duration = TimeSpan.FromMilliseconds(backoff + noise);
            await Task.Delay(duration);
            supervisor.RestartChildren(child);
        }
```

or would you prefer to convert to Nanoseconds?

tomliversidge:
Also, if I use Task.Delay(duration).Wait() it is blocking, do we want HandleFailure to be async?

adamhathcock:
Yes.  Thanks a big sin to commit right there.

tomliversidge:
Yeah that's what I though, but it's going to have a ripple effect as HandleFailure isn't currently awaited

adamhathcock:
Async all the things!

adamhathcock:
I can give it a go to async it.  Will have to be tomorrow as it's a U.K. Holiday today

adamhathcock:
I gave it a go to make that code path be async.

rogeralsing:
In Go ExponentialBackoff:

```go
func (strategy *exponentialBackoffStrategy) HandleFailure(supervisor Supervisor, child *PID, rs *RestartStatistics, reason interface{}, message interface{}) {
	strategy.setFailureCount(rs)

	backoff := rs.FailureCount * int(strategy.initialBackoff.Nanoseconds())
	noise := rand.Intn(500)
	dur := time.Duration(backoff + noise)
	time.AfterFunc(dur, func() {
		supervisor.RestartChildren(child)
	})
}
```

[...]

PR #209, opened 2017-05-01
Made Async: system message processing in LocalContext and HandleFailu…

adamhathcock:
…re in Supervision.

Related to https://github.com/AsynkronIT/protoactor-dotnet/pull/208 and https://github.com/AsynkronIT/protoactor-dotnet/issues/197

### DIFF
--- a/src/Proto.Actor/LocalContext.cs
+++ b/src/Proto.Actor/LocalContext.cs
@@ -175,45 +175,49 @@ namespace Proto
             return ProcessMessageAsync(message);
         }
 
-        public Task InvokeSystemMessageAsync(object msg)
+        public async Task InvokeSystemMessageAsync(object msg)
         {
             try
             {
                 switch (msg)
                 {
                     case Started s:
-                        return InvokeUserMessageAsync(s);
+                        await InvokeUserMessageAsync(s);
+                        break;
                     case Stop _:
-                        return HandleStopAsync();
+                        await HandleStopAsync();
+                        return;
                     case Terminated t:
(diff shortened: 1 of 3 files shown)


tomliversidge:
You can squash that "oops" commit by doing (on the branch) 
`git rebase -i HEAD~2`

then selecting `f` for fixup on the oops commit.

then do 

`git push --force`

adamhathcock:
Done.  I usually hate rewriting history.

tomliversidge:
I usually do it when on my own branch and no-one else is working on it, as I prefer a single commit to implement something - it makes it easier to read the history and cherry-pick / revert if ever needed (which admittedly is rare, but when you need to it can be awful to go through a bad history...)

One other thing I noticed is there is 

```
private void HandleRootFailure(Failure failure)
        {
            Supervision.DefaultStrategy.HandleFailure(this, failure.Who, failure.RestartStatistics, failure.Reason);
        }
```

which also presumably needs async-ifying?

rogeralsing:
Under what conditions would we need to await the supervisor?
When a child actor fails, it will be suspended, the failure is then escalated to the parent.
The parent handles the failure using its supervisor strategy.
This don't really need to be awaited by the parent actor, it could simply branch off and do it's own thing.

e.g. in the ExponentialBackoffStrategy, the supervisor could await in it's own task, separate from the actor.

if the actor awaits, this means that the parent actor cannot do anything while awaiting the supervisor.

adamhathcock:
Don't need this as any timer that delays actions will NOT be awaited

PR #210, opened 2017-05-01
Support multiple address spaces in the same AppDomain

bearbytes:
It can be useful to seperate actors into different address spaces. This is always done when the actor runs on a different machine, or maybe a different AppDomain on the same machine, but it is not possible to maintain multiple addresses within the same AppDomain, since `ActorRegistry` is a singleton with a fixed address.

This PR allows to define multiple `ProcessHost`, which are containers that hold the references to Actor Processes from a single address. There is always one default ProcessHost created by the `ProcessRegistry`. When spawning an actor, one can specify the ProcessHost to be used in the `Props`, and the PID address will be set accordingly.

This solves two problems:
- It allows us to create private addresses, which can only be reached from within the AppDomain, not from Remotes.
- It allows us to use the same Id for multiple actors, when they have different addresses. This is useful for example to simulate copies of the same client within the AppDomain, when the client uses named actors.

Example:
```
var client1Host = new ProcessHost("client1");
var client2Host = new ProcessHost("client2");
ProcessRegistry.Instance.RegisterProcessHost(client1Host);
ProcessRegistry.Instance.RegisterProcessHost(client2Host);

var props = Actor.FromFunc(ctx => {});

// = PID("client1", "AccountManager")
var pid1 = Actor.Spawn(props.WithProcessHost(client1Host), "AccountManager");

// = PID("client2", "AccountManager") 
var pid2 = Actor.Spawn(props.WithProcessHost(client2Host), "AccountManager");

pid1.Tell(""); // Look ma, no remote!
```

Remarks:
`RegisterProcessHost` is not threadsafe, so should be done once during startup.
Props use `ProcessRegistry.Instance.DefaultProcessHost` if not specified otherwise.
The additional addresses are only visible within the program, and cannot be reached from the outside.

There might be an argument to allow these additional ProcessHosts to also have additional RemotingSystems. I have not yet enough experience with the RemotingSystem to know if this is feasible. There might also be overlapping concerns with the Cluster system.


rogeralsing:
Interesting!
So TLDR;
There are no changes to PID's, right?
The process host is resolved by the address of the PID?

Meaning that if we receive a PID via a message from a remote machine, we can resolve that and know where to connect.
This needs to be carefully tested before merging as it will most likely have some impact on Remote and Cluster packages.

pancomputationalist:
No changes to PID.
ActorRegistry resolves PID by first resolving ProcessHost from PID.Address (falling back on Remote otherwise), then the ProcessHost resolves the Process from PID.Id.
When we receive messages from a Remote host, currently always the DefaultProcessHost is used. IIRC, the Address-Part is stripped from Remote Messages. So we'd have to have multiple Endpointreader if we'd wanted to have the other Addresses remote-available.

cpx86:
It's an interesting change and intuitively I like it, but I'd like to first discuss what issues it might solve. Please fill in :)

> It allows us to create private addresses, which can only be reached from within the AppDomain, not from Remotes.

This highlights an important point that we haven't really addressed yet, namely, how to secure actors? One way that you could do this currently, I think, is to separate sensitive actors to a separate process, using a gRPC channel that requires certain credentials. This would of course introduce an extra network hop for some situations. OTOH, if remote performance is good enough, an extra network hop should be acceptable. A limitation of going remote though is that it's more expensive if you need to share state between actors that live in different processes.

> It allows us to use the same Id for multiple actors, when they have different addresses. This is useful for example to simulate copies of the same client within the AppDomain, when the client uses named actors.

Is there a production use-case for this, or is it mainly a testability problem?

pancomputationalist:
I will explain my setup and thoughts on security.

My clients are running Unity3D, which does only support .NET 3.5. I tried backporting Proto.Actor, but couldn't find a suitable gRPC Implementation, so instead I use a websocket connection to transfer messages from my clients to the server. Notably, I'm using the server as a proxy for client-to-client messages, because mobile p2p doesn't always work, and it does not expose the IPs of other users.

So I'm not working with the Remote Package and can use whatever address I want for my PIDs, which is `"user:xyz"` and `"server"`. When my server receives a message, it either forwards them to another client, or wraps them into a `UserMessage`, which also contains Session/Context info about the user connection. So I can have client-facing Actors, which unwrap this UserMessage and can query the context if the message is valid/allowed, while all other actors in the process just ignore these messages.

I don't know if something like that is possible to do with the default Remote package (as I have not used it), but it should be easily with some kind of incoming middleware, which can rewrite the receiver, message or sender of incoming messages.

Now, I have some "fake" User-Accounts, which are actually bots, but also have a `"user:xxx"` address, but live in the server process. What I have to do now is rewrite the PIDs to some other naming scheme. Since I do NOT wrap these messages into a `UserMessage`, I don't want them to accidentally land in a "private" actor on my server, so idealy, I could use something like a `ProcessHost` to ensure that they are in a completely different namespace, which generally makes them easier to spawn and secure, and avoids naming conflicts.

As I understand it now, the Remote package just "blindly" accepts all incoming messages and forwards them directly to the actors, which doesn't allow for much security, unless its handled in the actor itself (again, I may have overlooked something here).

For my use case, the following combination of features would work best:
- (Have a .NET 3.5 Client, but I wouldn't expect that anytime soon, since Unity is going to .NET 4.6)
- Allow non-ip addresses with dynamic routing (maybe already possible?)
- Allow multiple address spaces within the same process (like processes in an os), so they can talk to each other, but are protected against the internet by a security layer/remote middleware (like a firewall)

rogeralsing:
Is there a way to make this more of an opt-in?
The reason why I'm on the fence about merging this is that it could have a huge impact on cluster mechanics.

I do like the gist of it though but this is a mouthful when it membership and remote termination etc.

rogeralsing:
Closing as this is out of scope for now.
I am personally playing around with other transports so I might ressurect this topic in the near future

PR #211, opened 2017-05-01
add ExponentialBackoff supervisor strategy

tomliversidge:
closes https://github.com/AsynkronIT/protoactor-dotnet/issues/197

### DIFF
--- a/src/Proto.Actor/Supervision.cs
+++ b/src/Proto.Actor/Supervision.cs
@@ -7,6 +7,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 
 namespace Proto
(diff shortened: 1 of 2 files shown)


src/Proto.Actor/Supervision.cs
```diff
@@ -171,4 +172,51 @@ private bool RequestRestartPermission(RestartStatistics rs)
...
+            return nanoseconds / 1000000;
+        }
+
+        public void HandleFailure(ISupervisor supervisor, PID child, RestartStatistics rs, Exception reason)
+        {
+            SetFailureCount(rs);
+            var backoff = rs.FailureCount * ToNanoseconds(_initialBackoff);
+            var noise = _random.Next(500);
+            var duration = TimeSpan.FromMilliseconds(ToMilliseconds(backoff + noise));
+            Task.Delay(duration).ContinueWith(t =>
```
adamhathcock: Probably should only continue on success with this.  Maybe continue with a log when an error occurs

adamhathcock: Though might be overkill with just a Delay :/

rogeralsing:
LGTM 👍

Issue #196, opened 2017-04-23
Supervision does not get triggered if child actor throws in Started message


rogeralsing:
If a child actor throws when receiving the `Started` message, the parent will not get notified.
I assume that there is something missing in that sequence, e.g. that the parent is not yet set or something.

tomliversidge:
Is it because the exception is swallowed? https://github.com/AsynkronIT/protoactor-dotnet/blob/dev/src/Proto.Actor/LocalContext.cs#L236

cpx86:
@rogeralsing Is there any reason why we have that catch? Wouldn't the parent always want to get notified on failed system messages, so they can be retried?

rogeralsing:
Well spotted, agree we can get rid of the try catch there, ofc failing system messages should also be escalated to parent.

tomliversidge:
I modified the existing supervisor tests to throw an exception when the started message is received (after cahnging the code above to rethrow the exception) and 2/4 of them failed... i think it was restart and possibly escalate strategies. It was pretty difficult to debug but i think the flow was correctly sending a restart message, but then inside the processing of that it was throwing another exception, as another started message must have been processed and this then resulted in a failed task for processing the restart message... but then my kids, partner and mother came in and that was the end of that!

PR #212, opened 2017-05-01
do not swallow exception when processing system messages

tomliversidge:
re-throw exception when processing system messages rather than swallowing
test to prove supervision works when failure/exception processing Started message

Fixes https://github.com/AsynkronIT/protoactor-dotnet/issues/196


rogeralsing:
👍

PR #216, opened 2017-05-04
NonBlockingBoundedMailbox

rogeralsing:
This is a first draft of a nonblocking bounded mailbox.
This is very useful for load shedding in case your system gets too much load.
Better to drop load than to fall over completely.

The mailbox has three arguments, the max size, the action to perform if the message was not enqueued and the timespan to wait.

### DIFF
--- a/src/Proto.Mailbox/NonBlockingBoundedMailbox.cs
+++ b/src/Proto.Mailbox/NonBlockingBoundedMailbox.cs
@@ -1,0 +1,48 @@
+﻿// -----------------------------------------------------------------------
+//   <copyright file="NonBlockingBoundedMailbox.cs" company="Asynkron HB">
+//       Copyright (C) 2015-2017 Asynkron HB All rights reserved
+//   </copyright>
+// -----------------------------------------------------------------------
+
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+
+namespace Proto.Mailbox
+{
+    public class NonBlockingBoundedMailbox : IMailboxQueue
+    {
+        public NonBlockingBoundedMailbox(int maxSize, Action<object> overflowAction, TimeSpan timeout)
+        {
+            _maxSize = maxSize;
+            _overflowAction = overflowAction;
+            _timeout = timeout;
+        }
+
+        private readonly ConcurrentQueue<object> _messages = new ConcurrentQueue<object>();
+        private readonly int _maxSize;
+        private readonly Action<object> _overflowAction;
+        private readonly TimeSpan _timeout;
+
+        public void Push(object message)
+        {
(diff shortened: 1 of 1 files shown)


src/Proto.Mailbox/NonBlockingBoundedMailbox.cs
```diff
@@ -0,0 +1,48 @@
...
+namespace Proto.Mailbox
+{
+    public class NonBlockingBoundedMailbox : IMailboxQueue
+    {
+        public NonBlockingBoundedMailbox(int maxSize, Action<object> overflowAction, TimeSpan timeout)
+        {
+            _maxSize = maxSize;
+            _overflowAction = overflowAction;
+            _timeout = timeout;
+        }
```
rogeralsing: the overflowAction can be used to allow dropped messages to go the dead letters.

src/Proto.Mailbox/NonBlockingBoundedMailbox.cs
```diff
@@ -0,0 +1,48 @@
...
+        private readonly Action<object> _overflowAction;
+        private readonly TimeSpan _timeout;
+
+        public void Push(object message)
+        {
+            if (SpinWait.SpinUntil(() => _messages.Count < _maxSize, _timeout))
+            {
+                //this will be racy, but best effort is good enough..
+                _messages.Enqueue(message);
+            }
```
rogeralsing: Once the condition is met, another thread could have enqueued a message on the queue and the condition is now false again. but best effort is good enough here. 
the mail purpose is to prevent the system to tip over due to over load. 1-2 messages too many on the queue will not cause that.

tomliversidge: I've been wondering if we need the SpinWait around this code - I thought a ConcurrentQueue allows multiple threads to Count and Enqueue? I'm not an expert in multithreading by any means but when I remove this SpinWait it seems to work pretty much the same as with it, only much faster (just messing around with a test that creates lots of Tasks that post to the mailbox)

rogeralsing: W/O the spinwait, it will not wait for the timeout to occur, it will simply check there and then if the condition is met.
With the spinwait it will wait for e.g 0.5 sec before giving up and moving the message to the overflow action

tomliversidge: Ok I think I get it now - the idea is you give the actor a short window of opportunity to process messages (the timeout value) and if there are still too many, give up and call the overflow... And this is because producers can be too fast and overwhelm the mailbox.

PR #221, opened 2017-05-13
fix issue with persistence when recovering state

tomliversidge:
After loading a snapshot, the recover state method would then attempt to replay events from the index that the snapshot was taken at (i.e. 10). However, it would include the last event the snapshot was taken at (i.e. 10!) which isn't correct - it should replay events from the next index (i.e. 11)

example
what happened: 1,2,3,4,Snapshot,5,6
recover state: snapshot,4,5,6 < THE 4 HERE IS WRONG

should be:
what happened: 1,2,3,4,Snapshot,5,6
recover state: snapshot,5,6

### DIFF
--- a/src/Proto.Persistence/Persistence.cs
+++ b/src/Proto.Persistence/Persistence.cs
@@ -76,7 +76,7 @@ namespace Proto.Persistence
 
             if (UsingEventSourcing)
             {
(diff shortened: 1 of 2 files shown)


tomliversidge:
@AsynkronIT/developers am I ok to merge this in?

raskolnikoov:
Approved!

PR #222, opened 2017-05-13
Persistence API - Allow replaying of events within a range

### LINKED ISSUE #188: Persistence API - Support replaying up to a certain event index and time
We currently only support rebuilding the state of an actor to the latest version / index, but there are use cases (mainly debugging / testing) where it would be useful to be able to replay up to a certain point in time or a certain index. For example, if we want to replay until just before something happened (i.e. unexpected behavior of the system, bug, crash, "the system went down at 12:00" etc..) then apply some messages and observe what happens.

tomliversidge:
There are use cases (mainly debugging / testing) where it would be useful to be able to replay up to a certain index. For example, if we want to replay until just before something happened (i.e. unexpected behavior of the system, bug, crash, "the system went down at 12:00" etc..) then apply some messages and observe what happens.

closes #188

### DIFF
--- a/src/PersistenceProviders/Proto.Persistence.Couchbase/CouchbaseProvider.cs
+++ b/src/PersistenceProviders/Proto.Persistence.Couchbase/CouchbaseProvider.cs
@@ -20,28 +20,20 @@ namespace Proto.Persistence.Couchbase
         {
             _bucket = bucket;
         }
-
-        public Task GetEventsAsync(string actorName, long indexStart, Action<object> callback)
-        {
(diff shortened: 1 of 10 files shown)


tomliversidge:
I added a second commit that reduced the IProvider interface to have a single GetEventsAsync method that accepts a toIndex - when not being used (i.e. in RecoverStateAsync), long.MaxValue is used by default to read all events. This makes it a bit easier to write providers. Could squash this second commit if preferred.

rogeralsing:
This one needs a rebase

tomliversidge:
Ok, done

Issue #202, opened 2017-04-26
Allow actors to be IDisposable


rogeralsing:
Sometimes, you might want to use disposable objects from within your actors.
IMO, we should check if the actor instance also implements IDisposable and call dispose upon stopping/restarting the actor.

adamhathcock:
I don't mind adding syntactic sugar like this but it did seem some other sugar was being removed or moved to middleware.  Just trying to get a sense of what's Core and what isn't.

rogeralsing:
But would you consider this sugar?
If an actor has IDisposable dependencies, when and where should we dispose them?
intercept Stopped message and do it there?

adamhathcock:
`intercept Stopped message and do it there?`

That's what I would have said is the "right" thing to do.

rogeralsing:
BTW. to clarify what I want to move out as opt-in, anything that comes with weight, either in CPU usage and/or memory.

e.g. stashing requires the context to keep a stack of messages, meaning it has weight.

Anything that just makes life easier w/o affecting perf, then that should be considered.
but I agree, it might be better to just be consistent when it comes to how we recommend developers do things.

cpx86:
I think this could make sense to implement, since the context owns the `IActor` instance. Generally, whoever owns an `IDisposable` object should also dispose it.

One potential caveat though - how might this interact with a container if one uses one? If a container created the actual actor instance, shouldn't it be the one to dispose it and it's dependency graph?

rogeralsing:
>One potential caveat though - how might this interact with a container if one uses one? If a container created the actual actor instance, shouldn't it be the one to dispose it and it's dependency graph?

The eternal issue with IoC containers and actors.

The thing is, as an actor can restart w/o actually dying, the resources the old incarnation held on to must be disposed.

The same goes when the actor is eventually terminated for good. 
but there is a mismatch between IoC container lifecycle and the actor lifecycle as the actor can be restarted many times.

adamhathcock:
That's a good point.  If an IoC is involved, there should be a implicit tie to the container.  So if the container is disposed then it signals the actors it owns to terminate.  An actor restarting or dying doesn't affect IoC, just usage of the actor if it's dead when someone tries to use it.

cpx86:
Thinking a bit more about it though, it doesn't matter if the context disposes of the actor and then a container does so as well, since any `IDisposable` implementation should keep track of whether it's already been disposed.

rogeralsing:
[...]

PR #225, opened 2017-05-16
Call Dispose on IDisposable actors when Stopped or Restarted

tomliversidge:
Attempt at implementing https://github.com/AsynkronIT/protoactor-dotnet/issues/202. When Restarting or Stopping, the context looks to see if the actor implements IDisposable and if so called Dispose on it.

### DIFF
--- a/src/Proto.Actor/LocalContext.cs
+++ b/src/Proto.Actor/LocalContext.cs
@@ -522,14 +522,7 @@ namespace Proto
         {
             if (Actor is IDisposable disposableActor)
             {
-                try
-                {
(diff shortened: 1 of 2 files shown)


cpx86:
Looks good! Just one thing, should we really swallow exceptions thrown from Dispose? If an actor fails to dispose of unmanaged resources, that could cause problems. Might be better to leave that decision to the actor instead

tomliversidge:
Yeah makes sense :) - I didn't have the try-catch in there originally.

cpx86:
LGTM!

Issue #228, opened 2017-05-30
Include exception in lifecycle messages


tomliversidge:
When raising a Restart / Stop / Resume message, we do not currently pass along the reason why this message was generated, even though the supervisor has access to it through the 
 HandleFailure method. 

We should add the reason to the message to enable actors to act on it when they receive the lifecycle messages

tomliversidge:
i guess in the Stop scenario, we might be stopping because N number of exceptions have been received within a time period, so doesn't make sense to pass down a single exception in this case

For Resume we're essentially saying "something went wrong, but just carry on", so it might be useful to know what the thing was that went wrong

PR #229, opened 2017-05-30
pass exception to restart message

tomliversidge:
Added reason why a restart was deemed necessary

https://github.com/AsynkronIT/protoactor-dotnet/issues/228

### DIFF
--- a/src/Proto.Actor/Supervision.cs
+++ b/src/Proto.Actor/Supervision.cs
@@ -23,8 +23,8 @@ namespace Proto
     public interface ISupervisor
     {
         IReadOnlyCollection<PID> Children { get; }
-        void EscalateFailure(PID who, Exception reason);
-        void RestartChildren(params PID[] pids);
+        void EscalateFailure(Exception reason, PID who);
+        void RestartChildren(Exception reason, params PID[] pids);
         void StopChildren(params PID[] pids);
         void ResumeChildren(params PID[] pids);
     }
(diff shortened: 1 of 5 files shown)


adamhathcock:
Are Reasons always going to be exceptions?

tomliversidge:
I _think_ so. As far as I understand, actors only get restarted when their supervisors tell them to, and this supervision process only kicks in when an exception is thrown.

tomliversidge:
```
var t = _invoker.InvokeUserMessageAsync(msg);
                        if (t.IsFaulted)
                        {
                            _invoker.EscalateFailure(t.Exception, msg);
                            continue;
                        }
```

is the code in the mailbox, so it's only on a faulted task that failure is escalted

adamhathcock:
Works for me.  I was only thinking if someone wanted to have an explicit reason but that's probably an edge case that isn't worth accounting for.

src/Proto.Actor/Supervision.cs
```diff
@@ -23,8 +23,8 @@ public enum SupervisorDirective
     public interface ISupervisor
     {
         IReadOnlyCollection<PID> Children { get; }
-        void EscalateFailure(PID who, Exception reason);
-        void RestartChildren(params PID[] pids);
+        void EscalateFailure(Exception reason, PID who);
```
tomliversidge: I switched this around to keep it consistent with RestartChildren, as that method uses params, which need to come last

PR #233, opened 2017-06-02
Reduce API surface

cpx86:
Relates to #176 

Basically I've just made a bunch of classes internal, that aren't useful to expose to users :)


cpx86:
Hmm, making ActorClient non-public broke the Middleware example. Should it be public or should the Middleware example be changed to not reference that class?

cpx86:
Reverted the ActorClient change for now since it was breaking.

PR #235, opened 2017-06-03
Add KafkaProducerProcess

cpx86:
As discussed here and there before, `Process` can be used to abstract things like message queues behind a `PID`.

I've added an initial stab at a process that wraps a Kafka producer. It doesn't currently have any way to customize the serializer, so ideas ot fix this would be welcome :)


tomliversidge:
I missed discussions around this, so could you explain what's the difference between implementing this at the process level rather than as an actor? I.e. Do we lose supervision capabilities this way?

src/Proto.Actor/ProcessRegistry.cs
```diff
@@ -47,12 +47,24 @@ public Process Get(PID pid)
...
+            };
+            var ok = _localActorRefs.TryAdd(pid.Id, aref);
+            return (pid, ok);
+        }
+
         public (PID pid, bool ok) TryAdd(string id, Process aref)
         {
             var pid = new PID
             {
                 Id = id,
```
adamhathcock: Rather than have the above additional method can you just do: `Id = id ?? NextId();` ?

adamhathcock: I guess you like having the overload that takes no Id though.  Just trying to collapse duplicated code.

raskolnikoov:
I've built many systems using kafka and actors together and it works great. So this is basically a plugin implementation with the confluent api? Looks good but the producer needs to have more config options. And maybe we should have a ConsumerProcess also that reports back all the messages it consumes from kafka to its parent actor? Thats how i currently do it in my daily work applications.

rogeralsing:
Me and @cpx86 have been discussing this for quite a while, that Actors are just one form of message passing construct. there can be others, Kafka, Rabbit, SignalR etc.
All of those concepts could be given a PID so that you can send messages to that PID.
(Just like Routers do, as they have a special process which ignores the actor concurrency constraint)

This can be implemented as a different kind of process as done here, or it could be implemented using actors that then delegate the messages forward to the underlying construct.

Just brainstorm away on what would be the proper way forward.
Custom Process = We can ignore actor concurrency and mailbox.
Wrapper Actor = We can leverage supervision and lifecycles.

Go :-)

src/Proto.Actor/ProcessRegistry.cs
cpx86: Ah maybe the first method should call the second method?

cpx86:
One downside I noticed with the current PID/Process signatures is that they're not async. Would it make sense to have them return a Task?

rogeralsing:
@cpx86 agree, Tell and the downstream methods including mailbox methods really should have been async.

cpx86:
@rogeralsing Should we change it? Shouldn't break too much.

cpx86:
Added another example that I think highlights the usefulness of this a bit better: adding an actor and a KafkaProducerProcess to a broadcast router, so all messages are sent both to an in-memory actor and persisted onto a Kafka topic.

cpx86:
@raskolnikoov yeah I was thinking of adding a consumer process as well, but I'm not sure how it would look. Any suggestions? Maybe something like the EventStream?

cpx86:
On second thought, let's not go to Camelot. 'Tis a silly place.

PR #238, opened 2017-06-08
Use FirstOrDefaultAsync instead FirstAsync to avoid System.AggregateException(Sequence contains no element)

dominikus1993:


rogeralsing:
Thanks!

PR #239, opened 2017-06-08
Behavior tests

tomliversidge:
Demonstrates use of different states and globally handling some message types


tomliversidge:
Not sure where all those extra commits came from - can someone pick "Squash and merge" if PR accepted

PR #240, opened 2017-06-10
Rename props method WithSupervisor to WithChildSupervisorStrategy

tomliversidge:
This is to make it clearer that the supervision strategy you are creating is for the actor you are creating's children... i.e. Is-A supervisor rather than Has-A supervisor

Also has the advantage of appearing in intellisense still when you type With :)

### DIFF
--- a/examples/Supervision/Program.cs
+++ b/examples/Supervision/Program.cs
@@ -18,7 +18,7 @@ class Program
         Proto.Log.SetLoggerFactory(new LoggerFactory()
             .AddConsole(minLevel: LogLevel.Debug));
 
-        var props = Actor.FromProducer(() => new ParentActor()).WithSupervisor(new OneForOneStrategy(Decider.Decide, 1, null));
+        var props = Actor.FromProducer(() => new ParentActor()).WithChildSupervisorStrategy(new OneForOneStrategy(Decider.Decide, 1, null));
 
         var actor = Actor.Spawn(props);
         actor.Tell(new Hello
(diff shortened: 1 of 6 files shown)


tomliversidge:
Other considerations were "SupervisesUsing(strategy)" or "WithSupervisionStrategy(strategy)" - if anyone has a strong preference for something else I'll update this PR - i changed my mind a few times already :)

adamhathcock:
Not sure I like this as there seems to be a pattern with `With*` and renaming doesn't provide any immediate benefit?

rogeralsing:
WithChildSupervisorStrategy ?

adamhathcock:
Oh yeah...already more clear as to the intent as I forgot the intent of renaming this.

Other ideas: `SuperviseChildrenWith`  though I like @rogeralsing 's suggestion best I think.

tomliversidge:
Yeah I was conscious that this was breaking with the current With* naming convention - let's go for WithChildSupervisorStrategy as that makes it pretty clear what it is

tomliversidge:
I've been thinking a bit more about this and I actually prefer the "SuperviseChildrenWith" suggestion from @adamhathcock as I think it makes it clearer that it is a process rather than a thing - now it sounds like we have a child supervisor or a child has a supervision strategy :) (also keep intellisense when typing .With!)

Or use the word "supervision" rather than "supervisor" in this context as again it implies a process. "WithChildSupervisionStrategy" but again, I'd argue not as clear as "SuperviseChildrenWith"

:)

cpx86:
IMHO if we are going to break away from the With* convention then we should do it across the board. I don't like inconsistency :P

`WithChildSupervisionStrategy` makes sense to me - since you're not actually creating a "supervisor", and also it applies to the children, not the actor itself.

adamhathcock:
I agree about the inconsistency.  It makes my eyes twitch if things are inconsistent.  While I think `SuperviseChildrenWith` makes more sense "english" wise, I'd rather stick with the `With*` convention.

PR #241, opened 2017-06-11
Merge router actor and router process

rogeralsing:
This PR merges the router actor and the router process.
Previously, we had one router process then pointing to a separate router actor, with its own local process.

TLDR; the RouterProcess was a proxy before directing messages either to the router actor, or to the routees.
Now the router process and the router actor are the same thing


cpx86:
:+1:

PR #243, opened 2017-06-13
Cleanup Consul provider from Couchbase client

raskolnikoov:
The Consul provider has a reference to CouchbaseNetClient that is not needed.


cpx86:
Nice catch! 👍

Issue #236, opened 2017-06-06
Make Tell async?


rogeralsing:
Tell is currently synchronous and blocking, this is not an issue for now as we always just pass messages onto a mailbox.

But as we have been discussing integrating other technologies, not only actors, it might make sense to actually be able to await the Tell.
e.g. if we send messages to Kafka or SignalR, then you probably want to be able to wait until the message have actually been placed on the underlying infrastructure.

This would not have any major effect on performance of the default path as that would simply return a completed Task.
There will ofc be the overhead of returning that pointer instead of void, but that would be it.

cc @AsynkronIT/developers

rogeralsing:
Related:

Also, should we rename Tell to Send?

This is a much more common convention. Dataflow uses it, Elixir uses it, most MQ SDK's uses it.
Tell is very Akka specific.

tomliversidge:
I guess there are arguments for and against - sticking with it makes it more familiar to Akka users, but less familiar to non-Akka users.

I'd vote for Send as like you say it is more common overall and given that we have Receive methods would fit better with those too

adamhathcock:
+1 for naming it Send and making it async even though it doesn't need to be.  I like the consistency.

raskolnikoov:
I vote for change to Send. Is it a bad idea to have a Send and SendAsync method for sync/async operations?

adamhathcock:
I personally don't like *Async methods unless the API already had sync versions.  Just have one Send and make it async

cpx86:
+1 for Send. As for the Async suffix, I'm not sure. Suffixing seems to be pretty standard so it could make sense to stick to it to not surprise the user

adamhathcock:
if the entire API is async, then you would put suffix it all?  I don't think async suffix is standard.  Especially for new async-only APIs.

cpx86:
SO has a good answer here from Stephen Toub who worked on TPL :) https://stackoverflow.com/questions/15951774/does-the-use-of-the-async-suffix-in-a-method-name-depend-on-whether-the-async

adamhathcock:
¯\_(ツ)_/¯

I guess it should be used to make it easy to distinguish I guess but I still hate it :p

tomliversidge:
Yeah probably go with Async suffix as does make it super obvious.

jeffdoolittle:
[...]

PR #244, opened 2017-06-13
Process send messaging is now async (Preview)

adamhathcock:
Solves https://github.com/AsynkronIT/protoactor-dotnet/issues/236

This has a big knock-on affect.  Need to fix some remote tests.  This is a preview

Notable:
- Found lots of unit tests that should be `async Task`
- Exponential Backoff supervision should have done a null check on the supervisor.  Threw exceptions otherwise.
- I wish there was async main for the examples


src/Proto.Mailbox/Mailbox.cs
```diff
@@ -115,10 +115,9 @@ private Task RunAsync()
                     _stats[i].MailboxEmpty();
                 }
             }
-            return Task.FromResult(0);
         }
 
-        private bool ProcessMessages()
+        private async Task<bool> ProcessMessages()
```
rogeralsing: cc @cpx86 We should take a closer look at this as this part was highly optimized to avoid the async statemachine.
This will come at a pretty big perf penalty

rogeralsing:
Looks good. some concerns:

* async keyword introduced in the ProcessMessage pipeline, this will affect perf, we should measure and see how much it affects

* We should probably clean up the examples and move code from Main to an async func to get rid of the .Wait() calls.

* I need to digest the changes a bit as it is a big step in a new direction :-)

src/Proto.Mailbox/Mailbox.cs
adamhathcock: Could do Wait() on the failure messages as it's just on the error cases.

Is the perf benchmarks you usually do checked in?

adamhathcock:
I guess I cleaned up line endings too.  View without whitespace changes: https://github.com/AsynkronIT/protoactor-dotnet/pull/244/files?w=1

src/Proto.Mailbox/Mailbox.cs
cpx86: I think we could handle the Task returned by EscalateFailure the same way as the one returned when invoking the message. Continue if complete, otherwise bail out and reschedule

adamhathcock: So this?

```csharp
_invoker.EscalateFailureAsync(t.Exception, msg).ContinueWith(RescheduleOnTaskComplete, msg);
```

cpx86: Should work :) Tiny improvement:
```
var t = _invoker.EscalateFailureAsync(t.Exception, msg);
if (!t.IsCompleted)
{
    t.ContinueWith(RescheduleOnTaskComplete, msg)
    return false;
}
```

We should probably also check if the task is faulted - but I'm not sure what we can do if that happens. Log a warning?

adamhathcock: That's what happens in the `RescheduleOnTaskComplete` check for faulted that is.  Then tries to send to `EscalateFailureAsync` again with the fault from the task.

Can add some logging.

cpx86: It needs to be checked twice. Once for the case when the task runs synchronously. If it ran asynchronously then it needs to be checked in the continuation instead.

src/Proto.Mailbox/Mailbox.cs
```diff
@@ -5,11 +5,17 @@
 // -----------------------------------------------------------------------
 
 using System;
+using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 
 namespace Proto.Mailbox
 {
+    public static class Tasks
```
cpx86: Make this internal?

src/Proto.Mailbox/Mailbox.cs
```diff
@@ -187,10 +192,20 @@ public void Start()
             }
             catch (Exception e)
             {
-                await _invoker.EscalateFailureAsync(e, msg);
+                HandleProcessMessageException(e, msg);
             }
             return true;
         }
+        
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
cpx86: Can we be sure that the method will actually be inlined?

src/Proto.Mailbox/Mailbox.cs
adamhathcock: I was trying to share the Task object like `Actor.Done` but I guess having it internal is fine to not expose more code.

src/Proto.Mailbox/Mailbox.cs
adamhathcock: About as sure as I can be about anything.  Worth testing?

cpx86: I tried throwing an exception from `EscalateFailureAsync` and got the following stack trace, so it seems that it didn't inline:
```
System.Exception: Exception of type 'System.Exception' was thrown.
   at Proto.LocalContext.EscalateFailureAsync(Exception reason, Object message) in C:\code\github\protoactor-dotnet\src\Proto.Actor\LocalContext.cs:line 248
   at Proto.Mailbox.DefaultMailbox.HandleProcessMessageException(Exception exception, Object message) in C:\code\github\protoactor-dotnet\src\Proto.Mailbox\Mailbox.cs:line 205
   at Proto.Mailbox.DefaultMailbox.ProcessMessages() in C:\code\github\protoactor-dotnet\src\Proto.Mailbox\Mailbox.cs:line 150
```

adamhathcock: Does it require the JIT to do it as a hotspot?  might not happen for various reasons.

Would you rather the logic just be copy pasted into the method then?

cpx86: Honestly not sure what will happen, the JIT docs are very vague. `AggressiveInling` will apparently inline the code regardless of the method size, but it doesn't specify if that means immediately or only when it identifies it as a hot path. So I'm not sure if the JIT would eventually inline it, or if it decided that the method shouldn't/couldn't be inlined.

In any case though, actors shouldn't fail that often anyway, so I'm fine with leaving it as is. I doubt it will hurt performance in any meaningful way. I was mostly curious about whether the inlining will actually happen or not :)

adamhathcock:
from @rogeralsing on another issue :)
>Me and @cpx86 have been discussing this here and opt for putting this on hold for now.
>While having the same abstraction (PID) for any message based tech, it might very well be a bad abstraction, as in this case where we are forced to do major changes to the API just to add a bit of more guarantees.
>
>if anyone wants to pass a Kafka topic name around, that could just as well be sent as a string or custom message and we could use a custom abstraction for Kafka.
>
>This does not mean we shouldnt go this way, but lets wait for a better usecase or more usecases and evaluate if it makes sense or if one should just use whatever abstraction those cases need there and then

Your reasoning makes sense but I do think the API should be async everywhere so that things like Kafka can be plugged in.  I do think some rejiggering of the abstractions will make things make more sense.

Close this if needed!

cpx86:
I'm kind of having second thoughts about the initial idea of plugging in Kafka through a custom process. Does it really add any value to the user? In order for it to be useful then you need to be able to fully customize the Kafka producer, and in that case, you would probably also want to have full control over the lifecycle of it. It's starting to feel like we're maybe just adding it because it seems cool, not because it's actually solving an important problem. If a user does want to use a Kafka producer in a Proto.Actor solution, then writing the code yourself is pretty trivial and will probably be way easier to reason about than if we try to abstract it behind a process. Just my 2c :)

cpx86:
Another factor is that one of the sort of "guiding principles" behind the library is that it should be quite minimalist and only contain the absolute essentials for building actors. A Kafka integration doesn't quite feel like it matches that description..

adamhathcock:
I agree about plugging in things that are cool.  I don't see why Kafka has to be "integrated."

cpx86:
So should we drop this idea then?

rogeralsing:
Yes

PR #246, opened 2017-06-14
Make a devbuild target.

adamhathcock:
It was annoying to have my csprojs be modified when I used this script.  I thought it would be helpful to have a target that was explicitly just restore/build/tests for dev on my local machine.

run: `./build.sh -t devbuild`

I removed the task dependencies to allow reordering with new targets.  The default one runs in the old order now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,8 @@ language: csharp
 solution: ProtoActor.sln
 matrix:
   include:
-    - dotnet: 1.0.0-preview4-004233
+    - dotnet: 1.0.4
       mono: none
       env: DOTNETCORE=1
 script:
-  - dotnet restore
-  - dotnet build
-  - dotnet test ./tests/Proto.Actor.Tests/Proto.Actor.Tests.csproj
+  - ./build.sh
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,6 +8,7 @@ environment:
 before_build:
 - nuget install Cake -ExcludeVersion
 build_script:
+- .\Cake\Cake.exe --target=Restore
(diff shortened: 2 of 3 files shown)


cpx86:
That has been annoying me for a while too :) How about inverting it? So the default is a dev build and we do the other tasks in a special CI build instead?

adamhathcock:
Looks like we don't even need a build for CI.  It runs the custom steps.  So I just made the default be what devbuild was.

I tried to clean up the CI a little.  Not sure if Travis needs some work to make it not build full framework on linux.

Do we not want to run CI on pull requests?

cpx86:
Ah, nice. Didn't think of that.

We can actually delete the Travis file. Must be a leftover from when I tried (and failed) to get the CI build to work there.

And yes, CI on PRs would definitely be nice :)

adamhathcock:
An admin would have to add the Appveyor build to protected branches to make it work, I believe.

Travis should work if the cake script was adjusted to not try to build mono on it.  That's what I do for SharpCompress: https://github.com/adamhathcock/sharpcompress/blob/master/build.cake#L14

I can just remove travis if you'd like.

rogeralsing:
Me and @cpx86 have been discussing this here and opt for putting this on hold for now.
While having the same abstraction (PID) for any message based tech, it might very well be a bad abstraction, as in this case where we are forced to do major changes to the API just to add a bit of more guarantees.

if anyone wants to pass a Kafka topic name around, that could just as well be sent as a string or custom message and we could use a custom abstraction for Kafka.

This does not mean we shouldnt go this way, but lets wait for a better usecase or more usecases and evaluate if it makes sense or if one should just use whatever abstraction those cases need there and then

adamhathcock:
@rogeralsing is that message on the wrong pull request?

rogeralsing:
Yes it is :P my bad

cpx86:
I tried adding pull requests to the CI webhook. Let's see if it works :)

PR #247, opened 2017-06-16
fix ConsulProviderOptions to allow override settings

raskolnikoov:
The ConsulProvider class task **RegisterMemberAsync** always resets the provided ConsulProviderOptions settings so no overrides are allowed that is passed  upon instance creation.

### DIFF
--- a/src/ClusterProviders/Proto.Cluster.Consul/ConsulProvider.cs
+++ b/src/ClusterProviders/Proto.Cluster.Consul/ConsulProvider.cs
@@ -15,10 +15,25 @@ namespace Proto.Cluster.Consul
 {
     public class ConsulProviderOptions
     {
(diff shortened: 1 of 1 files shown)


raskolnikoov:
@cpx86 when this is merged, maybe a new nuget package should be released since this is a bug.

cpx86:
Small stylistic idea - how about moving the default values from the `ConsulProvider` constructor into default values on `ConsulProviderOptions`? Like this:
`public TimeSpan ServiceTtl { get; set; } = TimeSpan.FromSeconds(3);`

Might also be good to add XML docs to the properties to show what the default values are?

raskolnikoov:
@cpx86 good one, I have updated to code. :tada:

cpx86:
Awesome! :)

PR #249, opened 2017-06-18
ProtoGrain generator update

raskolnikoov:
It works simple but does the job. Here are two usecases:

Example 1:
dotnet protograin.dll "MyProtoFile.Proto"

Will output a csharp class named MyProtoFile.cs

Example 2:
dotnet protograin.dll "c:\test\PathToMyProtoFile.Proto" "OutputNameOfCSharpClass"

Will output a csharp class named OutputNameOfCSharpClass.cs


rogeralsing:
Ohhh awesome :-)

PR #260, opened 2017-06-22
Forgot to put PatchVersion for full appveyor build

adamhathcock:
Also removed the travis config


cpx86:
Almost there! Seems that `APPVEYOR_REPO_BRANCH` is set to `(no branch)` on PR builds. We could just skip the version patching and nuget pack/publish steps (in the Cake script) if that value is there.

PR #261, opened 2017-06-23
Custom serializers

rogeralsing:
This PR adds initial support for custom serializers.
There is now an ISerializer interface in the remote package.
There is also a list of serializers available in the Serializer class. currently the only supported serializer is Protobuf, but now we have the ability to add more.

Each serializer gets an ID which is then used in the payload to identify what serializer to use.

This does not work like serializers in Akka, where each type or base type can have their own serializer. this is are used for all objects. 
The purpose of this is to allow multiple serializers to be used at the same time for the same message.
e.g. We default to Protobuf for everything, but using JSON for CLI integration.
Or replace the default serializer completely with e.g. Wire for .NET based systems.

### DIFF
--- a/src/Proto.Remote/EndpointReader.cs
+++ b/src/Proto.Remote/EndpointReader.cs
@@ -4,6 +4,7 @@
 //  </copyright>
 // -----------------------------------------------------------------------
 
+using System;
 using System.Threading.Tasks;
 using Grpc.Core;
 using Grpc.Core.Utils;
@@ -28,7 +29,8 @@ namespace Proto.Remote
                     var target = new PID(ProcessRegistry.Instance.Address, targetName);
                     var sender = envelope.Sender;
                     var typeName = typeNames[envelope.TypeId];
-                    var message = Serialization.Deserialize(typeName, envelope.MessageData);
+
+                    var message = Serialization.Deserialize(typeName, envelope.MessageData, envelope.SerializerId);
 
                     if (message is Terminated msg)
                     {
--- a/src/Proto.Remote/EndpointWatcher.cs
+++ b/src/Proto.Remote/EndpointWatcher.cs
@@ -56,7 +56,7 @@ namespace Proto.Remote
                     _watched[msg.Watcher.Id] = null;
 
                     var w = new Unwatch(msg.Watcher);
-                    RemoteProcess.SendRemoteMessage(msg.Watchee, w);
+                    RemoteProcess.SendRemoteMessage(msg.Watchee, w,Serialization.DefaultSerializerId);
                     break;
                 }
                 case RemoteWatch msg:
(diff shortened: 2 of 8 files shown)


src/Proto.Remote/EndpointReader.cs
```diff
@@ -28,7 +29,8 @@ public class EndpointReader : Remoting.RemotingBase
                     var target = new PID(ProcessRegistry.Instance.Address, targetName);
                     var sender = envelope.Sender;
                     var typeName = typeNames[envelope.TypeId];
-                    var message = Serialization.Deserialize(typeName, envelope.MessageData);
+
+                    var message = Serialization.Deserialize(typeName, envelope.MessageData, envelope.SerializerId);
```
rogeralsing: New param in the remote proto definition, `serializer_id`.

src/Proto.Remote/EndpointWatcher.cs
```diff
@@ -56,20 +56,17 @@ public Task ReceiveAsync(IContext context)
                     _watched[msg.Watcher.Id] = null;
 
                     var w = new Unwatch(msg.Watcher);
-                    RemoteProcess.SendRemoteMessage(msg.Watchee, w);
+                    RemoteProcess.SendRemoteMessage(msg.Watchee, w,Serialization.DefaultSerializerId);
```
rogeralsing: When sending remote messages explicitly w/o using a PID, we need to state what serializer to use

src/Proto.Remote/EndpointWriter.cs
```diff
@@ -62,20 +62,21 @@ public EndpointWriter(string address, IEnumerable<ChannelOption> channelOptions,
                             targetNameList.Add(targetName);
                         }
 
-                        var typeName = rd.Message.Descriptor.File.Package + "." + rd.Message.Descriptor.Name;
+                        var typeName = Serialization.GetTypeName(rd.Message, rd.SerializerId);
```
rogeralsing: offload the construction or resolution of typenames to the active serializer

src/Proto.Remote/RemoteProcess.cs
```diff
@@ -64,14 +64,17 @@ public static void SendRemoteMessage(PID pid, object msg)
...
-        public RemoteDeliver(IMessage message, PID target,PID sender)
+        public RemoteDeliver(object message, PID target,PID sender,int serializerId)
         {
             Message = message;
             Target = target;
             Sender = sender;
+            SerializerId = serializerId;
         }
-        public IMessage Message { get;  }
+        public object Message { get;  }
```
rogeralsing: Remote delivered messages do no longer need to be of IMessage type.

cpx86:
Looks good!

raskolnikoov:
Are there some example apps ?

PR #262, opened 2017-06-24
Wire serializer

rogeralsing:
This PR adds the ability to replace the protobuf serialization with Wire.
This allows .NET systems to pass normal POCO objects w/o defining a proto contract.

cc @raskolnikoov 

In order to use, place this code in the start up of your nodes:

```csharp
Serialization.RegisterSerializer(new WireSerializer(), true);
```

It should be noted, that in it's current form, this comes with quite a bit of performance penalty compared to the default serializer.
The reason for this is that Wire needs to roundtrip throough a Stream when serializing/deserializing, which adds extra cost compared to the raw ByteString as protobuf uses.

I've added a copy of the RemoteBenchmark specific to the Wire serializer.
It shows a throughput of ~ 1.3 mil msg/sec compared to the ~2.4 mil msg/sec using the default Protobuf serializer.


rogeralsing:
Seems like there is some flakyness in one of the receive timeout tests:
`ReceiveTimeoutTests.receive_timeout_received_within_expected_time`

src/Proto.Remote/RemoteProcess.cs
```diff
@@ -46,19 +46,12 @@ private void Send(PID _, object msg)
             }
         }
 
-        public static void SendRemoteMessage(PID pid, object msg,int serializerId)
+        public static void SendRemoteMessage(PID pid, object msg, int serializerId)
         {
             var (message, sender, _) = Proto.MessageEnvelope.Unwrap(msg);
 
-            if (message is IMessage protoMessage)
```
rogeralsing: Removing leftover type assertion for protobuf

raskolnikoov:
Sweet!

tomliversidge:
ReceiveTimeout tests could be flaky as they use the world renowned harbinger of doom Thread.Sleep 

Can increase the sleep values to be more slow and stable rather than fast and flaky

raskolnikoov:
Is this ready?

rogeralsing:
@raskolnikoov yes

Jesse1205:
Why delete the wire-serializer branch? it looks good

cpx86:
@Jesse1205 it was already merged :)

PR #269, opened 2017-06-26
Automatically push to NuGet on push to master

cpx86:
This PR removes the dependency on me to publish master packages to the official NuGet feed 😅

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -16,7 +16,7 @@ before_build:
 
 build_script:
 - .\Cake\Cake.exe --target=PrintEnvironment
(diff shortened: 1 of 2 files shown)


cpx86:
Oh, two things:
1. This hasn't been tested yet, for obvious reasons :)
2. Also changed the version pattern for pre-release so the build number precedes the branch name. That way the version ordering should be a bit more logical.

cpx86:
Damnit, I broke something...

cpx86:
There, now all seems to be well again :)

adamhathcock:
Not sure if the PR detection should be passed in as a parameter or not.  Just feels weird to use that env var directly while the others are passed in.

Not a big deal though.

cpx86:
Yeah that should be moved out.

tomliversidge:
Is this good to merge?

tomliversidge:
I squashed a lot of these commits @cpx86 hope you don't mind (you'll need to reset to remote locally if you work again on this branch)- trying to force another build on this branch

PR #274, opened 2017-07-04
remove unused project references from Proto.Persistence

tomliversidge:
removes Proto.Actor and Proto.Mailbox as they were not being used anywhere


raskolnikoov:
Nice!

Issue #186, opened 2017-04-21
Persistence API - Separation of Events and Snapshots


tomliversidge:
As discussed in https://github.com/AsynkronIT/protoactor-dotnet/issues/137 and https://github.com/AsynkronIT/protoactor-dotnet/pull/136 there is an idea to separate out the concept of events vs snapshots, at different levels: 

1. At the provider level, i.e. break out IProviderState into IEventState and ISnapshotState. This would allow different stores to be used for each if desired.
2. At the actor level, i.e. explicitly opt-in to either Events, Snapshotting or both. This is a bigger change due to the coupling of the index shared between Events and Snapshots.

rogeralsing:
Can we close this? this has been fixed, right?

tomliversidge:
Hmmm.. 2. was done, but we still have a single interface for persistence - IProviderState, and we can still only have a single store for both snapshots and events. Having a single interface also makes it messier to decorate if someone wanted to. Don't think there is a particular demand to split it out yet though so we could always close it and re-open later if needed?

raskolnikoov:
I also think we can close this for now and open a new tricket if users really want it separated. 👍

PR #275, opened 2017-07-09
allow separate event store and snapshot stores

tomliversidge:
Minimal set of changes to allow different storage for events and snapshots. If you don't care there's no impact, as IProvider implements both. 

Arguably makes the Persistence API and internal code a little cleaner

closes https://github.com/AsynkronIT/protoactor-dotnet/issues/186

### DIFF
--- a/examples/Persistence/Persistence/Program.cs
+++ b/examples/Persistence/Persistence/Program.cs
@@ -37,7 +37,7 @@ class Program
 
         public MyPersistenceActor(IProvider provider)
         {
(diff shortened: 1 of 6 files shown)


raskolnikoov:
@tomliversidge can this be ready to merge soon?

tomliversidge:
Looks like there's a problem needs fixing first:

error CS0246: The type or namespace name 'Provider' could not be found (are you missing a using directive or an assembly reference?) [C:\projects\protoactor-dotnet\tests\Proto.Persistence.Tests\Proto.Persistence.Tests.csproj]

PR #279, opened 2017-07-26
router refactor

### LINKED ISSUE #242: Remove Props from group router methods
There's no reason to pass any `Props` for the group routers, because they don't own their routees. We should refactor the `Router` class so that the group router methods only take a list of routees, and the pool router methods should take a list of `Props` (for the routees) and an `ISupervisorStrategy` to control how the group router supervises it's routees.

This also means that the `RouterActor` needs to changed, because it currently always takes a `routeeProps` argument and it's only used for group routers. We could split the actor into `PoolRouterActor` and `GroupRouterActor` and also move the logic in `IRouterConfig.Started` into the `Started` message handler for each router actor.

rogeralsing:
Fixes #242 and aligns the implementation to Kotlin.

cc @AsynkronIT/developers


examples/Router/Program.cs
```diff
@@ -63,7 +63,6 @@ private static void Main()
         private static void TestBroadcastGroup()
         {
             var props = Router.NewBroadcastGroup(
-                MyActorProps,
```
rogeralsing: props removed from Group routers as they make no sense there

src/Proto.Remote/EndpointReader.cs
```diff
@@ -50,7 +50,10 @@ public class EndpointReader : Remoting.RemotingBase
                     }
                     else
                     {
-                        target.Request(message, sender);
+                        if (envelope.Sender != null)
```
rogeralsing: This change should not have been in this specific PR, but it is a small optimization in Remote, not allocating a MessageEnvelope if there is no sender... carry on ;-)

cpx86:
Looks good! Lots of nice improvements I've wanted for a while :)

src/Proto.Remote/EndpointReader.cs
```diff
@@ -50,7 +50,10 @@ public class EndpointReader : Remoting.RemotingBase
                     }
                     else
                     {
-                        target.Request(message, sender);
+                        if (envelope.Sender != null)
+                            target.Request(message, envelope.Sender);
+                        else
```
adamhathcock: Nitpick: always have braces, yes?

PR #281, opened 2017-07-30
Mailbox tuning

rogeralsing:
This PR implements the same mailbox tweak as I did on the Kotlin version.
TLDR, it maintains an atomic message counter for system messages.
So instead of asking the backing queue for empty state, it checks the counter.

This raised throughput from 80 to 90 mil (ish)
```
Is Server GC True
Dispatcher              Elapsed         Msg/sec
300                     202             79207920
400                     174             91954022
500                     218             73394495
600                     192             83333333
700                     177             90395480
800                     176             90909090
900                     174             91954022
```

ps. also some number tweaking in the inprocbenchmark for more consistent numbers

### DIFF
--- a/examples/InprocessBenchmark/Program.cs
+++ b/examples/InprocessBenchmark/Program.cs
@@ -33,23 +33,17 @@ class Program
             var echos = new PID[clientCount];
             var completions = new TaskCompletionSource<bool>[clientCount];
 
-            var echoProps = FromFunc(ctx =>
-            {
-                switch (ctx.Message)
-                {
-                    case Msg msg:
-                        msg.Sender.Tell(msg);
-                        break;
-                }
-                return Done;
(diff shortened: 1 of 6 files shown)


rogeralsing:
meh. apparently I was in my router-refactor branch when creating the new branch..

rogeralsing:
Also updated the mailbox to bounded mailbox, as we use that in the Go benchmark too.

This leads to the following stats:

```
Dispatcher              Elapsed         Msg/sec
300                     222             72072072
400                     145             110344827
500                     130             123076923
600                     129             124031007
700                     128             125000000
800                     130             123076923
900                     128             125000000
```

I also did some testing before this, netstandard 2.0 bumps perf by ~1.4 times.
if this still applies to the above numbers, it would bring us up to 175 mil msg/sec.

Issue #283, opened 2017-08-03
Support eventstore-driven event counter (index)


alexeyzimarev:
Currently, the event index, which is used to read events and create/read snapshots, is driven by the Proto.actor. This creates issues when trying to support immutable event stores (like geteventstore.com) because there the event index is driven by the store. Since it is impossible to remove anything there, the only way to support `DeleteEventsAsync` and `DeleteSpanshotsAsync` is to soft-delete streams, new events will get version not starting from zero. With current model, when event index starts from zero and event version returned from the store cannot be used, makes it impossible to support immutable event stores at least with snapshotting, since querying by index will only be possible by traversing all events, so snapshotting makes no sense there.

tomliversidge:
I'd like to see what the this would look like. Persistence is a pretty small class so maybe know up a demo to discuss?

Something I was wondering - would you ordinarily save a snapshot in the same stream as the events, or use a different stream?

tomliversidge:
@jrouaix weren't you working on an event store provider too?

alexeyzimarev:
I will check if I can actually make a change. I spent some time on the provider itself, got stuck after realising that snapshot index would require the event index, which I have not found where to store yet - either I need a wrapping class or store the version in the metadata.

Concerning snapshot storage - I would keep it in a separate stream. This would mean that to read the last snapshot I just need to read one event from the end of the stream, which is very fast. From there I would know the snapshot version and can read events from that version up to the end of the event stream.

As you can find [in my code](https://github.com/alexeyzimarev/Proto.Persistence.EventStore/blob/master/src/Proto.Persistence.EventStore/EventStoreProvider.cs), I offer a possibility to specify custom stream names with default strategy adding `snapshot-` before the actor name, which normally would be the event stream name.

tomliversidge:
I've just remembered one of the first things i did with Proto.Actor was try and write an event store client. man, i'd forgotten all about that. it is **very** out of date now: https://github.com/tomliversidge/protoactor-dotnet/blob/EventStorePersistenceProvider/src/PersistenceProviders/Proto.Persistence.EventStore/EventStoreProviderState.cs

And this:

```
public async Task DeleteSnapshotsAsync(string actorName, long fromIndex)
        {
            throw new NotSupportedException("EventStore is an immutable database that does not support deletes");
        }
```

:)

alexeyzimarev:
[...]

PR #288, opened 2017-08-10
#283 Support for server-drive event index (stream version)

alexeyzimarev:
Check my attempt to solve it. I am not sure about the API change impact. It only concerns providers since call results were not used anyway, so all usages are not affected.

### DIFF
--- a/src/PersistenceProviders/Proto.Persistence.Sqlite/SqliteProvider.cs
+++ b/src/PersistenceProviders/Proto.Persistence.Sqlite/SqliteProvider.cs
@@ -69,7 +69,7 @@ namespace Proto.Persistence.Sqlite
             }
         }
 
-        public Task GetEventsAsync(string actorName, long indexStart, long indexEnd, Action<object> callback)
+        public async Task<long> GetEventsAsync(string actorName, long indexStart, long indexEnd, Action<object> callback)
         {
             try
             {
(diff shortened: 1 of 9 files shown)


tomliversidge:
Looks ok to me!

raskolnikoov:
Im gonna have a look to just in a sec

src/PersistenceProviders/Proto.Persistence.Sqlite/SqliteProvider.cs
```diff
@@ -69,7 +69,7 @@ public SqliteProvider(string datasource = "states.db")
             }
         }
 
-        public Task GetEventsAsync(string actorName, long indexStart, long indexEnd, Action<object> callback)
+        public async Task<long> GetEventsAsync(string actorName, long indexStart, long indexEnd, Action<object> callback)
```
raskolnikoov: remove async

src/PersistenceProviders/Proto.Persistence.Sqlite/SqliteProvider.cs
```diff
@@ -84,14 +84,13 @@ public Task GetEventsAsync(string actorName, long indexStart, long indexEnd, Act
                     {
                         callback(JsonConvert.DeserializeObject<object>(item.EventData, new JsonSerializerSettings { TypeNameHandling = TypeNameHandling.Auto }));
                     }
+                    return items.Any() ? items.LastOrDefault().EventIndex : -1;
```
raskolnikoov: should looks like this instead?

return Task.FromResult(items.Any() ? items.LastOrDefault().EventIndex : -1);

alexeyzimarev: Yeah, if you remove `async`...

Issue #290, opened 2017-08-14
Sender middleware does not receive current message


cpx86:
If you run the Middleware sample, you get the following output:
```
middleware 1 enter Proto.Started:Proto.Started
middleware 2 enter Proto.Started:Proto.Started
actor got Proto.Started:Proto.Started
middleware 2 exit Proto.Started:Proto.Started
middleware 1 exit Proto.Started:Proto.Started
sender middleware 1 enter System.String:
sender middleware 2 enter System.String:
sender middleware 2 exit System.String:
sender middleware 1 exit System.String:
middleware 1 enter System.String:hello
middleware 2 enter System.String:hello
TraceID = 1000
SpanID = 2000
ParentSpanID =
actor got System.String:hello
middleware 2 exit System.String:hello
middleware 1 exit System.String:hello
```

`ISenderContext.Message` is for some reason never assigned. Not even sure where it should be assigned? ping @AsynkronIT/developers

masteryee:
The fix has been merged into dev. The output is now:

```
middleware 1 enter Proto.Started:Proto.Started
middleware 2 enter Proto.Started:Proto.Started
actor got Proto.Started:Proto.Started
middleware 2 exit Proto.Started:Proto.Started
middleware 1 exit Proto.Started:Proto.Started
sender middleware 1 enter System.String:hello
sender middleware 2 enter System.String:hello
sender middleware 2 exit System.String:hello
sender middleware 1 exit System.String:hello
middleware 1 enter System.String:hello
middleware 2 enter System.String:hello
TraceID = 1000
SpanID = 2000
ParentSpanID =
actor got System.String:hello
middleware 2 exit System.String:hello
middleware 1 exit System.String:hello
```

rogeralsing:
Closing

PR #293, opened 2017-08-16
fix: print envelope message in sender middleware example

masteryee:
Fix for issue #290 - the message in the sender middleware is wrapped in a MessageEnvelope per the [Middleware documenation](http://proto.actor/docs/dotnet/middleware).


cpx86:
What do we even need the Message property for? @rogeralsing do you remember what we were thinking when we designed that part? :)

rogeralsing:
context.Message and context.Headers are the current message and headers known by the acting actor, the envelope contains the outgoing message and headers as a result.

current message --> behavior --> outgoing message

rogeralsing:
👍

Issue #153, opened 2017-04-10
Tests are flaky / failing


tomliversidge:
https://ci.appveyor.com/project/protoactor-ci/protoactor-dotnet/build/dev-199/tests
https://ci.appveyor.com/project/protoactor-ci/protoactor-dotnet/build/process-redesign-195/tests
https://ci.appveyor.com/project/protoactor-ci/protoactor-dotnet/build/dev-190/tests

etc...

alexpantyukhin:
@tomliversidge  Is it still reproducing?

tomliversidge:
Hmmm... not sure. We've had a few failures in the last week on the build server, but I'm not sure if they are about mailbox issues

rogeralsing:
Agree, there seems to be random failures going on in the buildserver.
I guess it's the usual non deterministic stuff, slow server, unfair scheduler, tasks are timing out

cpx86:
The Mailbox tests fail because we use `Task.ContinueWith` which will asynchronously schedule the continuation task. If we could inject a synchronous `TaskScheduler` (perhaps via the `IDispatcher` interface) then we should be able to get rid of the time-dependence and `Thread.Sleep`.

tomliversidge:
I _think_ I've been able to remove some of the Thread.Sleeps in the mailbox tests by Wait()-ing in the IDispatcher in TestMailboxHandler.cs:

```
public void Schedule(Func<Task> runner)
        {
            runner().Wait();
        }
```

This is what we do in TestDispatcher.cs already

masteryee:
Several builds have passed successfully with no test failures after the merge. Please close this issue.

PR #295, opened 2017-08-21
test: fix timing issue in mailbox/actor tests for #153

masteryee:
Replaces Thread.Sleep usages in the tests with more deterministic code that waits for the mailbox to complete processing before Asserting.

Originally fixed only EscalateFailureTests and MailboxSchedulingTests based on the example build failures, but also updated ReceiveTimeoutTests as one of the tests failed randomly in VS.

### DIFF
--- a/tests/Proto.TestFixtures/TestMailboxHandler.cs
+++ b/tests/Proto.TestFixtures/TestMailboxHandler.cs
@@ -3,12 +3,14 @@ using System.Collections.Generic;
 using System.Threading.Tasks;
 using Proto.Mailbox;
 using System.Threading;
+using System.Collections.Concurrent;
 
 namespace Proto.TestFixtures
 {
     public class TestMailboxHandler : IMessageInvoker, IDispatcher
     {
-        private AutoResetEvent _onScheduleCompleted;
+        private ConcurrentQueue<TaskCompletionSource<int>> _taskCompletionQueue =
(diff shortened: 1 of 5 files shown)


cpx86:
Won't the use of an AutoResetEvent make the the test run indefinitely in case the functionality breaks?

rogeralsing:
I'd vote for using `TaskCompletionSource` in a similar way.
and then instead of `.WaitOne` we could just `await` the `tsc.Task`

But also, like @cpx86 said, we should use timeouts also.

tests/Proto.TestFixtures/TestMailboxHandler.cs
```diff
@@ -2,11 +2,16 @@
...
 using System.Threading.Tasks;
 using Proto.Mailbox;
+using System.Threading;
+using System.Collections.Concurrent;
 
 namespace Proto.TestFixtures
 {
     public class TestMailboxHandler : IMessageInvoker, IDispatcher
     {
+        private ConcurrentQueue<TaskCompletionSource<int>> _taskCompletionQueue =
```
masteryee: This allows for better parity between Schedule() and ResumeMailboxProcessingAndWaitAsync in case ResumeMailboxProcessingAndWaitAsync needs to be called more than once.

masteryee:
Updated based on comments. More async/timeout friendly.

rogeralsing:
:+1: Great!

PR #300, opened 2017-08-27
Change HashRing hash algorithm to FNV1A32 to match with Go.

PotterDai:


PotterDai:
The build error seems to be same as this one https://github.com/AsynkronIT/protoactor-dotnet/pull/295, and has no relationship with the FNV1A32 hash and the cluster module.

rogeralsing:
Looks great!, failed test is unrelated.

PR #302, opened 2017-08-31
Raise a EndPointConnected Event to change EndpointWatcher to corrent state after reconnected.(#301)

### LINKED ISSUE #301: Cluster has some problems.
1. In ClusterGrainHelloWorld example, when i start 2 server node and a client which call Tell per second, after client connected to a server, I close the connected server node, the problem is some times, client can connect to the other server node, but some times can't.  ------------> this problem is caused by the implementation of MemberListActor:ReceiveAsync, when a address is not alive, it has not removed from _members. 

2. In MemberListActor.cs:46, foreach (var (address, old) in _members), in the loop, Notify is called, but inside Notify, _members is modified by _members.Remove(old.Address); and this will raise an exception.

3. In ClusterGrainHelloWorld example, when i start 2 server node and a client which call Tell per second, after client connected to a server(s1), I close the connected server node(s1), then client will connected to another server(s2), then I start s1 and close s2, after client connected to s1, a Terminated message will send per second form client's Watcher and pidcache's entity of s2 will add/remove per second, I am not clear why this happened.

yindongfei:
Fixes #301 
all the problem mentioned is fixed.


rogeralsing:
Looks great!

PR #308, opened 2017-09-01
DynamoDB Persistence Provider

urossmolnik:


urossmolnik:
I was testing two ways I could implement deletion in DynamoDB.  
I'll remove commented code later today.

Issue #303, opened 2017-09-01
Cluster Performance


yindongfei:
```
        Task.Run(async () =>
        {
            var t = DateTime.Now;

            List<Task<HelloGrainTestResponse>> l = new List<Task<HelloGrainTestResponse>>();
            for (int i = 0; i < 10000; i++)
            {
                l.Add(client.SayHelloTest(new HelloGrainTestRequest()));
            }

            await Task.WhenAll(l);

            Console.WriteLine("time: {0}", (DateTime.Now - t).TotalMilliseconds);
        });
```

I run this and performance is 1k message per second. maybe a cache can solve the problem?

![image](https://user-images.githubusercontent.com/2337398/29956207-0968ad8c-8f18-11e7-806d-0b4702047cd5.png)

yindongfei:
call a router to get pid for each RPC is too expensive. a cache for pid location can saved locally and updated when a pid can not found during a RPC.

rogeralsing:
What are your numbers if you remove the router in-front of the pid cache?

rogeralsing:
Related #304

yindongfei:
after remove router, performance up to 25 times faster than before, 25k messages per seconds. 

![image](https://user-images.githubusercontent.com/2337398/29991585-5d331d9a-8fbc-11e7-9492-7bb3450ff5b0.png)
![image](https://user-images.githubusercontent.com/2337398/29991591-7ba97b7a-8fbc-11e7-885f-3e9e68e5e77b.png)
![image](https://user-images.githubusercontent.com/2337398/29991593-8d526dbe-8fbc-11e7-9a07-bf3389814bba.png)
![image](https://user-images.githubusercontent.com/2337398/29991604-b774fb5c-8fbc-11e7-939e-089926714670.png)
![image](https://user-images.githubusercontent.com/2337398/29991608-cb99c176-8fbc-11e7-8f33-57fc5f82e0e8.png)

rogeralsing:
Wonderful :-)

Yet again, it shows the importance of measuring everything.

rogeralsing:
Can you make a PR for this?

yindongfei:
I have write a async rpc implementation using c# based on dotnetty and protobuf, 800k rpc calls per second, as you mentioned in #304 , the rpc method name is replaced by a index, and type is not passed.
code can be generated and type code can also be generated. some thing like this(it is used in our production):

```
[...]

PR #309, opened 2017-09-03
Remove the router in-front of the pid cache

yindongfei:
#303


PotterDai:
@rogeralsing 
Should this be changed in Go version as well? in pid_cache.go file from
props := router.NewConsistentHashPool(128).WithProducer(newPidCacheActor())
to
props := actor.FromProducer(newPidCacheActor())
?

rogeralsing:
>Should this be changed in Go version as well?

Yes, we need to keep all implementations as close as possible.

PR #311, opened 2017-09-15
Fix bug: HashAlgorithm not Initialize before ComputeHash.

PotterDai:
Fix bug where HashAlgorithm not being Initialized before computing hash.
This will result in wrong hash number since initial prime number is not applied.


PotterDai:
Seems unnecessary when checking the source code, since Initialize is being called within ComputeHash. = ) Close the request.

PR #312, opened 2017-09-16
Replace HashRing method with Rendezvous

PotterDai:
This is the same implementation of https://github.com/AsynkronIT/protoactor-go/pull/176 but in dotnet version.


rogeralsing:
Very nice!

PR #313, opened 2017-09-18
Graceful shutdown and others

PotterDai:
Same as https://github.com/AsynkronIT/protoactor-go/pull/177
This pull request contains multiple changes... not sure how to subdivide it but I will try to make all the changes clear here.

- Rendezvous (already has its own PR) https://github.com/AsynkronIT/protoactor-dotnet/pull/312
- When transferring ownership, the partition with the new ownership will only cache the transferred actor, but not watch it. So if that actor dies, the partition with new ownership will not be notified. This is fixed in f415bab
- PIDCache does not subscribe to membership status event, so if member status changed, PIDCache will not clear the cache of that member accordingly. This is fixed in  fee981c
- Watcher can only watch 1 remote watchee per endpoint, as is mentioned in https://github.com/AsynkronIT/protoactor-go/issues/172 and https://github.com/AsynkronIT/protoactor-go/pull/175. This is fixed in  3cc6663
- A simple implementation of graceful stop (currently works, but should be refined). Got one issue, GRPC does not want to stop and still try to figure out why. Now as a workaround, will kill GRPC server after 10 seconds. This is implemented in  a5691fa 
    - Move Remote.Start to Cluster.Start, this is to follow go version.  95e3cea
    - Using Behaviour mechanism to handle EndpointManager stop (It's a supervisor, so call stop will restart it) 86d744c
    - Update Consul Provider and IClusterProvider to have Shutdown/Deregister methods.  3d3856d

### DIFF
--- a/src/Proto.Remote/EndpointReader.cs
+++ b/src/Proto.Remote/EndpointReader.cs
@@ -14,7 +14,7 @@ namespace Proto.Remote
 {
     public class EndpointReader : Remoting.RemotingBase
     {
-        private bool suspended;
+        private bool _suspended;
         
         public override Task<ConnectResponse> Connect(ConnectRequest request, ServerCallContext context)
         {
@@ -29,7 +29,7 @@ namespace Proto.Remote
         {
             await requestStream.ForEachAsync(batch =>
             {
-                if (suspended)
+                if (_suspended)
                     return Actor.Done;
                 
                 var targetNames = new List<string>(batch.TargetNames);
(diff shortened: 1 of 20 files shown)


src/Proto.Remote/EndpointReader.cs
```diff
@@ -14,6 +14,8 @@ namespace Proto.Remote
 {
     public class EndpointReader : Remoting.RemotingBase
     {
+        private bool suspended;
```
rogeralsing: cosmetics, but we use `_suspended`

PotterDai: Just fixed! Thx for pointing out = )

Issue #314, opened 2017-09-20
Suggestion: extend ActorPIDResponse


PotterDai:
So there are 2 suggestions hope to be discussed:
1. ActorPIDResponse can contain a status code, and allow null PID to be passed.
2. Partition select Activator using Round Robin instead of Random

Reason:
If Activator for some reason cannot spawn the actor(for example: starting/stopping), currently it can only throw exception and restart, this will make Partition actor be blocked for 5 seconds until timeout, not processing any new spawn request. Most of time this timeout can be avoided by just passing it to client who requested that PID, or move on to the next Activator.

Changes:
Activator can return ActorPIDResponse with status code
0->OK, 1->UNAVAILABLE, 2,3,4...->WAIT,etc, etc...(User defined response code)
Partition will react to status code accordingly:
- Status: OK => return result PID to client
- Status: UNAVAILABLE => use RoundRobin to get the next activator.
- Status: WAIT... => forward status code to client to let client wait/or decide what to do.

PR #319, opened 2017-09-22
ActorPIDResponse will return ResponseStatusCode

PotterDai:
This relates to https://github.com/AsynkronIT/protoactor-dotnet/issues/314

If an Activator failed to activate an actor, Activator now can only throw exception, then restart. This is not good since Partition-actor will hang 5 seconds until timeout, at this time period no further ActorPIDRequest will be processed.
So even if Activator need to crash, it should at least return a response to Partition-Actor.


src/Proto.Cluster/Cluster.cs
```diff
@@ -67,13 +67,19 @@ private static (string host, int port) ParseAddress(string address)
...
-        public static Task<PID> GetAsync(string name, string kind) => GetAsync(name, kind, CancellationToken.None);
+        public static Task<(PID, ResponseStatusCode)> GetAsync(string name, string kind) => GetAsync(name, kind, CancellationToken.None);
 
-        public static async Task<PID> GetAsync(string name, string kind, CancellationToken ct)
+        public static async Task<(PID, ResponseStatusCode)> GetAsync(string name, string kind, CancellationToken ct)
         {
             var req = new PidCacheRequest(name, kind);
-            var res = await PidCache.Pid.RequestAsync<ActorPidResponse>(req, ct);
-            return res.Pid;
+            var res = await PidCache.Pid.RequestAsync<PidCacheResponse>(req, ct);
```
rogeralsing: Is it possible that the ProcessNameExists bubbles up here?
And if so, should we not return the existing PID in that case?

To me it looks like if two different threads/nodes at the same time tries to `GetAsync("SomeActor","SomeKind")` one of them might end up getting ProcessNameExists back, right?

PotterDai: Hi, you are right! Actually in case of ProcessNameExists, req.Pid will return null since Activator will only assign status code. 
https://github.com/PotterDai/protoactor-dotnet/blob/1063da1bf8c08d4bce461f86c9205fba14440bbd/src/Proto.Remote/Activator.cs#L45-L49

rogeralsing: Maybe it is enough to extend ActorPidResponse with the status code and do what you say?
This way, GetAsync would behave as expected no matter if it is OK or ProcessNameExists

PotterDai: OK, so I think it should be good to go! = ) Any exception now will return with a null PID and a not-OK StatusCode, user can decide what to do accordingly.

rogeralsing:
cc @PotterDai do you have a twitter account?

PotterDai:
@rogeralsing  Yep, it's @potterdai_ . But actually I have never used it...

PR #331, opened 2017-09-29
Revision for cluster

PotterDai:
This contains 3 fixes:
- Remove unneccesary MemberAddress:PID map I added in previous PR.
- Fix ActorPidRequest do not timeout bug
- Add a simple GRPC retry interval (2 seconds), future should be changed to exponential backoff


rogeralsing:
Looks great!

PR #333, opened 2017-09-30
Bundle consul and add auto boot to examples

rogeralsing:
This PR bundles Consul into the example source, it also adds the option to auto boot consul in dev mode.
This makes it a lot easier for newcomers to try our examples.

It would be nice if someone could fetch and verify if this works on their machines too.
Currently, this is Windows only as it bundles the win exe.
Maybe bundle other platforms too?

### DIFF
--- a/src/ClusterProviders/Proto.Cluster.Consul/ConsulProvider.cs
+++ b/src/ClusterProviders/Proto.Cluster.Consul/ConsulProvider.cs
@@ -6,6 +6,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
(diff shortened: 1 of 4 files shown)


src/ClusterProviders/Proto.Cluster.Consul/ConsulProvider.cs
```diff
@@ -6,6 +6,7 @@
```
raskolnikoov: I think "StartConsulDevMode should be moved into the cluster demo project instead of having this in the provider library. I guess this will only be used when starting the demo? @rogeralsing

PR #334, opened 2017-10-02
Revision for cluster

PotterDai:
- Fix bug where Future process will not deregister itself from ProcessRegistry on stop, causing memory leak.
- Improve Partition perf. Loop through entire partition dictionary each time an actor terminated is very expensive.

### DIFF
--- a/src/Proto.Actor/Futures.cs
+++ b/src/Proto.Actor/Futures.cs
@@ -55,11 +55,11 @@ namespace Proto
         {
             var env = MessageEnvelope.Unwrap(message);
             
-
             if (env.message is T || message == null)
             {
                 if (_cts != null && _cts.IsCancellationRequested)
                 {
+                    pid.Stop();
                     return;
                 }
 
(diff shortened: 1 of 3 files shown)


src/Proto.Actor/Process.cs
```diff
@@ -40,16 +45,19 @@ internal bool IsDead
 
         protected internal override void SendUserMessage(PID pid, object message)
         {
+            if (IsDead) return;
             Mailbox.PostUserMessage(message);
         }
 
         protected internal override void SendSystemMessage(PID pid, object message)
         {
+            if (IsDead) return;
```
PotterDai: @rogeralsing  Here we got a bit problem. Will there be a lot of performance hit if we check IsDead here?

The reason I'm doing this is I want to set Mailbox to null when stopping, otherwise all the messages in that Mailbox will not be garbage collected and causing memory leak.

Another way is to clear the queue in Mailbox.
What you think?

PotterDai:
@rogeralsing Nevermind, further test indicates that MemoryLeak is purely because Future process not deregister itself from ProcessRegistry on stop.

src/Proto.Actor/Futures.cs
```diff
@@ -55,11 +55,11 @@ protected internal override void SendUserMessage(PID pid, object message)
         {
             var env = MessageEnvelope.Unwrap(message);
             
-
             if (env.message is T || message == null)
             {
                 if (_cts != null && _cts.IsCancellationRequested)
                 {
+                    pid.Stop();
```
rogeralsing: good catch, not sure how this has been missed in the past :-/

PR #335, opened 2017-10-02
WIP: initial example of cluster routers - do not merge

rogeralsing:
This PR contains a first sketch on how cluster routers could be designed.

The reason this does not build on existing router code is that the cluster infrastructure is async, we need to resolve routees and members  async.
Messages are also going to be sent over network, so there is not the same need to avoid the actor mailbox overhead as this is minimal in this case.

### DIFF
--- a/src/Proto.Cluster.Router/ClusterRouter.cs
+++ b/src/Proto.Cluster.Router/ClusterRouter.cs
@@ -1,0 +1,62 @@
+﻿using System.Linq;
+using System.Threading.Tasks;
+using Proto.Mailbox;
+
+namespace Proto.Cluster.Router
+{
+    public class ClusterRouter
+    {
+        public static PID NewClusterBroadcastPool(string name, int poolSize, string kind)
+        {
+            var props = Actor.FromProducer(() => new ClusterBroadcastPool(name, poolSize, kind));
+            var pid = Actor.Spawn(props);
+            return pid;
+        }
+    }
+
+    //TODO: There are two approaches that could be viable for cluster routers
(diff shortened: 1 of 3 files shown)


rogeralsing:
TODO: There are two approaches that could be viable for cluster routers
This approach generates a name per routee and simply use Cluster.GetAsync on each
This works, but, this is not guaranteed to place routees evenly on different nodes.
Another approach could be to subscribe to cluster members and place actors on each of those
IMO, both are correct in their own ways

src/Proto.Cluster.Router/ClusterRouter.cs
```diff
@@ -0,0 +1,62 @@
...
+    //IMO, both are correct in their own ways
+    public class ClusterBroadcastPool : IActor
+    {
+        private readonly string[] _routees;
+        private readonly string _kind;
+
+        public ClusterBroadcastPool(string name,int poolSize, string kind)
+        {
+            _kind = kind;
+            _routees = Enumerable
```
rogeralsing: This does not generate the actual routees, just the routee names.

src/Proto.Cluster.Router/ClusterRouter.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                case AutoReceiveMessage _:
+                case SystemMessage _:
+                    break;
+                default:
+                    var m = context.Message;
+                    foreach (var routee in _routees)
+                    {
+#pragma warning disable 4014
+                        Cluster.GetAsync(routee, _kind).ContinueWith(t =>
```
rogeralsing: Resolve each name and forward the message to the routee

PR #338, opened 2017-10-11
Updated remote example to use Pid prop

cemremengu:
I guess the API changed since this example


cemremengu:
Struck by Github which made me think that the PR wasnt created

October 11, 2017

18:51 We're currently experiencing high replication delay on our primary database cluster. Recently performed site actions may not appear in the UI.

PR #339, opened 2017-10-11
Fixed broken RemoteActivate example

cemremengu:
Using the `Pid` property now, I guess the API changed since this example


PotterDai:
You are right! Thank you very much!

PR #341, opened 2017-10-12
Revision for membership

PotterDai:
This PR contains several revisions and fixes:
1. New Membership structure, cache members by kinds in MemberListActor. This is implemented 7870dd8 
2. When a node gracefully shutdown, its Partition now will transfer all its ownerships not located in itself to other Partitions, similar to what happened when a member join in. This is revised in 74dfff4, e2acb74
3. Implemented weighted round robin. This is a feature I personally want a lot, and I think it's very useful for other users as well. Now user can define node weight from 1-10(*see notes), and activator will select which node to activate new actor according to the weight. This is implemented in 090d0d2, 9e477bd 
4. Add graceful shutdown to the example.  c2c893d 

Notes:
The current weighted round robin implementation has a weight limit by 10. This limit will be removed in the future.


PotterDai:
@rogeralsing Most changes in this PR are mainly focused on internal fixes and improvements. There are 2 changes does affect a bit of public facing API:

1. Cluster -> user now can start cluster using both Cluster.Start(...) and Cluster.StartWithConfig(...)
Is this a bit messy?

2. In Consul, I store member weight value in key value store with memberID. Currently just using new line to split them.
Is this a bit hacky?

rogeralsing:
I love the direction of this, but I think that maybe we should abstract this somehow.
e.g. weights might be one strategy, cpu and mem metrics might be another.

What do you think about that? would it be possible to direct the node selection strategies to the cluster provider using such above mentioned strategy plugins?

The more of those specifics that could be handled in the provider the better and less complex the proto.cluster code would be.
It would also open up for providers for Zookeeper and ETCD where there are other behaviors and strategies.

It's just a thought but the topic comes up often about weights and resource usage

PotterDai:
@rogeralsing You are right. Actually I was thinking about it but didn't fully come up with a good plan to abstract it out.
What user can do in this PR is to call Cluster.UpdateWeight(...) which will update the weight during runtime. It's functional since user can just update weight repeatedly, but as you said, it could be much better if it is done using an abstraction = ).

Another question, should I separate this weight implementation out of this PR? Since this PR also contains multiple fixes and performance improvements like 
-- New Membership structure
-- Partition transfer all its ownerships not located in itself to other Partitions
-- Add Cluster.ShutDown in example codes
Also in Go --> SystemMessage and AutoReceivedMessage no longer be treated as UnknownMessage.

PotterDai:
@rogeralsing 
OK, so current version should be much better.
2 Major Implementations:
1. [IMemberStrategy](https://github.com/PotterDai/protoactor-dotnet/blob/2cde39e70fca12b524ab21d92857acc283f19a25/src/Proto.Cluster/MemberStrategy.cs)
This will be the interface to provide Partition/Activator strategy.
User now can also implement their own strategy following this [WeightedMemberStrategy example](https://github.com/PotterDai/protoactor-dotnet/tree/2cde39e70fca12b524ab21d92857acc283f19a25/src/MemberStrategies/WeightedMemberStrategy).

2. [IMemberStatusValue](https://github.com/PotterDai/protoactor-dotnet/blob/2cde39e70fca12b524ab21d92857acc283f19a25/src/Proto.Cluster/MemberStatus.cs)
This is for user to create custom status value to be stored in ClusterProvider. For example, CPU usage, total grains, weight, etc. Also check [WeightedMemberStatus](https://github.com/PotterDai/protoactor-dotnet/blob/2cde39e70fca12b524ab21d92857acc283f19a25/src/MemberStrategies/WeightedMemberStrategy/WeightedMemberStatus.cs) as example.

3. [ClusterConfig](https://github.com/PotterDai/protoactor-dotnet/blob/2cde39e70fca12b524ab21d92857acc283f19a25/src/Proto.Cluster/ClusterConfig.cs)
This should be a good start for user to customize Cluster.

I end up didn't set WeightedMemberStrategy as default one to keep Cluster as simple as possible, instead making it as an optional MemberStrategy.

rogeralsing:
This looks great! is it ready to merge now?

PotterDai:
@rogeralsing Just fixed some minor issues, now it should be good to go.

PR #342, opened 2017-10-13
Cluster stability test

rogeralsing:
This PR adds a test harness for the cluster support.

It spawns a client node and then 10 worker nodes and starts communicating with grains on all those nodes


examples/ClusterStabilityTest/Messages/Protos_protoactor.cs
```diff
@@ -0,0 +1,152 @@
...
+
+            async Task<HelloResponse> Inner() 
+            {
+retry:
+                //resolve the grain
+                var (pid, statusCode) = await Cluster.GetAsync(_id, "HelloGrain", ct);
+                if (statusCode == ResponseStatusCode.Unavailable)
+                {
+                    await Task.Delay(100);
+                    goto retry;
```
rogeralsing: Hack:
In the code generated grains, there was no handling of the unavailable status, this causes the client to fail with unavailable exception.

Adding this hacky retry loop solves this.
So some form of retry for x time should definitely be added to the code generator, both c# and go

rogeralsing: I see that we return err in Go:
https://github.com/AsynkronIT/protoactor-go/blob/dev/examples/cluster/shared/protos_protoactor.go#L50

Which might be fine there, but throwing in C# forcing users to try catch to retry feels too heavy.

I think that in both C# and Go, it would be nice to have some form of retry x times mechanism in place to avoid confusion.

e.g. it is likely that a kind is unavailable during cluster startup just because the cluster is forming and the other nodes have not yet registered completely.
allowing the caller to wait for a short period of time and then succeed, would be a better approach than forcing user to do retry

src/ClusterProviders/Proto.Cluster.Consul/ConsulProvider.cs
```diff
@@ -189,7 +189,8 @@ private void UpdateTtl()
                 let memberIdKey = $"{_clusterName}/{v.Service.Address}:{v.Service.Port}"
                 let memberId = GetMemberId(memberIdKey)
                 where memberId != null
-                let passing = Equals(v.Checks[1].Status, HealthStatus.Passing)
+                //Check #1 is the check we have declared, check 0 is the Consul Serf Status
+                let passing = v.Checks.Length > 1 && Equals(v.Checks[1].Status, HealthStatus.Passing)
```
rogeralsing: This fixes the status exception that was seen in some cases. simply a race condition where the custom health check had not yet been registered/become visible to external parties

rogeralsing:
From my tests, these changes made the cluster a lot more stable, I have not yet encountered any issues after this.

rogeralsing:
I Pushed a new commit, at 40 nodes. even though Consul UI reports all nodes as healthy.
`Cluster.GetAsync` returns Unavailable.

rogeralsing:
The cluster now looks stable at 100 nodes.
I've seen a few failed RPC calls. represented as "$" in the  output.

There is however a considerable amount of lag. not sure if this is due to the 100 processes weighing down the machine, or if there is some proto.actor related issue causing this.

With fewer nodes, everything seems to run fine

rogeralsing:
Some observations:

* Using a unique cluster name per run seems to make the cluster more stable.
My guess is that we have an issue where there are members from previous runs registered on the same address/port/clustername which are seen as separate registrations by consul.
This might confuse the cluster topology.

* Waiting for cluster to stabilize before start communicating with grains makes the cluster behave perfect.
My guess here is that when there is a surge of new nodes joining, this messes up the partition actor as new members join taking of the same hash slices.

Two sides to this, it's unlikely that this happens in reality, members might be added or removed a few at a time, but adding 50 or more members in one go is unlikely unless running at huge scale.

Even if the above sentences are true, we should still try to mitigate this and fix this as good as possible.

PR #344, opened 2017-10-17
Memberlist Get functions can be concurrent

PotterDai:
@rogeralsing  From my local test, it seems completely ok for MemberList to be concurrent as long as we lock it when triggering ClusterTopologyEvent. Moreover, it gives about 40% performance gain for spawning new actors through cluster (due to concurrently GetPartition/GetActivator)
And MemberListActor is no longer needed.


src/Proto.Cluster/MemberList.cs
```diff
@@ -4,109 +4,149 @@
...
-    }
-
-    internal class MembersByKindRequest
-    {
-        public MembersByKindRequest(string kind, bool onlyAlive)
+        internal static void UpdateClusterTopology(ClusterTopologyEvent msg)
         {
-            Kind = kind ?? throw new ArgumentNullException(nameof(kind));
-            OnlyAlive = onlyAlive;
+            lock (_lock)
```
rogeralsing: Replace with Monitor.TryEnter to allow timeouts IMO.
even if we just loop over that, at least we could send some notification that things are taking a long time

rogeralsing:
I love it, more red lines than green :) and nice perf gain too

PotterDai:
@roger yep. It just got fixed. I also fixed some naming conventions I didn’t pay a attention before. = )

PR #346, opened 2017-10-31
Reset timer when receiving timeout callback.

PotterDai:
During my stress test, ReceiveTimeoutCallback can be called twice in some cases due to task scheduling problem. This can be fixed by resetting timer when receiving the callback.


PotterDai:
@rogeralsing  Hey Roger, since this is a small fix, I'll just go ahead and merge it myself. :)

PR #349, opened 2017-11-07
Small updates for Future and Process

PotterDai:
- Use Stop.Instance instead of new Stop()
- In Future class, pid.Stop() is unnecessary, can directly call Stop(pid) instead, avoiding extra ProcessRegistry.Get(pid).


PotterDai:
@rogeralsing This is again a small fix. I'll go ahead and merge it.

PR #351, opened 2017-11-10
faster remote

rogeralsing:


rogeralsing:
This change resolves the target PID's in the RemoteReader so that each target is only resolved once per message batch, previously, it was resolved per message

cpx86:
LGTM!

PR #354, opened 2017-11-12
added  ReenterAfter(Task target, Func<Task, Task> action) and test fo…

halotron:
I could not figure out why there was no ReenterAfter(Task target, Func<Task, Task> action), so I added it.

### DIFF
--- a/src/Proto.Actor/IContext.cs
+++ b/src/Proto.Actor/IContext.cs
@@ -120,13 +120,6 @@ namespace Proto
         /// </summary>
         /// <param name="target">the Task to await</param>
         /// <param name="action">the continuation to call once the task is completed</param>
-        void ReenterAfter(Task target, Action<Task> action);
-        /// <summary>
-        ///     Awaits the given target task and once completed, the given action is then completed within the actors concurrency constraint.
-        ///     The concept is called Reentrancy, where an actor can continue to process messages while also awaiting that some asynchronous operation completes.
-        /// </summary>
-        /// <param name="target">the Task to await</param>
-        /// <param name="action">the continuation to call once the task is completed</param>
         void ReenterAfter(Task target, Action action);
 
 
--- a/src/Proto.Actor/IContext.cs
+++ b/src/Proto.Actor/IContext.cs
@@ -114,14 +114,21 @@ namespace Proto
         /// <param name="action">the continuation to call once the task is completed</param>
         /// <typeparam name="T">The generic type of the task</typeparam>
         void ReenterAfter<T>(Task<T> target, Func<Task<T>, Task> action);
-
         /// <summary>
         ///     Awaits the given target task and once completed, the given action is then completed within the actors concurrency constraint.
         ///     The concept is called Reentrancy, where an actor can continue to process messages while also awaiting that some asynchronous operation completes.
         /// </summary>
         /// <param name="target">the Task to await</param>
         /// <param name="action">the continuation to call once the task is completed</param>
-        /// <typeparam name="T">The generic type of the task</typeparam>
-        void ReenterAfter(Task target, Func<Task, Task> action);
+        void ReenterAfter(Task target, Action<Task> action);
+        /// <summary>
+        ///     Awaits the given target task and once completed, the given action is then completed within the actors concurrency constraint.
+        ///     The concept is called Reentrancy, where an actor can continue to process messages while also awaiting that some asynchronous operation completes.
+        /// </summary>
+        /// <param name="target">the Task to await</param>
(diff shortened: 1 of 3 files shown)


src/Proto.Actor/IContext.cs
```diff
@@ -113,6 +113,15 @@ public interface IContext : ISenderContext
         /// <param name="target">the Task to await</param>
         /// <param name="action">the continuation to call once the task is completed</param>
         /// <typeparam name="T">The generic type of the task</typeparam>
-        void ReenterAfter<T>(Task<T> target, Func<Task<T>,Task> action);
+        void ReenterAfter<T>(Task<T> target, Func<Task<T>, Task> action);
```
raskolnikoov: Instead of having to return a task maybe it could look like this?

`void ReenterAfter<T>(Task<T> target, Action<Task<T>> action);`

src/Proto.Actor/IContext.cs
```diff
@@ -113,6 +113,15 @@ public interface IContext : ISenderContext
...
+
+        /// <summary>
+        ///     Awaits the given target task and once completed, the given action is then completed within the actors concurrency constraint.
+        ///     The concept is called Reentrancy, where an actor can continue to process messages while also awaiting that some asynchronous operation completes.
+        /// </summary>
+        /// <param name="target">the Task to await</param>
+        /// <param name="action">the continuation to call once the task is completed</param>
+        /// <typeparam name="T">The generic type of the task</typeparam>
+        void ReenterAfter(Task target, Func<Task, Task> action);
```
raskolnikoov: And here we do:

`void ReenterAfter(Task target, Action<Task> action);`

src/Proto.Actor/LocalContext.cs
```diff
@@ -215,6 +215,14 @@ public void ReenterAfter<T>(Task<T> target, Func<Task<T>, Task> action)
             target.ContinueWith(t => { Self.SendSystemMessage(cont); });
         }
 
+        public void ReenterAfter(Task target, Func<Task, Task> action)
+        {
+            var msg = _message;
+            var cont = new Continuation(() => action(target), msg);
```
raskolnikoov: Create the continuation on both methods like this:

```
var cont = new Continuation(() =>
{
    action(target);
    return Task.FromResult(0);
}, msg);
```

tests/Proto.Actor.Tests/LocalContextTests.cs
```diff
@@ -29,6 +32,111 @@ public void Given_Context_ctor_should_set_some_fields()
...
+                if (ctx.Message as string == "hello1")
+                {
+                    var t = Task.Run(async () =>
+                    {
+                        await Task.Yield();
+                        await Task.Yield();
+                        queue.Enqueue("bar");
+                        await Task.Yield();
+                    });
+                    ctx.ReenterAfter(t, task =>
```
raskolnikoov: The method could then be used like this:

```
ctx.ReenterAfter(t, task =>
{
    queue.Enqueue("baz");
    ctx.Respond("hey1");
});
```

tests/Proto.Actor.Tests/LocalContextTests.cs
```diff
@@ -29,6 +32,111 @@ public void Given_Context_ctor_should_set_some_fields()
...
+                if (ctx.Message as string == "hello1")
+                {
+                    var t = Task.Run(async () =>
+                    {
+                        await Task.Yield();
+                        await Task.Yield();
+                        queue.Enqueue("bar");
+                        return "hey1";
+                    });
+                    ctx.ReenterAfter(t, task =>
```
raskolnikoov: And be used like this with T:

```
ctx.ReenterAfter(t, task =>
{
    queue.Enqueue("baz");
    ctx.Respond(task.Result);
});
```

rogeralsing:
@halotron, can you explain what the main purpose of the Task returning version is?
ReenterAfter is a void method, what is the returned Task used for in this case?

Or is it to enable async await in the reentrant block?

halotron:
@rogeralsing it just seemed logical to have both Task and Task<T> in the API. No, no need to return a Task. The Action<Task> and Action<Task<T>> is more what I needed, as @raskolnikoov said. I tried to do them but got some weird locking issue or something.

raskolnikoov:
@halotron hi, are you finished with this pr?

halotron:
I did an Action T and Action variant. Made the test a bit more robust.

halotron:
Ok, now I am done. It is just not needed to have Action Task. Just Action.

rogeralsing:
Looks great!

PR #355, opened 2017-11-13
EndpointWatcher dictionary should use remove instead of set to null.

PotterDai:
EndpointWatcher dictionary should use remove instead of set to null.


PotterDai:
@rogeralsing This is a small bug fix. So I will again go ahead and merge it.

PR #357, opened 2017-11-17
Fix wrong await code.

PotterDai:
This await code is wrong... fixed it in this pr.


PotterDai:
Hey @rogeralsing , I will merge this since it's a small fix. = )

PR #360, opened 2017-11-19
wirefix

rogeralsing:


rogeralsing:
Fixes the issue with using the Wire serializer and Proto.Cluster

src/Proto.Actor.Wire/WireSerializer.cs
```diff
@@ -1,27 +1,88 @@
...
+using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.IO;
 using Google.Protobuf;
 using Proto.Remote;
 using Wire;
+using Wire.SerializerFactories;
+using Wire.ValueSerializers;
 
 namespace Proto.Serialization
```
raskolnikoov: Update namespace to Proto.Serialization.Wire

PR #361, opened 2017-11-20
Persistence index fix

### LINKED ISSUE #340: Persistence Index increment before saving the event
I started to use Persistence with EventSourcing and encountered unexpected behavior that Index property is incremented before saving the event https://github.com/AsynkronIT/protoactor-dotnet/blob/dev/src/Proto.Persistence/Persistence.cs#L117. Seem like a bug to me as we never persist the event with index 0. Should it be incremented after the fact maybe? Should I sent a PR to fix this?

raskolnikoov:
Fixes #340

### DIFF
--- a/examples/Persistence/Persistence/Program.cs
+++ b/examples/Persistence/Persistence/Program.cs
@@ -29,8 +29,6 @@ class Program
         Console.ReadLine();
     }
 
-    public class RequestSnapshot { }
-
     class MyPersistenceActor : IActor
     {
         private PID _loopActor;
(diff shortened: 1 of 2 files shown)


PotterDai:
There seems to be an error in AppVoyer after this pr, which prevents it uploading the latest version to Myget.
https://ci.appveyor.com/project/protoactor-ci/protoactor-dotnet/branch/dev

raskolnikoov:
@PotterDai the failing test has been fixed now and dev branch is not passing...

PotterDai:
@raskolnikoov Thank you for the fix! =)

PR #363, opened 2017-12-03
Header can be sent between the nodes

PotterDai:
This pr allows header to be sent between different nodes.

### DIFF
--- a/examples/Middleware/Program.cs
+++ b/examples/Middleware/Program.cs
@@ -48,7 +48,6 @@ class Program
 
         var root = new ActorClient(headers, next => async (c, target, envelope) =>
                                    {
-                                       envelope.NewHeaderIfDefault();
                                        envelope.SetHeader("TraceID", c.Headers.GetOrDefault("TraceID"));
                                        envelope.SetHeader("SpanID", c.Headers.GetOrDefault("SpanID"));
                                        envelope.SetHeader("ParentSpanID", c.Headers.GetOrDefault("ParentSpanID"));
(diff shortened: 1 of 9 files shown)


PotterDai:
@rogeralsing This pr will reduce about 15%-20% of remote performance. This is mainly due to https://github.com/google/protobuf/blob/31c54d1289f2ae44158eff631526762366ba1707/csharp/src/Google.Protobuf/Collections/MapField.cs#L77
Each RemoteDeliver message will need allocate a new Dictionary and LinkedList for header even if it is null.

This perf drawback does not exist in Go version.

Considering this is necessary to make header/middleware actually working, I think we should just let it be? = )

cpx86:
@PotterDai Would it make a difference if we make the `message_header` optional in the proto file?

rogeralsing:
What if we make an intermediate message holding the map, and then have a reference to that message from the envelope?

rogeralsing:
```proto
message MessageEnvelope {
  int32 type_id = 1;
  bytes message_data = 2;
  int32 target = 3;
  actor.PID sender = 4;
  int32 serializer_id = 5;
  MessageHeader header = 6;
}

message MessageHeader {
     map<string, string> message_header = 1;
}
```

something like that.
(also, never reassign the proto field tags, that breaks all version cmpat.)

PotterDai:
@cpx86 @rogeralsing  I ended up with Roger's solution. = )
@rogeralsing  It just worked! Now perf has no drawback. I will fix Go version as well.

rogeralsing:
Looks good to me!

PotterDai:
@rogeralsing  Thx!
Just another question, shall we add ReceiveMiddleware property to Future? In that case, context.RequestAsync can pass its ReceiveMiddleware to Future, and when Future receive response, it can go through that middleware as well.

rogeralsing:
@PotterDai sounds good, that might ever help diagnose some issues if we can see that there is a future process involved in the message flow.

PR #364, opened 2017-12-10
 Avoid consensus problem when multiple nodes join/leave cluster in short amount of time.

PotterDai:
Same as https://github.com/AsynkronIT/protoactor-go/pull/207


rogeralsing:
Looks good

PR #365, opened 2017-12-12
ReceiveMiddleware in Future

PotterDai:
@rogeralsing 
Hey Roger, this PR is to allow Future received message be handled in relative middleware as well.
However the FutureContext and FutureActor seems a bit bulky, I didn't come up any nice way to handle that. Good thing is that it should have no perf impact = )

What's your opinion?

### DIFF
--- a/src/Proto.Actor/Futures.cs
+++ b/src/Proto.Actor/Futures.cs
@@ -5,11 +5,56 @@
 // -----------------------------------------------------------------------
 
 using System;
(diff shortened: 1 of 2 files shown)


rogeralsing:
I'll have a look.
I just realized that there might be implications of doing this.
e.g. if you have a middleware that is very specific for your actor, e.g. holds some state or does any other per instance related things.

PR #367, opened 2017-12-23
working docker compose example

mastoj:
To run it the whole project must be built first. After that you can run the `build.sh` script in the root of the `DockerComposeClusterHelloWorld` example folder.

In the latest commit I've also updated the original `ClusterHelloWorld` example to be parameterized so we can use the example in different scenarios, like this one.


rogeralsing:
LGTM cc @AsynkronIT/developers

Issue #369, opened 2017-12-27
Remote.SpawnNamedAsync fail after 10 ProcessNameAlreadyExist


jrouaix:
I have something really weird here :
![devenv_2017-12-27_15-27-27](https://user-images.githubusercontent.com/2705498/34384664-d9e8a992-eb1d-11e7-9e04-8f272745cfd5.png)

jrouaix:
After 10 Remote.SpawnNamedAsync() with the same name, all next calls will timeout.

PotterDai:
I think this is because Activator actor is using default supervisor strategy, which will kill itself after 10 failures. This can be fixed by using a custom strategy. I will try make a PR for this issue when I got some time. = )

PotterDai:
@rogeralsing Hey Roger, so this problem is a bit tricky. Default supervisor strategy has a failure cap of 10, if we want to prevent Activator from stop, I only came up 2 ways:
1 - Change default supervisor strategy (Easiest way to implement)
2 - Create a supervisor for Activator. (Ideally this supervisor should supervise all system actors, but will require a bit amount of code changes and increase complexity)
What do you think?

rogeralsing:
I've been thinking about the root actor supervision for a long time.
The idea of a parent supervising the children comes from Akka, in Erlang this is not an explicit case.
Erlang allows any actor to supervise another actor (afaik, not a big Erlang user)
In ProtoActor supervision has been somewhat extracted, we have an interface for it.

Maybe the right solution is to be able to pass such supervision instance to the actor to be created.

props.withSelfSupervisior(...)

or..

props.withParent(parentPid)

The latter set the actors parent to the passed pid, this way, we could have a system supervisor actor that act as some form of foster parent to actors created with the above approach.

Thoughts?

jrouaix:
Ho, I didn't see the ProcessNameExistException was rethrown !

I'd be glad to inject any kind of supervision implementation but for this remote activation problem, could the activator just send the response and swallow ProcessNameExistException ?

It seems to me it a normal thing to have multiple call for the same actor name (let's say I have 12 distinct client instances on this one ?). Perhaps also it's a questionable design to throw an exception on an expected situation ?

rogeralsing:
>activator just send the response and swallow ProcessNameExistException ?

It should, for clustering we even have semantics for that so you get status codes back.
This should be the behavior of the remote activator too.
[...]

PR #371, opened 2017-12-28
Implement guardian

PotterDai:
As is discussed here: https://github.com/AsynkronIT/protoactor-dotnet/issues/369

### DIFF
--- a/src/Proto.Cluster/PidCache.cs
+++ b/src/Proto.Cluster/PidCache.cs
@@ -23,7 +23,7 @@ namespace Proto.Cluster
 
         internal static void Setup()
         {
-            var props = Actor.FromProducer(() => new PidCacheWatcher()).WithGuardianSupervisorStrategy(Supervision.KeepAliveStrategy);
+            var props = Actor.FromProducer(() => new PidCacheWatcher()).WithGuardianSupervisorStrategy(Supervision.AlwaysRestartStrategy);
             watcher = Actor.SpawnNamed(props, "PidCacheWatcher");
             clusterTopologyEvnSub = Actor.EventStream.Subscribe<MemberStatusEvent>(OnMemberStatusEvent);
         }
--- a/src/Proto.Cluster/PidCache.cs
+++ b/src/Proto.Cluster/PidCache.cs
@@ -23,7 +23,7 @@ namespace Proto.Cluster
 
         internal static void Setup()
         {
-            var props = Actor.FromProducer(() => new PidCacheWatcher());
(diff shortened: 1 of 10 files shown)


PotterDai:
@rogeralsing 
OK, so this is what come up with:
https://github.com/PotterDai/protoactor-dotnet/blob/868a98295b21a6ea30241de5250c6d988dd57ccb/src/Proto.Actor/Guardians.cs

src/Proto.Cluster/PidCache.cs
```diff
@@ -23,7 +23,7 @@ internal static class PidCache
 
         internal static void Setup()
         {
-            var props = Actor.FromProducer(() => new PidCacheWatcher());
+            var props = Actor.FromProducer(() => new PidCacheWatcher()).WithGuardianSupervisorStrategy(Supervision.KeepAliveStrategy);
```
rogeralsing: IMO, AlwaysRestartStrategy would be a more descriptive name, as "KeepAlive" could mean anything, e.g. ignore errors etc.

PotterDai: Yep, just got it updated.

rogeralsing:
Looks great.

One thought, this PR uses the props style for setting guardian: `WithGuardianSupervisorStrategy`

That means that props containing this which are spawned in a child context.
e.g. `context.Spawn(props)` would have some strange semantics, or throw.
(Is the parent actor the supervisor, or is the guardian in this case?)

What if we have a copy of the Spawn methods on a Guardian class.

`var pid = RestartingGuardian.SpawnNamed("someActor",props);`

Not saying the above suggestion is the way to go, but we need to have a clear path for what happens to props used in a child context

PotterDai:
Right, however considering there might be some cases where user want to use Guardian as well, RestartingGuardian.SpawnNamed will prohibit it from happening, for example when spawning in remote Activator.

I think we should just throw an exception to say that Guardian is not supported in context.Spawn ?
Or it might also be OK to just leave it as is (Maybe just print out an warning message), since GuardianStrategy will only be checked in Actor.Spawn.

rogeralsing:
>I think we should just throw an exception to say that Guardian is not supported in context.Spawn ?

I'd say we should go with that. seems least surprising

PotterDai:
Yep, just got it updated. Please check again = )

rogeralsing:
Looks good!

PotterDai:
I'll find a time to get this in Go as well = )

PR #380, opened 2018-01-10
implement stopping when max number of restarts exceeds allowed number…

### LINKED ISSUE #379: OneForOneStrategy don't reset failurecount correctly
In OneForOneStrategy when setting maxNrOfRetries to 3 and withinTimeSpan to 100 ms, I expected that it should restart if I send 3 messages and then wait for 200 ms and send another 3 messages (all failing). 
But instead it fails and shuts down the actor. 
The problem seems to be in how to count how many errors occur within a given time, it only checks time against the last error and not how many errors happens in the "withinTime". 
Did I missunderstand how it's suppsed to work?

`

	 var childMailboxStats = new TestMailboxStatistics(msg => msg is Stopped);

	var strategy = new OneForOneStrategy((pid, reason) => SupervisorDirective.Restart, 3, TimeSpan.FromMilliseconds(10));

	var childProps = Actor.FromProducer(() => new ChildActor())
		.WithMailbox(() => UnboundedMailbox.Create(childMailboxStats));

	var parentProps = Actor.FromProducer(() => new ParentActor(childProps))
		.WithChildSupervisorStrategy(strategy);

	var parent = Actor.Spawn(parentProps);
	parent.Tell("hello");
	parent.Tell("hello");
	parent.Tell("hello");
	Thread.Sleep(100);
	parent.Tell("hello");
	parent.Tell("hello");
	parent.Tell("hello");

	childMailboxStats.Reset.Wait(2000);

	Assert.DoesNotContain(Stop.Instance, childMailboxStats.Posted);
	Assert.DoesNotContain(Stop.Instance, childMailboxStats.Received);

`

I think that the RestartStatistics needs to be a bit more advanced, and storing not only how many failures has happend, and when the last one happend, but when all failures happens and remove the ones that are older then the time limit. This however is not needed for other strategys like the escalating.

tomliversidge:
… in a period of time

Uses the _withinTimeSpan value properly to fix https://github.com/AsynkronIT/protoactor-dotnet/issues/379
Reversed the control flow in the Restart section as it kept confusing me
Adds tests 

NOTE: ExponentialBackoffStrategy looks wrong to me, we are lacking tests around this as the current two still pass, could do with some input into what this strategies behaviour is supposed to be.

### DIFF
--- a/src/Proto.Actor/RestartStatistics.cs
+++ b/src/Proto.Actor/RestartStatistics.cs
@@ -5,30 +5,32 @@
 // -----------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace Proto
 {
     public class RestartStatistics
     {
+        private readonly List<DateTime> _failureTimes = new List<DateTime>();
+        
+        public int FailureCount => _failureTimes.Count;
+
         public RestartStatistics(int failureCount, DateTime? lastFailuretime)
         {
-            FailureCount = failureCount;
-            LastFailureTime = lastFailuretime;
+            for (int i = 0; i < failureCount; i++)
+            {
+                _failureTimes.Add(lastFailuretime ?? DateTime.Now);
+            }
         }
 
-        public int FailureCount { get; private set; }
-        public DateTime? LastFailureTime { get; private set; }
+        public void Fail() => _failureTimes.Add(DateTime.Now);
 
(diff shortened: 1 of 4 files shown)


raskolnikoov (approved):

tomliversidge:
I've fixed exponential backoff too.- should now reset if no errors have occurred within the window.

PR #384, opened 2018-01-25
address issue (TryDequeue return True but get null result)

lust4life:
https://github.com/alexandrnikitin/MPMCQueue.NET/issues/2


alexandrnikitin:
I just wanted to note that it's important to merge it. The queue could lose messages during high contention and small buffer length. [A test for it](https://github.com/alexandrnikitin/MPMCQueue.NET/blob/3750330e25d11617c8a4e450cba80435eac70532/tests/MPMCQueue.NET.Tests/FieldReports/Issue2_TryDequeueTrueButGetNull.cs)
Thanks for fixing it @lust4life 

BTW, I can wrap the queue into a NuGet package if there's interest.

potterdai:
@rogeralsing I think I'll go and merge this PR = )
@alexandrnikitin @lust4life  Thank you very much! That will be great if the queue can be a NuGet package.

PR #385, opened 2018-01-25
No need to count messages

Scooletz:
This PR changes the way the mailbox is determined whether it has messages or not. `IsEmpty` is a faster alternative than counting messages.


potterdai:
@rogeralsing I'll go ahead and merge this one as well.
@Scooletz  Thank you very much for the PR! This will definitely boost our perf = )

Issue #141, opened 2017-04-07
How to test EventStream (in the context of DeadLetter testing)


tomliversidge:
I've been looking at how we can test that messages are sent to the DeadLetterProcess and subsequently published on the EventStream and the current design makes it difficult to test. Both DeadLetterProcess and EventStream have Instance static properties that are used to access them. The result of this is that you do not (by default) get isolation between tests (i.e. when Test1 and Test2 run in parallel and both use the static Instance of EventStream they will interfere with each other's assertions)

I can currently think of two options to deal with this:

1. Run any tests that use static Instances in a custom xunit test collection (https://xunit.github.io/docs/running-tests-in-parallel.html). This would make them run serially, so there would be no interference. Each test would be responsible for clearing up any state that is setup (i.e. remove subscriptions in the EventStream etc.)
2. Redesign - allow passing in of an EventStream in the Props, same way you can pass a Mailbox or Dispatcher. This would then be passed into the DeadLetterProcess ctor inside of the ProcessRegistry.Get method. We'd need a way to know about the EventStream - looks like it would need storing on the PID object. 

No. 1 is basically an anti-pattern - shared state between tests etc.. - but it would be the easiest to implement. There are quite a few downsides to this approach though:
1. You have to remember to use the xunit test collection on all tests that might publish anything to the EventStream. This becomes a problem when you don't realize you're using the EventStream (i.e. it's called from within something else deeper down the chain) - the worse part is tests will pass in isolation, but _sometimes_ fail when running all together.
2. You have to remember to clean up any left over state
3. Tests using the collection run serially, thus the overall test suite becomes slower

The advantage is that no src code has to change - there's no _[test induced design damage](http://david.heinemeierhansson.com/2014/test-induced-design-damage.html)_

No. 2 - i don't know the code base well enough to know the impact of such a redesign. It looks do-able - just pass some references down, but I think someone with more experience of the code base would have to ok this.

Thoughts? Might also be related to https://github.com/AsynkronIT/protoactor-dotnet/issues/90

Here's an example of testing using no1:

https://gist.github.com/tomliversidge/0bb0d84931e4de369925d7630bb17a96

and no2:

https://gist.github.com/tomliversidge/63082b3e8c6e2496c024feee3adac369

rogeralsing:
There is now the `EventProbe` that can be used exactly for this

PR #387, opened 2018-02-08
fix: utilize EventStream.Instance for DI example

masteryee:


src/Proto.Actor.Extensions/ServiceCollectionExtensions.cs
```diff
@@ -9,7 +8,6 @@ public static class ServiceCollectionExtensions
         public static void AddProtoActor(this IServiceCollection services, Action<ActorPropsRegistry> registerAction = null)
         {
             services.AddSingleton<IActorFactory, ActorFactory>();
-            services.TryAdd(ServiceDescriptor.Singleton(typeof(EventStream<>), typeof(EventStream<>)));
```
masteryee, line 12: The DI framework can't resolve EventStream<> as a dependency because there is no public constructor, so removing this line since it has no impact.

adamhathcock:
I wish EventStream was injected and not static @rogeralsing :)

rogeralsing:
Injected into what?

adamhathcock:
Well, I just mean that EventStream objects could be created and used.  The global one isn't required.

I'm probably just misusing/misunderstanding the purpose of EventStream which is why I wrote my DI example that way.

tomliversidge:
Relates to this too https://github.com/AsynkronIT/protoactor-dotnet/issues/141

PR #392, opened 2018-02-27
Add ContextState.Stopped.

potterdai:
@rogeralsing  This fixed bug mentioned on gitter, that messages are still being handled after actor is stopped, if those messages are already in mailbox.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,7 +12,7 @@ nuget:
   disable_publish_on_pr: true
 
 before_build:
-- nuget install Cake -ExcludeVersion
+- nuget install Cake -ExcludeVersion -Version 0.25.0
 
 build_script:
 - .\Cake\Cake.exe --target=PrintEnvironment
--- a/examples/ActorStopTest/ActorStopTest.csproj
+++ b/examples/ActorStopTest/ActorStopTest.csproj
@@ -1,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+  </PropertyGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Proto.Actor\Proto.Actor.csproj" />
+    <ProjectReference Include="..\..\src\Proto.Mailbox\Proto.Mailbox.csproj" />
+  </ItemGroup>
+</Project>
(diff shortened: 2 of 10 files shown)


potterdai:
@rogeralsing  We got a problem here.
If we reject any user messages when LocalContext status is Stopped(https://github.com/PotterDai/protoactor-dotnet/blob/bcc4596c9585cd7b722b85bb3bd29144e782355b/src/Proto.Actor/LocalContext.cs#L337), we will face a situation like this:
```
endpoint.Watcher.Tell(EndpointTerminatedEvent)
endpoint.Watcher.Stop()
```
0. There is certain possibility EndpointTerminatedEvent and Stop queuing up in mailbox due to threading.
1. Mailbox will process Stop message first since it is a SystemMessage
2. LocalContext will become ContextState.Stopped
3. EndpointTerminatedEvent will never be processed since Actor has stopped already.

Solution could be 
1. we should tell our user never call pid.Stop() right after pid.Tell(...), but that's a bit counter-intuitive.
2. Or we just don’t prioritize Stop message in mailbox, treat Stop as a special message and process it at last.

rogeralsing:
Imo we go for 1.

```
endpoint.Watcher.Tell(EndpointTerminatedEvent)  //make the procesing of this event stop internally
endpoint.Watcher.Stop() //remove this
```

That would work, right?

potterdai:
@rogeralsing 
Yep, that would work. But it might break other things as well, for example:
https://github.com/AsynkronIT/protoactor-go/blob/dev/actor/actor_example_test.go#L21

Do you think there are many real world cases that user will call Stop() right after call Tell(...)?

rogeralsing:
There is also the option to send a PoisonPill, which is a user message, but does a stop.

potterdai:
@rogeralsing Yep, I just got that fixed in this PR.
However I think it is still better to treat Stop as a special message, or let's just remove Stop message type,  and do that using
```
pid.Stop() => Mailbox.Stop() => Mailbox.RunAsync(){ if (stopped) ... } => Context.InvokeStopAsync()
```
In this case it can completely eliminate this Tell() then Stop() conflicts.

rogeralsing:
Problem with not being able to stopp directly is if you have an overloaded actor, mailbox is huge, and the system wants to kill it. then the stop would be stuck at the end of the queue.

potterdai:
@rogeralsing  In this case it might be necessary to introduce Kill() and Stop(), but this will lead to a lot of structure changes. = )
Let's merge this PR first, and I'll see if I got time to explore this direction.

potterdai:
@rogeralsing  I'll do the merge.
I just realized that PoisonPill can be used if user want to have all messages in mailbox processed before actor stop, so it's not necessary to re-implement a new system to achieve that.

Issue #407, opened 2018-06-12
Proto.Cluster.SingleRemoteInstance


jrouaix:
I did some pretty nasty things on our stack to have some abstraction on Cluster/Remote/Local in order to start developpements in the same way/api in local without Consul, than on the expected target when going in production ... and it's a mess. 

As I discussed that with my boss @Johnny-Bee said : 
> "We should always be in cluster mode !"

And then now bazzinga! Let's create a ClusterProvider that encapsulate Remote calls, then we are good to always code Cluster thing, and decide at runtime to run the single remote or the Consul one.

Anyone interested ? sounds good ? If good I plan to code that quick to have it in vNext.

jrouaix:
@rogeralsing @cpx86 how you feel about this ?

rogeralsing:
I'm not quite sure I understand what is being suggested here.
There are quite different semantics between Local, Remote and Cluster.
Remote is similar to Local, you still control life cycle of actors, but with the potential risk of a node dying.

Cluster, uses the Virtual Actor model and identifies actors by name instead of PID (well that is used under the cover, but the surface API is for names)

Is your suggestion to build a cluster provider that allows you to start a single node, that thinks it is a cluster?

In that case, that should be completely fine. 👍

cpx86:
> a cluster provider that allows you to start a single node, that thinks it is a cluster

This sounds interesting for sure. However in this scenario the term cluster is a bit misleading. You're rather running non-remote virtual actors (as opposed to remote virtual actors).

There have been ideas before about being able to use virtual actors even in local scenarios. Perhaps it could be possible to factor out the virtual actors into a VirtualActor package that's usable locally, and let Cluster be the glue between VirtualActor and Remote? Not sure if that solves @jrouaix's problem though.

rogeralsing:
@cpx86 theoretically, it should be completely possible to create a custom ClusterProvider that is a single node.

But cluster starts the remote subsystem, so I think you might be onto something about extracting the virtual actor part. 
It is somewhat entangled right now. Cluster is bound to Remote via code and messages.

jrouaix:
> Is your suggestion to build a cluster provider that allows you to start a single node, that thinks it is a cluster?

Exactly !

@cpx86 for sure a virtual actor concept independent from location/communication concerns would be a plus I think

I'll publish a PR soon, comments are welcome

rogeralsing:
Closing as this has been merged

PR #408, opened 2018-06-12
WIP : Single remote instance cluster- do not merge

jrouaix:
as discussed here : https://github.com/AsynkronIT/protoactor-dotnet/issues/407


src/ClusterProviders/Proto.Cluster.SingleRemoteInstance/SingleRemoteInstanceProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+            _koStatus = serializer.FromValueBytes(Encoding.UTF8.GetBytes("Ko!"));
+
+            _isServer = host == _serverHost && port == _serverPort;
+
+            return Actor.Done;
+        }
+
+        public void MonitorMemberStatusChanges()
+        {
+            var t = new Thread(_ =>
```
rogeralsing, line 59: 👍 I was going to say it could be done using Async Await to not hog a thread.
But this is better as if the threadpool is under stress, the notifications will still run at the correct intervals.

jrouaix:
@rogeralsing you think it's mergable material to dev ?

Issue #409, opened 2018-06-14
Context Decorators


rogeralsing:
In an effort to intercept calls to methods of the different contexts.
We need a decent way to decorate the context instances.

What I mean here is:

`IContext.RequestAsync` →  `MyDecoratorContext.RequestAsync` →  `ActorContext.RequestAsync`

And something similar for RootContext.

First of all, RootContext is currently just created with `new`, for this to work properly, we need some factory for root contexts.

e.g. RootContext.Create() // returns an IRootContext with potential decorators wrapped

in the case of actors, it's easier. we can simply have a `WithContextDecorator` on `Props`
Then when the ActorContext invokes ReceiveAsync on the actor, we just pass the wrapper context.

Thoughts ?

rogeralsing:
Relates to #406

jrouaix:
I tried to used something like that but ... Proto is not ready ... and it's not 4.5.2 compatible.

```csharp
  class OpenTracingDecorator<T> : DispatchProxy
    {
        private OpenTracingDecorator() { }

        T _context;

        public static T Create<T>(T context, SpanSetup sendSpanSetup, ITracer tracer)
        {
            var proxy = Create<T, OpenTracingDecorator<T>>();
            var t = proxy as OpenTracingDecorator<T>;
            t._context = context;
            return proxy;
        }

        protected override object Invoke(MethodInfo targetMethod, object[] args)
        {
            if (targetMethod.Name == nameof(ISenderContext.RequestAsync))
                throw new InvalidCastException("Just to know I passed here");

            return targetMethod.Invoke(_context, args);
        }
    }
```

As our need is not that general, we could implement some "ContextBuilder" that will be able to add middleware/aspect for each methods :
```csharp
props.WithContextDecorator(ctxBuilder => {
  ctxBuilder.WithRequestAsync(async (next, target, message) => {
    // do something before
    var result =   await next(target, message);
    // do something after
  };
});
```

rogeralsing:
I'm thinking if we just implement a base class for decorators, that manually forwards each call. no magic, just code.

Then the implementor of a decorator could just override the method they want to intercept

rogeralsing:
I've pushed this: https://github.com/AsynkronIT/protoactor-dotnet/blob/dev/src/Proto.Actor/ActorContextDecorator.cs

Inherit and override the request methods. that should at least technically work.
[...]

PR #412, opened 2018-06-15
Opentracing improvements

jrouaix:
@rogeralsing. Some more OpenTracing delivery. Could be enough to start using it (that i'll try on our projects next week using the myget prerelease packages)

Some frustration :
- Cannot decorate Context.Receive
  * so I can't fill the proto.actortype tag on the receive side
  * nor the proto.targetpid tag on the receive side (it should be easy, we are just on the target side :-|)

Tried some tricky changes : https://github.com/jrouaix/protoactor-dotnet/commit/28da2aa15f16cb1d1c16eab29ba997486001052e
No success so far on that part https://github.com/AsynkronIT/protoactor-dotnet/issues/409


rogeralsing:
"Cannot decorate Context.Receive" 

How so?

`public virtual Task Receive(MessageEnvelope envelope)`

It's in there and should be possible to override?

jrouaix:
Yes it's possible to override. Then when you do it, it's not called.

rogeralsing:
Ok, Ill take a peek at that then. it should be invoked

rogeralsing:
See https://github.com/AsynkronIT/protoactor-dotnet/issues/413

jrouaix:
yep, thank you : https://github.com/AsynkronIT/protoactor-dotnet/pull/414

PR #417, opened 2018-06-21
Chainable context decorators

jrouaix:
Hello @rogeralsing. Needed this one on our project.

### DIFF
--- a/examples/ContextDecorators/Program.cs
+++ b/examples/ContextDecorators/Program.cs
@@ -21,16 +21,19 @@ namespace ContextDecorators
     
     public class LoggingDecorator : ActorContextDecorator
     {
-        public LoggingDecorator(IContext context) : base(context)
+        private readonly string _loggerName;
+
+        public LoggingDecorator(IContext context, string loggerName) : base(context)
         {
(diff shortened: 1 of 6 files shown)


examples/ContextDecorators/Program.cs
```diff
@@ -39,15 +42,19 @@ static void Main(string[] args)
...
-                }
+                    if (ctx.Message is string str)
+                    {
+                        Console.WriteLine("Inside Actor: " + str);
+                        ctx.Respond("Yo!");
+                    }
+
+                    return Actor.Done;
+                })
+                .WithContextDecorator(c => new LoggingDecorator(c, "logger1"))
```
rogeralsing, line 54: Should we align this with WithReceiveMiddleware and make it variadic in terms of params?

rogeralsing: NVM, I see that it already is

jrouaix: Well, I'm not sure it will do more.
By the way what is the goal of this code style (variadic funcs) in current middlewares ?

PR #419, opened 2018-07-05
Fixed issue with namespaces with deeper nesting

mattdurham:
This is a fix to resolve namespaces that are several layers deep.


rogeralsing:
Thanks!

Issue #403, opened 2018-06-06
Cluster fail to stabilize after node kill


jrouaix:
Ran some tests to ensure the problem reported on gitter was not our stack fault.
And I seems I can reproduce it.

![image](https://user-images.githubusercontent.com/2705498/41046555-eb4d41a0-69aa-11e8-98aa-82c1c21189dc.png)

To reproduce : 
Run a cluster with some nodes, then kill any node. 
The cluster won't ever recover properly and a sh!t load of errors is passing in the VS debug window.

jrouaix:
You can find details about the code I ran today in this branch : https://github.com/jrouaix/protoactor-dotnet/tree/cluster-stability-test

⚠️ I updated consul (it changed nothing), it took me to remove my /tmp/consul folder so the single node consul agree to start properly

rogeralsing:
Will look into this. 👍

jrouaix:
Any update on this one ? Do you need me to dig something to help ?

jrouaix:
@rogeralsing ? any update on this one ? Is it related to the https://github.com/AsynkronIT/protoactor-dotnet/issues/411 refactoring you wanna do ?

jrouaix:
My bad it was this branch : https://github.com/jrouaix/protoactor-dotnet/tree/cluster-stability-tests . with an 's' at the end. 

Still, kill one node and nothing work anymore.

alexeyzimarev:
I think I have some insights now about this issue. The read address of an endpoint is different from the advertised address. I kind of understand the reason behind it, but still not sure if it is needed.

The result is that at some point when a node shuts down and anther node enters the cluster, and these nodes are different, somehow they differ in advertised and real addresses.

Example from the log (using the cluster stability test):

```
dbug: Proto.Remote.EndpointManager[0]
      Forwarding message for 127.0.0.1:53844 through EndpontWriter { "Address": "127.0.0.1:53822", "Id": "EndpointSupervisor/$11613" }
```

I will look closer to the remote configuration, maybe I can figure out how this happens.

alexeyzimarev:
Ok, there were some issues, mostly with Consul provider. I spent a couple of days fixing it.

First, I refactored it to be done in actors instead of threads. Not only in uses less threads, but it is now more resilient. Then, I handled these cases:

- Service recovery after it was unregistered by Consul. The current version just crashes.
- Everything is asynchronous
[...]

Issue #223, opened 2017-05-15
Building / developing on mac


tomliversidge:
Currently the solution does not build on macOS and does not load in Visual Studio for Mac. This is because we are setting the target frameworks as net452 and netstandard1.5. When running the build.sh script, projects fail on building for net452 

We can fix this by adding in a hint for MSBuild to find the mono runtime to build against.

```
<PropertyGroup Condition="'$(TargetFramework)' == 'net452' AND '$(OS)' == 'Unix'">
<FrameworkPathOverride>/Library/Frameworks/Mono.framework/Versions/5.0.0/lib/mono/4.5.2-api/</FrameworkPathOverride>
  </PropertyGroup>
```

This is the location of mono if installed with Xamarin Studio. If installed through brew, it will be different. See https://github.com/dotnet/netcorecli-fsc/wiki/.NET-Core-SDK-rc4#using-net-framework-as-targets-framework-the-osxunix-build-fails for details. If we do this we should have a nice way to determine where mono is installed. NOTE Visual Studio for mac doesn't seem to support having multiple target frameworks. 

An alternative way to fix this is to remove net452 from the target frameworks. We can just target netstandard and it builds on mac, and can be opened in Visual Studio for mac. If we go down this route we would lose support for certain versions of full framework, and we need to decide if we want to target each project to the same version of netstandard or go for the minimum required for each project:

Proto.Actor netstandard1.2
Proto.Mailbox netstandard1.2
Proto.Router netstandard1.3
Proto.Remote netstandard1.5
etc...

If we lower things this way, someone with a full framework 4.5.1 app could use Proto.Actor and Proto.Mailbox for in-memory, single machine app only, but someone running 4.6.0 could use Proto.Router, as well as Martin and RavenDB persistence, as these can also both be lowered to netstandard1.3

In general though, would people find this micro-style versioning more annoying / confusing than just saying "Proto requires X" and stick to a single target? Being more micro versioned would mean we could support certain scenarios in more places, but could be confusing.

Outstanding issues to investigate:
* Loading projects in Visual Studio for mac when targeting multiple frameworks
* Dynamically determine mono runtime location on mac

adamhathcock:
I wish you there was a better flag to ignore certain frameworks with msbuild.

Cake scripting can do the desired thing and filter out full framework when non windows which is what I've done in the past.

tomliversidge:
Multiple target frameworks are not supported on Visual Studio for mac:

[...]

PR #421, opened 2018-08-03
Release 0.2.0

cpx86:
Not sure how to review this even - biggest PR I've seen in a while 😄


jrouaix:
Won't you wait until oppened bug are resolved ? 
https://github.com/AsynkronIT/protoactor-dotnet/labels/bug

cpx86:
@jrouaix Was #403 introduced since the last release (current master)? If yes then we should definitely fix it before merging; if no then I think it's fine to merge since we're not introducing the bug.

#223 I'm not sure if it's still an issue or not; we'd have to re-investigate it. Regardless though it's a build problem, so it's not relevant for the release IMO.

jrouaix:
@cpx86 #403 was discovered on a stable nuget, then I reproduced it also on a pretty "late updated" develop branch.

PR #433, opened 2018-09-30
hack the request bug

### LINKED ISSUE #432: Future<> BUG - need help
Hello @rogeralsing @cpx86 @PotterDai 
I'm turning mad on our production because of a bug repoduced on this branch : https://github.com/jrouaix/protoactor-dotnet/tree/hotfix/remote-local-nightmare

First that was a problem on local/remote calls mixing, then I narrowed it in Future<> usage in some Task situation : https://github.com/jrouaix/protoactor-dotnet/blob/f0f88456d983ad55fc913a0c429db6c84c0623dc/tests/Proto.Actor.Tests/FutureTests.cs#L118

Weirder this will break : 
```
await Context.RequestAsync<object>(pid, "hello", TimeSpan.FromSeconds(1));
Context.RequestAsync<object>(pid, "hello", TimeSpan.FromSeconds(1)).Wait();
```
but it will go fine in this order :
```
Context.RequestAsync<object>(pid, "hello", TimeSpan.FromSeconds(1)).Wait();
await Context.RequestAsync<object>(pid, "hello", TimeSpan.FromSeconds(1));

```
This one will break, but not if I had a breakpoint
```
Context.RequestAsync<object>(pid, "hello", TimeSpan.FromSeconds(1)).Wait();
await Context.RequestAsync<object>(pid, "hello", TimeSpan.FromSeconds(1));
Context.RequestAsync<object>(pid, "hello", TimeSpan.FromSeconds(1)).Wait();
```
This one is passing
```
// but if you add just one await version here, it won't
Context.RequestAsync<object>(pid, "hello", TimeSpan.FromSeconds(1)).Wait();
Context.RequestAsync<object>(pid, "hello", TimeSpan.FromSeconds(1)).Wait();
await Context.RequestAsync<object>(pid, "hello", TimeSpan.FromSeconds(1));
Context.RequestAsync<object>(pid, "hello", TimeSpan.FromSeconds(1)).Wait();
Context.RequestAsync<object>(pid, "hello", TimeSpan.FromSeconds(1)).Wait();
await Context.RequestAsync<object>(pid, "hello", TimeSpan.FromSeconds(1));
Context.RequestAsync<object>(pid, "hello", TimeSpan.FromSeconds(1)).Wait();
Context.RequestAsync<object>(pid, "hello", TimeSpan.FromSeconds(1)).Wait();
await Context.RequestAsync<object>(pid, "hello", TimeSpan.FromSeconds(1));
await Context.RequestAsync<object>(pid, "hello", TimeSpan.FromSeconds(1));
await Context.RequestAsync<object>(pid, "hello", TimeSpan.FromSeconds(1));
```

rogeralsing:
This is a hack to fix #432

Anyone that can come up with a better solution is welcome to do so.
All tests run and pass. behavior is now correct. even if the code is ugly.

### DIFF
--- a/src/Proto.Actor/Futures.cs
+++ b/src/Proto.Actor/Futures.cs
@@ -61,22 +61,21 @@ namespace Proto
                         
                         Stop(pid);
                     });
-                Task = _tcs.Task.ContinueWith(x =>
-                {
-                    if (x.IsFaulted)
-                    {
-                        throw new TimeoutException("Request didn't receive any Response within the expected time.");
-                    }
-     
-                    return x.Result;
-                });
+                Task = WrapTask(_tcs.Task);
             }
             else
             {
-                Task = _tcs.Task.ContinueWith(x => x.Result);
+                Task = WrapTask(_tcs.Task);
             }
             
             
+        }
+
+        private static async Task<T> WrapTask(Task<T> task)
+        {
+            await System.Threading.Tasks.Task.Yield();
+            var res = await task;
+            return res;
         }
 
         public PID Pid { get; }
@@ -93,7 +92,7 @@ namespace Proto
                 }
                 else
                 {
-                    _tcs.SetException(
+                    _tcs.TrySetException(
                         new InvalidOperationException(
                             $"Unexpected message. Was type {msg.GetType()} but expected {typeof(T)}"));
                 }
--- a/src/Proto.Actor/Futures.cs
+++ b/src/Proto.Actor/Futures.cs
@@ -31,6 +31,7 @@ namespace Proto
         private FutureProcess(CancellationTokenSource cts)
         {
             _tcs = new TaskCompletionSource<T>();
+
             _cts = cts;
 
             var name = ProcessRegistry.Instance.NextId();
@@ -41,6 +42,8 @@ namespace Proto
             }
 
             Pid = pid;
+
+
 
             if (cts != null)
             {
@@ -55,11 +58,25 @@ namespace Proto
 
                         _tcs.TrySetException(
                             new TimeoutException("Request didn't receive any Response within the expected time."));
+                        
                         Stop(pid);
                     });
+                Task = _tcs.Task.ContinueWith(x =>
+                {
+                    if (x.IsFaulted)
+                    {
+                        throw new TimeoutException("Request didn't receive any Response within the expected time.");
+                    }
+     
+                    return x.Result;
+                });
             }
-
-            Task = _tcs.Task;
+            else
+            {
+                Task = _tcs.Task.ContinueWith(x => x.Result);
+            }
+            
+            
         }
 
         public PID Pid { get; }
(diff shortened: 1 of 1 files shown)


jrouaix:
It's not that ugly ! 
Don't you commit the unit test to be sure it won't come back ?
Then if you could merge to dev tomorrow, i'll use the myget package and give you feedback from real case usage.

rogeralsing:
The test is already pushed to dev

jrouaix:
Current code is hiding an exception : 

Test proof :
https://github.com/jrouaix/protoactor-dotnet/blob/9a87d9d4999e7abf0136cc378f8ed1e39a4db733/tests/Proto.Actor.Tests/FutureTests.cs#L139

You can rethrow the correct exception here : https://github.com/jrouaix/protoactor-dotnet/blob/9a87d9d4999e7abf0136cc378f8ed1e39a4db733/src/Proto.Actor/Futures.cs#L72

Also on this line it might be wiser to "Try" ? :
https://github.com/jrouaix/protoactor-dotnet/blob/9a87d9d4999e7abf0136cc378f8ed1e39a4db733/src/Proto.Actor/Futures.cs#L94

rogeralsing:
This change breaks two tests where the expected exception was TimeoutException, which is now an AggregateException. not sure what is the kosher way to handle this

jrouaix:
You're right, it shouldn't be an Aggregate exception, yet it should be InvalidOperationException, not Timeout.

rogeralsing:
I've updated the PR.
Much cleaner impl, can you check if this works better?

rogeralsing:
See: https://github.com/AsynkronIT/protoactor-dotnet/pull/433/files#diff-59b30f82cf314c8b602e674276e92cdbR74

jrouaix:
Well the Yield() solution reproduce a timeout bug.
Instead a Delay(1) is ugly but works fine : https://github.com/jrouaix/protoactor-dotnet/blob/be70c8558a4e2396a94e194327dcebfc7474e957/src/Proto.Actor/Futures.cs#L74

dpavsrtrl:
I've a feeling these hacks are around the fact that futures process doesn't have a mailbox.

cpx86:
@dpavsrtrl Hmm, you mean if we used an actor or at least a mailbox we could avoid this situation?

dpavsrtrl:
@cpx86 I thought so, but see #436

alexeyzimarev:
@rogeralsing Feedback: after applying this fix, out test cluster that went down after being one hour under production load, stays alive for the night. Previously, we got nodes that were sending messages to remove actors in the cluster, stuck rather quickly on requesting actor address from the partition node. We have to senders. One uses request-response to ensure that all messages are persisted on the other side. It was getting stuck after 30-40 minutes. Another one was doing fire-and-forget and it was working longer but still was getting stuck on request-response with the partition node. Now everything f*cking works, that's awesome!

rogeralsing:
Great to hear! :-)

potterdai:
@rogeralsing 
After this PR, it seems ClusterStabilityTest no longer passes, digging out what's going on...

potterdai:
@rogeralsing 
It seems `await System.Threading.Tasks.Task.Yield();` in `WrapTask` hangs indefinitely.
Remove that line solved the issue, but that also removes the purpose of this PR right?

potterdai:
Further investigation indicates that `await Task.Yield()` only hangs when involving remote/GRPC, and when there are a lot of remote request(>1000 from i7 2.2Ghz) happening simultaneously.

PR #434, opened 2018-10-05
Return future on different thread (issue 432)

dpavsrtrl:


alexeyzimarev:
I believe it is not an issue anymore. The test mentioned in the issue is passing now and the initial hack is removed. It seems the solution was to set the task creation options in the `FutureProcess` constructor, like in one of the PRs in EventStore for the issue mentioned previously:

```
            _tcs = new TaskCompletionSource<T>(TaskCreationOptions.RunContinuationsAsynchronously);
```

PR #438, opened 2018-10-17
Add Saga project to examples

goblinfactory:
The Saga pattern sample project contains a rich example of a use case typical of developers considering actor frameworks.


cpx86:
Huh, weird it wasn't in the sln file. Thanks!

cpx86:
Seems there are some build errors though :(

goblinfactory:
@cpx86 Hi Christian, yes, just seen the build errors. Submitted the pull request on the way in to client on the train, so took a chance that the project had accidentally been left off the sln in a recent change. Obviously it's been left off for a while, at least long enough that because it's not in the sln it has effectively been left off `CICD`  and my commit correctly causes it to fail.

my change didn't break the build, the build was already broken. 
I will take a stab at fixing the project, will be a fun way to get stuck in :)
I have not looked to see if there are any other projects not included, that was one I stumbled across because I was specifically looking for a saga-unit of work example.

(he says, rolling up sleeve)

examples/Patterns/Saga/Program.cs
```diff
@@ -11,6 +11,8 @@ namespace Saga
 {
     internal class Program
     {
+        public static readonly RootContext Context = new RootContext();
```
goblinfactory: I have no idea if this is how context's should be managed. I will look through the docs to see if I can find out, however, given that this particular sample looks like a very high profile sample (because of the extra effort that appears to have gone into the accompanying readme documentation, I would really appreciate any extra eyes on this change to make sure the resulting fix is in line with coding best practices for protoactor.

My main concern with this, is that it's a global static, and I would expect that the context would be injected somehow during props constructions or whatever mechanisms protoactor uses.

examples/Patterns/Saga/Program.cs
```diff
@@ -22,11 +24,11 @@ public static void Main(string[] args)
             var busyProbability = 0.01;
             var provider = new InMemoryProvider();
 
-            var props = Actor.FromProducer(() => new Runner(numberOfTransfers, uptime, refusalProbability, busyProbability, retryAttempts, false))
+            var props = Props.FromProducer(() => new Runner(numberOfTransfers, uptime, refusalProbability, busyProbability, retryAttempts, false))
                 .WithChildSupervisorStrategy(new OneForOneStrategy((pid, reason) => SupervisorDirective.Restart, retryAttempts, null));
             
             Console.WriteLine("Spawning runner");
-            var runner = Actor.SpawnNamed(props, "runner");
+            var runner = Context.SpawnNamed(props, "runner");
```
goblinfactory, line 32: I'm not sure if this is correct. please check, thanks.

examples/Patterns/Saga/Runner.cs
```diff
@@ -34,8 +34,10 @@ public Runner(int numberOfIterations, double uptime, double refusalProbability,
 
         private PID CreateAccount(string name, Random random)
         {
-            var accountProps = Actor.FromProducer(() => new Account(name, _uptime, _refusalProbability, _busyProbability, random));
-            return Actor.SpawnNamed(accountProps, name);
+            var accountProps = Props.FromProducer(() => new Account(name, _uptime, _refusalProbability, _busyProbability, random));
+            //TODO: I've never used ProtoActor before, so don't know is this is an appropropriate fix to access a global static Context
+            // perhaps should be injected somehow during proto construction?
+            return Program.Context.SpawnNamed(accountProps, name);
```
goblinfactory: accessing global static Context feels wrong. If this can please be checked, would be much appreciated! txs

examples/Patterns/Saga/TransferProcess.cs
```diff
@@ -146,16 +146,16 @@ private async Task AwaitingCreditConfirmation(IContext context)
...
             {
                 case Started _:
-                    // if we are in this state when started then we need to recreate the TryCredit actor
+                    // if we are in this state when started then we need to recreate the TryCredit actorew
                     context.SpawnNamed(TryCredit(_to, +_amount), "CreditAttempt");
                     break;
                 case OK msg:
-                    decimal fromBalance = await _from.RequestAsync<decimal>(new GetBalance(), TimeSpan.FromMilliseconds(2000));
-                    decimal toBalance = await _to.RequestAsync<decimal>(new GetBalance(), TimeSpan.FromMilliseconds(2000));
+                    decimal fromBalance = await  context.RequestAsync<decimal>(_from, new GetBalance(), TimeSpan.FromMilliseconds(2000));
```
goblinfactory, line 158: this is me guessing how to fix this. Never done protoactor before, please check!

examples/Patterns/Saga/TransferProcess.cs
```diff
@@ -146,16 +146,16 @@ private async Task AwaitingCreditConfirmation(IContext context)
...
+                    decimal toBalance = await context.RequestAsync<decimal>(_to, new GetBalance(), TimeSpan.FromMilliseconds(2000));
```
goblinfactory, line 159: this is a guess, please check.

goblinfactory:
@adamhathcock  Sorry, forgot to cc Adam. Adam this looks like your labour of love, can you please give these fixes a once over to make sure I'm not butchering your sample project? Looks like you've put a lot of effort into it.

goblinfactory:
Lastly: the build on appveyor is still failing after my last fixes, and I don't think that has anything to do with my pull request. The build is failing on this test, `Proto.Tests.FutureTests.TestInATask`
> Error Message: Request didn't receive any Response within the expected time`

If someone can help stabilise that test so this pull request can go in, would be most appreciated. 
[protoactor-dotnet/tests/Proto.Actor.Tests/FutureTests.cs](protoactor-dotnet/tests/Proto.Actor.Tests/FutureTests.cs)
      
      

Cheers, A

goblinfactory:
@adamhathcock  Just realised, this sample project has quite a bit of code in the readme and the code in the readme should also be checked. (damn, wish github supported a simple way to reference, or #include  code blocks from unit tests so that you can guarantee that readme code is always up to date without having to run some build step tool to replace tokens, those approaches are too messy, and takes a lot of fun out of working on the solution. imho.

adamhathcock:
I may have reorganized the examples and added some but I don't believe I had anything to do with the Saga one.  Doens't look familiar anyway.

I contributed the ASP.NET Core DI integration which might be relevant.

tomliversidge:
Saga one was mine @goblinfactory if it's the bank account example, although I'm not very up-to-date with recent changes to Proto.Actor as I've been bogged down with other things for too long :( 

Best way to check if it all still works is run it and see what sort of output you get 👍

examples/Patterns/Saga/Program.cs
cpx86: A new RootContext is really only needed if there's context that needs to be passed through middleware - correct me if I'm wrong @rogeralsing? :) Meaning you could skip declaring your own instance and just use RootContext.Empty instead.

examples/Patterns/Saga/Program.cs
cpx86: It's a correct replacement. As mentioned above though it's prolly simpler to just use RootContext.Empty.

examples/Patterns/Saga/TransferProcess.cs
cpx86: This is correct. Basically, before we used to have two ways of doing Tell and Request - either directly on the PID and via a context. With the context it's possible to have contextual data passed through the chain, allowing for e.g. tracing headers to be propagated between actors and applications. The PID however didn't support this, so we killed it off in favor of only supporting context.

examples/Patterns/Saga/TransferProcess.cs
cpx86: Looks good.

cpx86:
@goblinfactory I think the failing test is a repro of the bug reported in #432, which isn't yet solved. So I think it should be fine to merge this PR even though that test is still not green.

goblinfactory:
> @goblinfactory I think the failing test is a repro of the bug reported in #432, which isn't yet solved. So I think it should be fine to merge this PR even though that test is still not green.

Thanks, that's good to know. I think there's a little bit of work to do still to make sure the sample project actually works. I may have to check out an older commit and run the original code. If I can do that, I can create a simple acceptance test for this PR, something along the lines of 

` when the runner runs, 
then the console output should be . . . `

As soon as the code produces the same output, I'll ask to merge this PR, assuming no other failing tests.
txs all for the quick responses

cheers, 
Alan

goblinfactory:
> Saga one was mine @goblinfactory if it's the bank account example, although I'm not very up-to-date with recent changes to Proto.Actor as I've been bogged down with other things for too long :(
> 
> Best way to check if it all still works is run it and see what sort of output you get 👍
 busy with that now :)

tomliversidge:
I'll try and take a look at this sometime over the weekend. I'm not sure you'll be able to add the sort of acceptance test you're talking about though, as there was a lot of randomness involved if I remember correctly to simulate error conditions

examples/Patterns/Saga/InMemoryProvider.cs
```diff
@@ -65,5 +65,10 @@ public Task DeleteSnapshotsAsync(string actorName, long inclusiveToIndex)
         {
             throw new NotImplementedException();
         }
+
+        Task<long> IEventStore.PersistEventAsync(string actorName, long index, object @event)
```
tomliversidge, line 69: This will cause the saga to fail as the exception will bubble up and kill the actor. You can remove this implementation and have the one above return a Task of long instead

tomliversidge:
I checked out your branch and after updating all the pid.Tell methods to use the new Context.Send approach and removing the exception-throwing persistence implementation it looks like the saga example is working

goblinfactory:
hi all, just seen @tomliversidge 's comment, need to get my head back into this branch as I stopped looking at protoactor when this failed, so still don't have any domain knowledge. will look a bit later. I think it sounds like we're good or very close to being able to merge this depsite the CI build error if that's still there? Possibly might make sense to park this until that issue is resolved? i..e. that's what I'd do if this was a standup at live client project, and we'd all pile in (all systems stop) to remove the bottleneck? What do you think?

examples/Patterns/Saga/InMemoryProvider.cs
goblinfactory: Done.

examples/Patterns/Saga/Program.cs
goblinfactory: Done. Global `Context` is now set to `RootContext.Empty`.

examples/Patterns/Saga/Program.cs
goblinfactory: Done. `RootContext` is now `RootContext.Empty`

goblinfactory:
This is now ready for merging, 
if someone can please merge this in that will be most appreciated.
....I will create a separate integration test for this so that we can confirm the behavior remains consistent and that this can be used as living documentation.
@cpx86  Can you please merge this in? 

cheers,
Alan

examples/Patterns/Saga/InMemoryProvider.cs
tomliversidge: I meant this `IEventStore.PersistEventAsync` implementation should be deleted - the `public Task PersistEventAsync` on line 32 should be updated to return the correct value (`Task.FromResult(nextEventIndex)`) and the signature updated to match the interface (`Task<long>`), rather than us keeping two implementations of the same thing, otherwise it is confusing

tomliversidge:
thanks @goblinfactory - I'll try and check this out tonight and make sure it all runs ok

tomliversidge:
@goblinfactory i'm seeing weird results after checking out the latest code:
screenshot 2018-12-03 at 22 53 42

Running 100 transfers with a 99.99% uptime i would have expected better results than this. I also don't seem to be able to get anything other than success or unknown results back. 

There's some possible explanations:
1. There's still quite a few warnings about obsolete uses of the old .Tell API methods - maybe fixing these will help
2. Changing to RootContext somehow breaks things?
3. Some of the refactorings have broken something?

tomliversidge:
hmm. maybe i spoke too soon... 

screenshot 2018-12-03 at 23 20 14

goblinfactory:
@tomliversidge Yes, I saw that too, however I have no idea if this ever did work or what the correct behavior was supposed to be. I didn't want to "fix" anything, since I don't "know" protoactor. I just wanted to get this in the way it was in, so that the build passes and we can all work on it (properly fix it) so that it can be a learning tool.  As far as I can tell, this is what it's supposed to do? I mean, this is the behavior of the code. I agree that with an uptime of 99% and chance of failure and refusal configures to 0.01 we should get 100% succesful transfers. The refactorings I've done are my attempt to remove the presentation from the actual behavior so that I can start to dig in to what it's doing and so that other can contribute to fixing it. As a new developer never worked with protoactor or complex actor pattern sagas the existing code was too complex to reason about.

I've created another branch with even more refactoring, introduced a 'chaosmonkey' so that the 'probabalistic behavior' can be seperated from the actors. 

The main point of the refactorings was to be able to configure a (seedable) `DeterministicChaosMonkey` together with an `IConsole` ResultsReporter that I can write a series of acceptance tests with. i.e. given these probabilities and starting seeds, when I ... I expect the following results.

Here's the branch:
https://github.com/goblinfactory/protoactor-dotnet/tree/samples/add-saga-project-to-samples-tests/examples/Patterns

And this is the style of test I would like to see:
```csharp
        [Fact]
        public void Happy_path_smoke_test() // todo add in retries and tokenise results, make test deterministic, remove the randomness.
        {
            var console = new TestWriter();
            var results = new TestWriter();
            var gate = new ManualResetEvent(false);

            var saga = new SagaProgram(console, results, false);
            saga.OnFinished += _ => { gate.Set(); };

            saga.Run();
            gate.WaitOne(3000);

            var expected = new [] {
                "RESULTS for 99.99 % uptime, 0.01 % chance of refusal, 0.01 % of being busy and 0 retry attempts:",
                "60 % (3 / 5) successful transfers",
                "0 % (0 / 5) failures leaving a consistent system",
                "0 % (0 / 5) failures leaving an inconsistent system",
                "40 % (2 / 5) unknown results"
            };
            var actual = results.Buffer.Select(l => l.TrimStart().TrimEnd()).ToArray();
            Assert.Equal(expected, actual);
        }
```
Of course, my task is made a tad more *challenging* and *interesting* because I'm starting with code that doesnt work, so breaking a ton of coding rules by refactoring and fixing at the same time.
Given that the code is quite out of date, I wanted to use the readme as a goal, and then simply learn about protoactor and then just implement something that delivered on those requirements using the existing code as a starter guide only. 

I did not realise just how big bite I was taking. Saga's are complex beasts! hahaha.

I love what the saga project teaches and it should be the main piece of teaching code for proto-actor, in my opinion. The code just needs some love.
:)

(update: the actual probabilities above in the actual acceptance test are obviously not correct as you point out, but this is where we can set what we expect, and given certain seed values we can parameterise this test so that we can give examples that show what happens when we have a high uptime, and a very low uptime etc.)

tomliversidge:
@goblinfactory Sounds awesome and your enthusiasm is appreciated! I agree it is quite complex... I think in a real life scenario, using something like Polly to do retries would simplify things, but I wanted to learn / demonstrate the supervision capabilities of the actor model. 

I'll merge this in now anyway, thanks for the contribution!

PR #447, opened 2018-12-20
Include Protos.proto in nuget package fixes #446

### LINKED ISSUE #446: Supply Proto.Actor/protos.proto in nuget package so it can be included
It is a common requirement to reference actor.PID in messages definitions, but the file that contains that definition is not distributed with the nuget package. 

This makes referencing the file in a way that will work across multiple developer machines tiresome.

I propose that the Proto.Actor/protos.proto file is included in the nuget package so that users can reference the included folder like this protoc -I=$(NugetPackageRoot)proto.actor/2.0.0/protos 

I have made a feature suggestion for protoc to include these files from within nuget packages without requiring th euser to specify the include path see: protocolbuffers/protobuf#5437

sudsy:


sudsy:
I don't understand why the appveyor build on this has failed. The changes in this PR are tiny and unlikely to be the cause of the failure. Also, the tests work properly locally. I can only guess that it is a transient error in appveyor. 

Should I try a new pull request or a dummy commit?

potterdai:
@sudsy my PR got failed too. I was working on go version yesterday. Will look into it!

Issue #445, opened 2018-12-12
Send a DeadLetter when the actor is stopping 


alexeyzimarev:
When the actor is gone and there is a message for it,  the `DeadLetter` is published. But when the actor is stopped but not removed from the registry yet, we log an error and then do nothing, so the message gets lost. I believe in such case we should publish a `DeadLetter` as well, thoughts?

```
        public Task InvokeUserMessageAsync(object msg)
        {
            if (_state == ContextState.Stopped)
            {
                //already stopped
                Logger.LogError("Actor {0} already stopped, ignore user message {1} {@2}", Self, msg.GetType().Name, msg);
                return Done;
            }
```

potterdai:
@alexeyzimarev Hmmm, makes sense, it means we should change the code you mentioned to the following?
```
        public Task InvokeUserMessageAsync(object msg)
        {
            if (_state == ContextState.Stopped)
            {
                //already stopped
                DeadLetterProcess.Instance.SendUserMessage(this.Self, msg);
                return Done;
            }
```
A PR is very welcomed = )

alexeyzimarev:
Yes, I did it in my fork. Is delivers more consistent behaviour, because when the actor has been already stopped, the messages to it will be published as dead letter. I see no reason why for a stopping actor messages just vanish.

potterdai:
Fixed in #448

PR #448, opened 2018-12-20
Send to DeadLetter process when actor has already stopped but not yet removed from registry

potterdai:
As is discussed here:
https://github.com/AsynkronIT/protoactor-dotnet/issues/445


alexeyzimarev:
Seems that features tests started to fail again...

PR #450, opened 2019-01-01
Add RequestWithCustomSender method to ISenderContext

potterdai:
This is necessary in situation like routing.


alexeyzimarev:
The idea is good but I don't really like the name. Why can't we have an overload for `Request` and `RequestAsync` that will take one additional parameter?

potterdai:
@alexeyzimarev Makes sense!
I found this method necessary when doing the porting to protoactor-go, and directly took the naming from there.
Will make a new PR for this change! Thank you for the suggestion! = )

PR #452, opened 2019-01-04
Add IStopperContext.

potterdai:
This PR is to add IStopperContext.
Moving pid.Stop, pid.StopAsync, pid.Poison, pid.PoisonAsync into IStopperContext.

### DIFF
--- a/examples/ClusterGrainHelloWorld/Messages/Protos_protoactor.cs
+++ b/examples/ClusterGrainHelloWorld/Messages/Protos_protoactor.cs
@@ -109,7 +109,7 @@ namespace Messages
                 }
                 case ReceiveTimeout _:
                 {
-                    context.Self.Stop();
+                    context.Stop(context.Self);
                     break;
                 }
                 case GrainRequest request:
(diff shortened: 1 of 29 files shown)


alexeyzimarev:
I am genuinely wondering, what is the reasoning behind this change?

potterdai:
@alexeyzimarev 
Hi, thank you for asking!
Previously, we have `pid.Tell()`, `pid.Request()`, and `pid.Stop()` is part of this API logic. As now we have deprecated them and move those functions into `ActorContext`, leaving `pid.Stop()` behind is not very consistent from my point of view.
What do you think?

alexeyzimarev:
So basically `PID` won't be able to do anything? Would it also make sense to move `Poison` out of the `PID` class as well?

potterdai:
@alexeyzimarev That's correct! It has been changed in this PR as well. Check this out:
https://github.com/AsynkronIT/protoactor-dotnet/blob/6c6e5eb3ddefce2fb33790e3bb4694d1c10fc3a3/src/Proto.Actor/IStopperContext.cs#L13-L17
I'm still waiting to see others feedback about this, and if there is any better/neater way of handling this.
But overall as you said, `PID` won't be able to do anything.

alexeyzimarev:
Shall we go on and merge this?

potterdai:
@alexeyzimarev Thank you for remind!
@rogeralsing what you think on this one?

PR #454, opened 2019-01-10
Fixed Task Was Cancelled Exception

andreabalducci:
Found an unexpected exception cancelling recurring scheduled message when my actor is Stopping.


alexeyzimarev:
Is that enough? I remember that cancelled tasks always throw, so in my fork, I wrapped the whole thing in a try-catch. It is especially relevant for the `ScheduleTellRepeatedly`, when I can cancel the task to remove the schedule.

andreabalducci:
yes, problem gone with async task.

PR #455, opened 2019-02-14
Move Parent()/Self()/Sender()/Actor() into InfoContext for shared access

potterdai:
Move Parent()/Self()/Sender()/Actor() into InfoContext for shared access in IContext, ISenderContext and IReceiverContext.


potterdai:
Merge following go version

PR #464, opened 2019-04-29
Fixes #462

### LINKED ISSUE #462: EndpointWriterMailbox loops infinitely on restart
I found it quite difficult to write a test for this. I am open to suggestions for how it might be done.

I only identified the problem in production when a remote went offline and there was still many messages to be delivered to the remote. The mailbox was pushing multiple messages to the actor in parallel.

I think it will be easier to write a test for this after #431 is complete.

For now, I will make a pull request with the fix, the changes bring the behaviour more into line with the normal mailbox so hopefully the logic behind my fix is pretty clear.

sudsy:


potterdai:
@sudsy Thank you! Merge right the way = )

Issue #460, opened 2019-03-27
dotnet core nuget packages outdated


kenavn:
The targeting of .Net Framework 4.5.2. causes problems when trying to upgrade nuget references; specifically the "Microsoft.Extensions.Logging" package which does not any longer provide "shims" for 4.5.2. It requires .Net Standard 2.0 compliancy, which is from .Net Framework 4.6.2 and onwards.

Generally there are a lot of outdated packages, and the project seems a bit stale at the moment. 

I have updated the targeting to .Net Standard 2.0 and .Net Framework 4.6.2. in a pending PR which is not yet committed. The thing is that one in reality can remove all 4.6.2 specifics, as it already is .Net Standard 2.0 compliant. At least to my understanding.

The RavedDB code has a few preprocessor directives that distinguished between different API versions. Except for that the upgrade seems pretty straight forward.

Any thoughts on this? Was there a particular reason that 4.5.2 is still targeted? Anyhow; the world moves forward and protoactor need to follow along to be relevant... :)

kenavn:
BTW; worth mentioning that upgrading to .Net Framework 4.7.2 is recommended, even though nuget considers 4.6.1 -> to be "pretty" compliant ;)

potterdai:
@kenavn Thank you for remind! I'll try make a PR to strip away 4.5.2! = )

potterdai:
@kenavn Should be addressed in #466

PR #466, opened 2019-05-02
Strip net 4.5.2. Upgrade libraries to latest.

potterdai:
This should also address the issue https://github.com/AsynkronIT/protoactor-dotnet/issues/460


potterdai:
@rogeralsing Should I go ahead and merge it? 4.5.2 will not be supported after the merge = )

Issue #475, opened 2019-08-10
Proposal: remove net461 as a target


sudsy:
I have been having some trouble building on my mac since the change of target framework from

`<TargetFrameworks>netstandard2.0;net452</TargetFrameworks>`

to

`<TargetFrameworks>netstandard2.0;net461</TargetFrameworks>`

I may be able to diagnose why I can't build but it has occurred to me that because protoactor consists primarily of libraries that are consumed by other applications, the inclusion of net461 may not be necessary. 

netstandard2.0 target will produce a library that is compatible with .net core and netfx.

As for the non library projects in the repo, they are all test projects so I would propose we target netcoreapp only for these. That way it will compile cross platform.

alexeyzimarev:
I don't think we should fix build issues by simply removing a legit target. There're a few ways to fix the build issue and I submitted a PR that fixes the issue without removing the target.

sudsy:
That builds for me, so I'm happy with that solution as an alternative to my PR in #476.

I had to look back at the guidance on this though to see what the impact of including the other target is:

The [guidance](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting) says :

❌ AVOID multi-targeting as well as targeting .NET Standard, if your source code is the same for all targets.

The .NET Standard assembly will automatically be used by NuGet. Targeting individual .NET implementations increases the *.nupkg size for no benefit.

But then it says:

✔️ CONSIDER adding a target for net461 when you're offering a netstandard2.0 target.

Using .NET Standard 2.0 from .NET Framework has some issues that were addressed in .NET Framework 4.7.2. You can improve the experience for developers that are still on .NET Framework 4.6.1 - 4.7.1 by offering them a binary that is built for .NET Framework 4.6.1.

It's certainly a bit vague about why you might consider adding 461 when in our case the code is the same for all targets. So tried to dig a little deeper and only came up with problems with having the dual target described [here](https://github.com/dotnet/docs/issues/8896).

I am happy to go with your solution, I have more of an academic curiosity about why we are including it more than anything. If you can shed any light on it that would be great.

alexeyzimarev:
I think when .NET Core 3 will be released, we'd need to make another decision on how to move forward.

sudsy:
Thanks

PR #477, opened 2019-08-13
Fixes the build issues explained in #475

alexeyzimarev:
I added the `netfx.props` file and imported to those projects which were failing to build on Mono.

Some formatting got changed by the IDE but I think it's not bad.


alexeyzimarev:
@PotterDai @rogeralsing any objections, can I merge this?

PR #480, opened 2019-11-18
Removing the KVs usage from consul and Move MemberStatusValue to Consul Metadata.

potterdai:
This is to follow this PR in protoactor-go
https://github.com/AsynkronIT/protoactor-go/pull/362


alexeyzimarev:
That's really good. The KV thing was a bit of a mess when running as a Deployment in k8s.

Issue #485, opened 2020-01-19
Get rid of net452 build target


alexeyzimarev:
Today, the .NET Framework 4.5.2 build target creates too many impediments for updating dependencies. Apparently, many libs like SQLite and tons of Microsoft own libs are .NET Standard 2.0 only.

I could suggest stopping to support .NET Framework 4.5.2 and to support .NET Standard 2.0 only.

cpx86:
I'm all for this - supporting anything below 4.6.1 isn't necessary. @alexeyzimarev up for submitting a PR? :)

alexeyzimarev:
Yeah, I started by simply removing a couple of remaining `async void`s in the scheduler but ended up doing more, have it almost ready. Will submit soon.

alexeyzimarev:
@cpx86 in my PR I removed 4.6.1 target as well, leaving only .NET Standard. Is it a good move? I would say yes, it allowed me to use the latest version of Sqlite client library without an ugly `#if`. If we need to ensure that everything works with 4.6.1, I can change test projects to have two targets.

PR #487, opened 2020-02-23
Addressing #485

alexeyzimarev:
It might look a bit too much but overall it is about #485:

- Removing obsolete targets
- Removing conditional code blocks
- Upgrading packages
- Fixing some remaining async void calls
- Little cleanup here and there, shortening code with C# 8
- Upgrading test projects to .NET Core 3.1

Adding the dotSettings file can be opinionated but it would be nice to find a way to address the code style issue. We can also use the `editorconfig` file but it's a bit more cumbersome to set up.


rogeralsing:
@alexeyzimarev conflicts

alexeyzimarev:
Yeah, my own changes, I suppose :)

PR #497, opened 2020-03-12
Actor system

rogeralsing:
BOOOM!

First PoC of removing the static Process registry from Proto.Actor.

I have only updated Proto.Actor and Proto.Actor.Tests.

_Most_ tests in there work. there are a few failures. 

This is a draft PR to let everyone play around with the concept and see if this is the direction we should go.

Things that needs to be sorted out, is how Props relates to ActorSystem.

Currently, you have to pass a system to the FromProducer and FromFunc.

Props.FromFunc(system, ....)

Maybe something like:

System.Props.FromFunc( myfunc);

might be better?

Anyway, I was able to do this without hardwiring PID to the actorsystem, getting around the major major issues in Akka with serialization.


rogeralsing:
cc @PotterDai @alexeyzimarev @cpx86

rogeralsing:
Edit. Fixed failing tests.

rogeralsing:
The need for System inside props, can probably be removed if there is some clever way to change the DefaultSpawner logic.

cpx86:
Looks interesting - gonna try it out :)

Damian-P:
In the pull request #498 I removed the need for ActorSystem inside Props. I changed the DefaultSpawner logic by changing the delegate to pass the ActorSystem.

Is it a good move ?

cpx86:
> In the pull request #498 I removed the need for ActorSystem inside Props. I changed the DefaultSpawner logic by changing the delegate to pass the ActorSystem.
> 
> Is it a good move ?

Conceptually, I think it makes sense. The Props really only describe how to create, setup and spawn an actor. I don't think it needs to be tied to a specific ActorSystem instance. 

An interesting part of your change is that it passes the ActorSystem to the Middlewares. In the current ActorSystem PR there would be no way for a middleware to access e.g. the ProcessRegistry, DeadLetter or EventStream. This could probably also be solved though by adding the ActorSystem to the IInfoContext and having the middlewares access it from there.

@rogeralsing Thoughts?

rogeralsing:
>This could probably also be solved though by adding the ActorSystem to the IInfoContext and having the middlewares access it from there.

Agree, that is probably the cleanest way imo.

@Damian-P very impressive PR, many thanks for this!

Damian-P:
@rogeralsing you're welccome !

alexeyzimarev:
Good stuff 👍

rogeralsing:
Let's merge both PR's?, and continue to reason about how to pass actor system to middleware, if we keep the current parameter approach, or, as @cpx86 says and add it to the infocontext

Any objections or go?

cpx86:
Let's go :)

PR #503, opened 2020-03-17
SimpleSchedulerExtensions

Damian-P:
## Description
Allows a SenderContext to be a scheduler.

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/src/Schedulers/Proto.Scheduler.SimpleScheduler/SimpleSchedulerExtensions.cs
+++ b/src/Schedulers/Proto.Scheduler.SimpleScheduler/SimpleSchedulerExtensions.cs
@@ -1,0 +1,17 @@
+using System;
+using System.Threading;
+
+namespace Proto.Schedulers.SimpleScheduler
+{
+    public static class SimpleSchedulerExtensions
+    {
+        public static ISimpleScheduler ScheduleTellOnce(this ISenderContext senderContext, TimeSpan delay, PID target, object message) =>
+        new SimpleScheduler(senderContext).ScheduleTellOnce(delay, target, message);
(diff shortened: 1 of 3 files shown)


rogeralsing:
Nice!

src/Schedulers/Proto.Scheduler.SimpleScheduler/SimpleSchedulerExtensions.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using System.Threading;
+
+namespace Proto.Schedulers.SimpleScheduler
+{
+    public static class SimpleSchedulerExtensions
+    {
+        public static ISimpleScheduler ScheduleTellOnce(this ISenderContext senderContext, TimeSpan delay, PID target, object message) =>
```
alexeyzimarev, line 8: I like these extensions, just have a suggestion to keep the code formatting aligned. That's why I included the `DotSettings` file to the repository. But it only works with Rider/RestSharp. If that's not enough, we could consider adding the editorconfig file.

Another suggestion is to pay attention to triple-slash comments for public APIs, so when we're ready to generate the API docs, everything gets properly documented.

Damian-P: Sorry, I'm using vscode with the default code formatter. You know how to adapt those settings to vscode ?

alexeyzimarev: VsCode only supports editorconfig but I am not sure how to export DotSettings to VsCode. I will fix it for you, no worries. Will check if it's not much work to create a proper editorconfig.

PR #504, opened 2020-03-17
Router creation from sender context

Damian-P:
### Allow a SenderContext to create router props.

This PR adds extension methods to ISenderContext to create router props. The ISenderContext instance used to create the router props is used in the router implementations to dispatch the messages.

**I removed the router class but maybe it is still needed ?**

Since #500 is adding the ActorSystem to IInfoContext, we can still use the RootContext to create routers or use the ActorContext.

## Purpose
This pull request is a:

## Checklist


rogeralsing:
@Damian-P build failure. any change that causes it or some random threading error?

Damian-P:
@rogeralsing I don't see any change wich could causes this test run failure.
All the commits before passed and the last is just an indentation fix.

But since there is no test method in the Proto.Cluster.Tests project and it has been executed at the very end this time, it maybe causes that failure ?
Or juste as you said, random threading error.

PR #505, opened 2020-03-18
Merge projects

rogeralsing:
Merging Mailbox and Router into core Proto.Actor project


src/Proto.Actor/Mailbox/Mailbox.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace Proto.Mailbox
 {
```
cpx86: What's the reason for changing these to public?

cpx86: Same for BoundedMailboxQueue

rogeralsing: We currently rely on the internalvisibleto thing for testing.
We could ofc still do that for the core lib. it just seems hacky.

Should we keep internal?

alexeyzimarev: I think internalvisibleto is ok as long as it won't create a lot of issues with signing

cpx86: For low-level unit tests I think InternalsVisibleTo is the way to go. Better than exposing internal classes in the API surface.

cpx86: (tbh in general I think we could and probably should make some other public classes internal)

rogeralsing: Brought back internal and also moved mailbox and router tests into same project as the actor tests.

PR #506, opened 2020-03-21
Fixing nullable warnings

alexeyzimarev:
After Mailbox and Router projects were removed, there are a lot of references to those projects left. I removed all those references.
Also, I fixed all nullable warnings.


alexeyzimarev:
Looks like my PR broke some tests, I will fix it today.

PR #511, opened 2020-04-02
Alternate grain factory

### LINKED ISSUE #510: Passing Grain id to grain factory function
I have been working with the grain implementation and think I may have found a small improvement.

With Orleans Grains, I have previously used the id of the grain to alter the behaviour of the grain.  For instance, I have a grain per user id that represents a data store for that particular user.

I noticed that under the current grain implementation there is no way to get the id of the grain passed to the grain client function.

I propose changing the signature of the Grain factory function from Func<ISomeGrain> to Func<String, ISomeGrain> and passing the id that is used to create the client to that factory function.

This would be a breaking change. Does anyone have any objections?

sudsy:
## Description

Resolves #510 

## Purpose
This pull request is a:

- [ x] New feature (non-breaking change which adds functionality)

## Checklist

I have tested this feature in my own project that uses protoactor. There are currently no tests for the grain generator only examples. I have rebuilt the ClusterGrainHelloWorld example and have run it using `docker-compose up --build` to make sure that the old factory function is still working.  I have tested the new and old factory functions in a seperate project.

### DIFF
--- a/examples/ClusterGrainHelloWorld/Messages/Protos_protoactor.cs
+++ b/examples/ClusterGrainHelloWorld/Messages/Protos_protoactor.cs
@@ -18,12 +18,17 @@ namespace Messages
             Cluster = cluster;
         }
 
-        internal Func<IHelloGrain> _HelloGrainFactory;
+        internal Func<String,IHelloGrain> _HelloGrainFactory;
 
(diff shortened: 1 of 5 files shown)


rogeralsing:
This is great! many thanks

Issue #518, opened 2020-05-22
Cluster partition Actor PID transfer issue (duplicate grains)


DenizPiri:
## Expected Behavior
When adding new nodes to a cluster, a single grain name should always map to a consistent PID.

## Actual Behavior
After N number of nodes, after different nodes of the cluster are requesting PIDs, they start getting different PIDs.

The actual problem is masked by the PidCache. When that is disabled, the problem transforms into PID changing as new nodes are added to the cluster.

I determined the problem to be related to Actors in a partition not being transferred correctly. When I dived deeper, I figured out that the problem is caused by a node being added to the MemberList even if the Alive property is false (consul provider actively uses this property). Together with that addition, code tries to determine which actors should be transferred, however, inside Rendezvous.GetNode() the Alive property of members are checked and skipped if they are not Alive.

So basically new node gets added, it is not Alive, no transfer gets performed. When the node later comes alive, code doesn't check for old.Alive != new.Alive state and transfer the actor PIDs.

Adding this at the top of the MemberList.UpdateAndNotify() function fixes the problem.

```
if (old != null && old.Alive == false)
    old = null;
if (@new != null && @new.Alive == false)
    @new = null;
```

However, I am thinking of making a better solution by adding a new MemberAlivenessUpdated message and triggering actor PID transfers only in that case. I will try to make a pull request soon.

DenizPiri:
https://github.com/AsynkronIT/protoactor-dotnet/pull/519 is merged and fixes this issue. Closing this. 🎉

PR #519, opened 2020-05-22
Fix for duplicate grain issue upon new nodes joining the cluster

DenizPiri:
## Description

This fixes the issue mentioned here https://github.com/AsynkronIT/protoactor-dotnet/issues/518

Approach, I took was to make sure that only Alive nodes are in the members lists. This technically affects the downing time to be shorter, from 30 secs (DeregisterCritical) to 10 secs (ServiceTtl) defaults. This also changes the startup behavior, by making sure that our current node is marked as alive before using it as a partition. This prevents potential issues with cluster startup, however, at the same time this introduces a bit of delay until the first cluster node can start spawning grains. I am thinking of adding a Cluster.WaitUntilAnyAliveAsync() like function for this purpose.

### DIFF
--- a/src/MemberStrategies/WeightedMemberStrategy/WeightedMemberStrategy.cs
+++ b/src/MemberStrategies/WeightedMemberStrategy/WeightedMemberStrategy.cs
@@ -18,7 +18,7 @@ namespace Proto.Cluster.WeightedMemberStrategy
         public WeightedMemberStrategy()
         {
             _members = new List<MemberStatus>();
-            _rdv = new Rendezvous(this);
+            _rdv = new Rendezvous();
             _wrr = new WeightedRoundRobin(this);
         }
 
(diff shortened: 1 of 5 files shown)


DenizPiri:
I actually forgot to check how protoactor-go behaves. Differences in behavior could cause issues for cross-language clusters. Then again, is there anyone using that in practice?

alexeyzimarev:
We actually moved cluster providers to the contrib repo. I need to complete the final step by removing projects from the core repo. I thought it's already done...

DenizPiri:
@alexeyzimarev Core fix is not for the cluster provider though, it is for the Cluster module. This commit is the core fix: https://github.com/AsynkronIT/protoactor-dotnet/pull/519/commits/ec9d0877411ac033a25a970f9d91b02176eea936

alexeyzimarev:
I understand that but also saw some files changed in the Consul provider.

DenizPiri:
@alexeyzimarev Alright, I removed that commit from the PR. I hope it is fine to merge now.

alexeyzimarev:
We need the contribution agreement signed by you.

DenizPiri:
@alexeyzimarev CLA link doesn't seem to be working. I mean, I see this, every single time I try: 
image
But the status here stays as "not signed yet".

DenizPiri:
@alexeyzimarev CLA seems to be signed now!

PR #522, opened 2020-06-16
Get Mailbox to support canceled tasks gracefully.

DenizPiri:
## Description

Without this when CancellationTokens were in use or the OperationCancelledException was explicitly thrown by an actor, it was causing that actor's Mailbox to freeze forever and never continue processing the new messages.

This technically fixes issues inside the consul cluster provider as well, because of the consuldotnet library using HttpClient and that throwing TaskCancelledException in certain cases, such as Consul agent being restarted.

## Purpose
This pull request is a:

## Checklist


cpx86:
LGTM! Nice catch :)

PR #523, opened 2020-06-16
Fix the incorrect measurement with InprocessBenchmark.

DenizPiri:
## Description

The existing code was sending messages not to the EchoActor but the PingActor itself after the first batch. This change makes the benchmark seem a lot worse in terms of performance, but it is a lot more realistic than an actor sending messages to itself only. However, I believe this benchmark could be improved a little bit to yield more messages/sec; by changing the batching idea a bit so that there would be 2 batches in progress at any given time, not just one.

## Purpose
This pull request is a:

## Checklist


rogeralsing:
You are correct! many thanks for spotting this

PR #525, opened 2020-06-30
Convert protograin to a dotnet tool

alexeyzimarev:
## Description

Protograin looks like a forgotten child and needs some love.
I added some proper command-line handling with help and versioning and made it pack to a don't cli tool, which can be installed by `dotnet tool install <--global>` and used on any platform.


rogeralsing:
Very nice.

I'm thinking it might be possible to take all the type metadata and feed it to a generic code template.
That would allow users to make custom implementations of grains too, to avoid vendor lockin of the exact details.

but this is a very very nice first step!

alexeyzimarev:
I now need to check if the package works :)

Issue #530, opened 2020-07-06
Take ownership process doesn't look complete 


alexeyzimarev:
When the new cluster member joins, the existing nodes try to rebalance the load by transferring ownership to the new node.

However, the process is fire and forget, the ownership transfer has no confirmation.

Also, actors that have been transferred are only removed from the partition but stay active in the process registry and in the PidCache. When attempting to get an actor on the same node, it will be returned from the PidCache, pointing to the current node. I think it might lead to the same actor being split and get to a split-brain problem and inconsistent state.

I'd say we need to stop transferred actors and by doing so wipe them from the PidCache, forcing the next call to `GetActorAsync` to ask the partition for the new location of the actor.

alexeyzimarev:
Another thing is that I might be missing something important here, but when the ownership transfer message gets to the new node, it just adds the PID to the partition. But it doesn't spawn the actor. When I try getting n actor by name and kind from the cluster on that node, the PID is returned by the partition but there's no actor.

alexeyzimarev:
Ok, I was tracing the issue and discovered the following:

- we send transfer ownership messages to the other node
- the other node puts the actor in its partition
- the sending node also moves it to the remote node partition
- that node goes down permanently

As a result, zombie addresses are present on the remaining nodes in PidCache and in the partitions list for that non-existent member and all attempts to talk to those actors naturally fail.

Removing the whole ownership transfer process and replacing it with the actor stop, partition cleanup and pid cache cleanup on the node that transfers the ownership from, solves the issue. Actors disappear from the node and requests to the actor triggers the cluster requesting the remote actor from the correct node.

alexeyzimarev:
Now I managed to remove zombie actors from partitions but some of them still hang in the pid cache on some nodes. Need to investigate more.

alexeyzimarev:
The solution that works in my setup now includes:

- Do now use the `TakeOwnership` message at all, don't send it and don't handle it in the `PartitionActor`
- Change the `PartitionActor.TransferOwnership` to
  - stop the transferred actor instead of unwatching it
  - remove the transferred actor `PID` from the cluster `PidCache`

After applying these changes, I am able to run the cluster, increase and decrease the number of nodes with the cluster rebalancing itself. Senders that use `Cluster.GetActorAsync` are getting the new remote address of the actor after it has been removed from the current node.

[...]

PR #531, opened 2020-07-09
Working code that fixes the ownership issues in a cluster

alexeyzimarev:
Here is what I finally came up with for fixing #530 

Basically, it doesn't involve any ownership transfer communication between nodes. Hopefully, it is in line with the idea of unmanaged partitioning of Proto.Actor.

### DIFF
--- a/src/Proto.Cluster/Cluster.cs
+++ b/src/Proto.Cluster/Cluster.cs
@@ -19,10 +19,12 @@ namespace Proto.Cluster
         private static readonly ILogger Logger = Log.CreateLogger(typeof(Cluster).FullName);
 
         internal ClusterConfig? Config { get; private set; }
-        
+
         public ActorSystem System { get; }
 
         public Remote.Remote Remote { get; }
+
+        public bool Ready => MemberList?.MembersCount > 0;
 
         public Cluster(ActorSystem system, Serialization serialization)
         {
@@ -33,9 +35,9 @@ namespace Proto.Cluster
             PidCache = new PidCache(this);
         }
 
-        internal Partition Partition { get; }
-        internal MemberList MemberList { get; }
-        internal PidCache PidCache { get; }
+        public Partition Partition { get; }
+        public MemberList MemberList { get; }
+        public PidCache PidCache { get; }
 
         public Task Start(string clusterName, string address, int port, IClusterProvider cp)
             => Start(new ClusterConfig(clusterName, address, port, cp));
@@ -75,6 +77,7 @@ namespace Proto.Cluster
         public async Task Shutdown(bool graceful = true)
         {
             Logger.LogInformation("[Cluster] Stopping...");
+
             if (graceful)
             {
                 await Config!.ClusterProvider.Shutdown(this);
@@ -124,13 +127,13 @@ namespace Proto.Cluster
                     : await System.Root.RequestAsync<ActorPidResponse>(remotePid, req, ct);
                 var status = (ResponseStatusCode) resp.StatusCode;
 
-                switch (status)
+                if (status == ResponseStatusCode.OK)
                 {
-                    case ResponseStatusCode.OK:
-                        PidCache.TryAddCache(name, resp.Pid);
-                        return (resp.Pid, status);
-                    default: return (resp.Pid, status);
+                    PidCache.TryAddCache(name, resp.Pid);
                 }
+
+                Logger.LogDebug("[Cluster] Obtained remote PID {PID} from {Partition}:{Remote} {Status}", resp.Pid, address, remotePid, status);
+                return (resp.Pid, status);
             }
             catch (TimeoutException e)
             {
--- a/src/Proto.Cluster/Cluster.cs
+++ b/src/Proto.Cluster/Cluster.cs
@@ -35,9 +35,9 @@ namespace Proto.Cluster
             PidCache = new PidCache(this);
         }
 
-        public Partition Partition { get; }
-        public MemberList MemberList { get; }
-        public PidCache PidCache { get; }
+        internal Partition Partition { get; }
+        internal MemberList MemberList { get; }
+        internal PidCache PidCache { get; }
 
         public Task Start(string clusterName, string address, int port, IClusterProvider cp)
             => Start(new ClusterConfig(clusterName, address, port, cp));
@@ -118,7 +118,7 @@ namespace Proto.Cluster
                 Name = name
             };
 
-            Logger.LogDebug("[Cluster] Requesting remote PID from {Partition}:{Remote} {@Request}", address, remotePid, req);
+            Logger.LogTrace("[Cluster] Requesting remote PID from {Partition}:{Remote} {@Request}", address, remotePid, req);
 
             try
             {
@@ -132,7 +132,7 @@ namespace Proto.Cluster
                     PidCache.TryAddCache(name, resp.Pid);
                 }
 
-                Logger.LogDebug("[Cluster] Obtained remote PID {PID} from {Partition}:{Remote} {Status}", resp.Pid, address, remotePid, status);
+                Logger.LogTrace("[Cluster] Obtained remote PID {PID} from {Partition}:{Remote} {Status}", resp.Pid, address, remotePid, status);
                 return (resp.Pid, status);
             }
             catch (TimeoutException e)
--- a/src/Proto.Cluster/Partition.cs
+++ b/src/Proto.Cluster/Partition.cs
@@ -13,7 +13,7 @@ using Proto.Remote;
 
 namespace Proto.Cluster
 {
-    public class Partition
+    class Partition
     {
         private readonly Dictionary<string, PID> _kindMap = new Dictionary<string, PID>();
 
@@ -82,7 +82,7 @@ namespace Proto.Cluster
         private readonly Dictionary<PID, string> _reversePartition = new Dictionary<PID, string>();                      //PID to grain name
         private readonly Dictionary<string, SpawningProcess> _spawningProcs = new Dictionary<string, SpawningProcess>(); //spawning processes
 
-        public Cluster Cluster { get; }
+        private Cluster Cluster { get; }
 
         public PartitionActor(Cluster cluster, string kind)
         {
@@ -132,7 +132,7 @@ namespace Proto.Cluster
         {
             Logger.LogInformation("[Partition] Kind {Kind} member left {Address}", _kind, memberLeft.Address);
 
-            EnsureNewAddresses(context);
+            EnsureNewAddresses();
 
             // Process Spawning Process
             MakeUnavailable(memberLeft.Address);
@@ -142,7 +142,7 @@ namespace Proto.Cluster
         {
             Logger.LogInformation("[Partition] Kind {Kind} member rejoined {Address}", _kind, memberRejoined.Address);
 
-            EnsureNewAddresses(context);
+            EnsureNewAddresses();
 
             // Process Spawning Process
             MakeUnavailable(memberRejoined.Address);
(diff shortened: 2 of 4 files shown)


src/Proto.Cluster/Cluster.cs
```diff
@@ -33,9 +35,9 @@ public Cluster(ActorSystem system, Serialization serialization)
             PidCache = new PidCache(this);
         }
 
-        internal Partition Partition { get; }
-        internal MemberList MemberList { get; }
-        internal PidCache PidCache { get; }
+        public Partition Partition { get; }
```
alexeyzimarev: This isn't required, I think I exposed these properties for monitoring purposes.

src/Proto.Cluster/Partition.cs
```diff
@@ -168,92 +156,53 @@ private void MakeUnavailable(string address)
...
-                {
-                    transferredActorCount++;
-                    TransferOwnership(actorId, address, context);
-                }
-            }
+            Logger.LogInformation("[Partition] Kind {Kind} member joined {Address}", _kind, msg.Address);
 
-            return transferredActorCount;
-        }
+            while (!Cluster.MemberList.IsMember(msg.Address)) { }
```
rogeralsing, line 163: What's going on here? full spinwait for address to become member?

src/Proto.Cluster/PidCache.cs
```diff
@@ -37,9 +40,12 @@ internal void Stop()
 
         private void OnMemberStatusEvent(MemberStatusEvent evn)
         {
-            if (evn is MemberLeftEvent || evn is MemberRejoinedEvent)
+            switch (evn)
             {
-                RemoveCacheByMemberAddress(evn.Address);
+                case MemberLeftEvent _:
+                case MemberRejoinedEvent _:
+                    RemoveCacheByMemberAddress(evn.Address);
```
rogeralsing, line 47: The means a full clear of pid cache on both join and leave events. right?

rogeralsing:
When a node joins, and should algorithmically be the new owner of an id, how is that handled now?
The previous owner clears its cache, but the newly joined node, what does that do?

src/Proto.Cluster/Partition.cs
alexeyzimarev: I might be overly cautious here but messages to the partition actor might not necessarily be processed yet. I know that event stream subscriptions process events immediately, so it might not be needed.

src/Proto.Cluster/PidCache.cs
alexeyzimarev: That's right because partitions change in both cases.

alexeyzimarev:
> The previous owner clears its cache, but the newly joined node, what does that do?

They do nothing as they weren't really doing anything before, that was the whole issue. They get supposedly new address in the partition but there's no actor that is spawned for that address. The PidCache says "I don't have it, let's as the partition", the partition says "here it is", we send a message to it and it never reaches anything because there's no actor.

Now, consumers ask the cluster for the actor using `GetActorAsync` and it cannot find it anywhere, so a new actor gets spawned in the correct partition of the correct node.

alexeyzimarev:
I can actually demonstrate it working with a dynamically scaled cluster in Kubernetes. I run it now for a few days, scaling it up and down without issues. The only small issue that remains is that the remote endpoint only dies after a while since it is not involved in cluster rebalancing process.

rogeralsing:
>They do nothing as they weren't really doing anything before, that was the whole issue. They get supposedly new address in the partition but there's no actor that is spawned for that address. 

There is something fishy here.
The idea with takeownership is that the new partition should know where the old actors live, there shouldnt be any new actors spawned.
spawning new actors will lead to a lot of zombie actors lingering. 

e.g.


Actor Foo lives on node 3, the ID lives on node 1

Someone asks for Foo, node 1 is the algorithmic owner, so we ask node 1, node 1 says Foo actor is on node 3.
we now cache and remember that Foo is on node 3.

If a new node joins, node 4, and if that happens to be the new algorithmic owner of ID Foo. the idea is to let node 1 send the owenrship over to node 4. so node 4 knows that Foo actor lives on node 3.

If node 4 does not know this, and don't get informed, it will spawn a new actor. leaving the old perfectly good actor around until it kills itself (in Grains)

Theoretically it will work with the approach in this PR. But, there will be a  problem with zombie actors. there is no cleaning up, and needlessly much new actor spawning.

I am however open to go with this PR as a first step as the existing code is clearly broken, so one step closer to stability is good.

alexeyzimarev:
> If node 4 does not know this, and don't get informed, it will spawn a new actor. leaving the old perfectly good actor around until it kills itself (in Grains)

Actors (not grains) do not have mandatory receive timeout and therefore live forever. But even for the period of the receive timeout, before the actor or grain gets killed, it will receive messages if the sender is on the same node. This is the first line of the `Cluster.GetActorAsync`:

```csharp
            if (PidCache.TryGetCache(name, out var pid)) return (pid, ResponseStatusCode.OK);
```

So, two actors will exist either forever (normal actor without the timeout) or during the timeout period. So, if the sender is on the same node, it will keep sending messages to the cached actor, whilst other nodes will send messages to the correct actor. It creates a lot of risks because state of both actors will be inconsistent.

alexeyzimarev:
Also, for leaving nodes we just clean the address from the partition, but the `PidCache` keeps all those actors for the node, which left. So in case the stopped node hasn't properly sent termination messages to all the watchers on other nodes, the address of the dead node will still be returned on other nodes and messages either disappear (`Send`) or timeout (`RequestAsync`)

rogeralsing:
>So, two actors will exist either forever (normal actor without the timeout) or during the timeout period. So, if the sender is on the same node, it will keep sending messages to the cached actor, whilst other nodes will send messages to the correct actor

If everything worked as intended, there wouldn't be two actors, it would just be one, the original one.


> the address of the dead node will still be returned on other nodes 

This should be possible to fix, we know what nodes are around in the member list. so any PID belonging to a node that is no longer in the member list is incorrect.

Potentially we could scan the pid cache whenever a node leaves

alexeyzimarev:
> If everything worked as intended

But how other nodes know where the right actor is if they have no history (just started) and the partitioning algorithm tells them that the actor lives somewhere else?

> Potentially we could scan the pid cache whenever a node leaves

I think my code solves it with the overhead of stopping the actors. The idea is to adjust the landscape according to the partitioning algorithm when any change in cluster members list occurs. By doing that we stop relying on the delivery of termination messages to remote watchers and ownership transfer altogether.

rogeralsing:
Some more details:

## Scenario 1 - Handover + Join

//this is currently broken, but the theory is:
//no history is lost here, we just move IDs around

Node 1 is the owner of ID, actor lives on Node 2
Node 3 joins, now algorithmic owner of ID. 
Node 1 detects join event, calculate which of its own ID's should be transfered. clears those ID from PID cache and its own partition.
Node 3 detects handover event, adds IDs to partition

All done


## Scenario 2 - Cached PIDs + Join

//no history is lost, we just remove invalid entries from the PID cache

Node 1 has a cached address of ID and is not owner of ID, actor lives on Node 2
Node 3 joins, now algorithmic owner of ID. 
Node 1 still points to the correct address.
(Node X, the current algorithmic owner of ID hands over ownership according to scenario 1)

All done

## Scenario 3 - Cached PIDs + Leave

//no history is lost, we just remove invalid entries from the PID cache

Node 1 has a cached address of ID, actor lives on Node 2
Node 2 leaves
Node 1 detects leave event, purges node2 addresses from PID cache

All done

## Scenario 4 Handover + Leave

//we need to solve this one

Node 1 is the owner of ID, actor lives on Node 2
Node 1 leaves, crashes, and thus cannot hand over anything.
Node 2 now have an orphaned actor

Now what? 
Now there are actors living on other nodes, whos IDs were owned by node 1. they are now orphaned.

Solution?

rogeralsing:
One thing that should be done here, is to make the cluster infra more testable.
So the above scenarios can be described as code, w/o any actors or any moving parts.

Just passing events back and forth.
Ill get started on this this weekend.

rogeralsing:
In Scenario 1.

If another node, say node 4, asks for ID when Node 3 joins, but before Node 1 have handed over.
Node 3 will consider the ID missing in its partition and spawn a new actor.

Leaving us with an orphaned actor on node 2.

This is fine imo, it is a best-effort approach, and in the majority of cases everything will be ok.

TLDR; we need a sane approach to deal with orphaned actors, and should strive to limit them as much as possible.
They are however a non escapable issue from the virtual actor model, always available.

alexeyzimarev:
Ok, I the fact that ID owner and the node where the actor lives could be different was unclear to me. Doesn't it complicate things? What if the ID owner is up but the node is down, unfracefully?

rogeralsing:
>What if the ID owner is up but the node is down, unfracefully?

Then the ID owner should have detected the leave event (as detected by consul/etcd) and purged that ID from the partition.

the general idea is that, it is expensive to move actors around, that takes time, startup, deserialize/materialize state.
moving only ID's around is cheaper.

alexeyzimarev:
But it stays in PidCache. That seems to be the main issue.

Doesn't it more cohesive to say that the ID owner and the actor host should be the same node? It's a much simpler model for me to digest :)

I see the point of startup/shutdown burden, yes. However, only moving ids won't rebalance the cluster equally, so actors that were already there will live on the hosted node until they eventually die (could be never). In that sense, adding a node to the cluster where the set of actors is stable won't really scale up the cluster since the new node would be underloaded and the existing nodes will be overloaded.

rogeralsing:
>In that sense, adding a node to the cluster where the set of actors is stable won't really scale up the cluster since the new node would be underloaded and the existing nodes will be overloaded.

This would be equally true if the actor and ID lived on the same node. 
The new node is empty, the older nodes have actors


>Doesn't it more cohesive to say that the ID owner and the actor host should be the same node? It's a much simpler model for me to digest :)

Easier, but consider moving 10 000 event sourced actors vs moving 10 000 ID's.
We don't want a stop the world case when a new node joins.

alexeyzimarev:
Not exactly. The code from my PR loops through all actors known to the partition and finds the new address. If the new address doesn't match the existing address, the record is removed. Actors that changed the address and are currently on this node are stopped. It happens for join and leave events.

alexeyzimarev:
> We don't want a stop the world case when a new node joins.

It's a legit point. I need to think about it.

alexeyzimarev:
Maybe I overthought it. Now it seems to me that cleaning up the PidCache would be enough.

rogeralsing:
Also, we should make all this pluggable.

Instead of this messy distributed hashtable for ID's
Just storing each and every ID in a fast key value-DB, redis, memcache or whatever, and keep the pid cache would likely be easier and safer.

asking for an ID, check PID cache, if not there, ask that KVP DB. add to cache.
Done...

rogeralsing:
cc @cpx86 we need your brains here too :)

alexeyzimarev:
Yeah, keeping actor addresses would greatly simplify things. We won't need to ask the remote partition actor for anything, we'd know where it is. The only question is who will update the registry when one member leaves. We cannot trust the leaving member for doing that since it might simply crash.

rogeralsing:
Consul allows us to take a lock on a key. that could be used for a naive form of leader election.
the node that owns the lock, is considered leader.

Not sure what features are in ETCD or other stores, but generally there should be some form of Compare And Swap feature I guess.

alexeyzimarev:
I know nothing about etcd, the k8s provider that I built uses the label query. It should be possible to assign a special label on the leader pod too. But when the leader suddenly leaves, the remaining nodes need to coordinate the election.

alexeyzimarev:
Maybe something can be used for uncoordinated election, like the highest hash of some id.

rogeralsing:
>Maybe something can be used for uncoordinated election, like the highest hash of some id.

Exactly. or all nodes just competes for a databaselock / update or something like that.

Should be totally solvable

rogeralsing:
Outdated and replaced with cluster rewrite

PR #536, opened 2020-08-31
Mogno identity

rogeralsing:
Full cluster rewrite


src/Proto.Cluster/Cluster.cs
```diff
@@ -72,48 +74,77 @@ public async Task Start(ClusterConfig config)
...
+            _logger.LogInformation("Started");
         }
 
-        public async Task Shutdown(bool graceful = true)
+        public async Task ShutdownAsync(bool graceful = true)
         {
-            Logger.LogInformation("[Cluster] Stopping...");
+            _logger.LogInformation("Stopping");
             if (graceful)
             {
```
Damian-P, line 95: I think that if the pid cache is not used when It tries to shut it down, it fails because the pidcacheupdater actor doesn't exist.

`if (Config.UsePidCache)
                    PidCacheUpdater!.Shutdown();`

rogeralsing:
What am I not seeing here?
Is there an issue in the ClusterExperiment1 project file?

```
/home/runner/.nuget/packages/grpc.tools/2.31.0/build/_protobuf/Google.Protobuf.Tools.targets(64,5):
 error MSB4062: The "Grpc.Tools.ProtoToolsPlatform" task could not be loaded from the assembly
 /home/runner/.nuget/packages/grpc.tools/2.31.0/build/_protobuf/netstandard1.3/Protobuf.MSBuild.dll. Assembly 
with same name is already loaded Confirm that the <UsingTask> declaration is correct, that the assembly and all its 
dependencies are available, and that the task contains a public class that implements 
Microsoft.Build.Framework.ITask. [/home/runner/work/protoactor-dotnet/protoactor-dotnet/src/cluster-
playground/ClusterExperiment1/ClusterExperiment1.csproj]
```

src/Proto.Cluster/Cluster.cs
rogeralsing: Good catch!

Issue #411, opened 2018-06-14
Proto.Remote refactoring


rogeralsing:
Currently, the structure for Proto.Remote looks something like this:

![protoactor](https://user-images.githubusercontent.com/647031/41424298-e840ba88-6ffd-11e8-9f7b-d76df20ae82f.png)

This could probably be simplified a lot to something like:

![protoactor-page-2](https://user-images.githubusercontent.com/647031/41424306-eda903c2-6ffd-11e8-88c5-f934664c5920.png)

cpx86:
@rogeralsing Would you be up for kicking this off? :) I started trying to integrate Proto.Remote with the AspNetCore pipeline again and boy is it fast becoming a mess. It works better now since we introduced the ActorSystem and can now have separate instances of Remote, but the design of that class doesn't feel fitting at all anymore. 

Some issues:
- The Start method handles both wiring up the internal processes, and starting the standalone gRPC server
- The Start method will override the EndpointManager and EndpointReader on each invocation.
- The Serialization class needs to be instantiated separately from the Remote class, which feels strange and unnecessary

In addition to the changes you've already proposed (which should make further refactoring a lot easier thanks to simplification), I'm thinking it would be nice to find a way to separate the wiring of Remote from the actual server hosting. Maybe something like this (still figuring out the proper mechanics for the AspNetCore pipeline...):

```csharp
// from Startup.Configure()
app.UseProtoRemote(remote => {
    remote.RegisterMessages(FooReflection.Descriptor);
    remote.RegisterKnownKind("foo-actor", fooProps);
});
app.UseEndpoints(endpoints => {
    endpoints.MapProtoRemoteService();
});
```

or

```csharp
// from Main method or wherever
var server = new SelfHostedRemoteServer("127.0.0.1", 31337, remote => {
    remote.RegisterMessages(FooReflection.Descriptor);
    remote.RegisterKnownKind("foo-actor", fooProps);    
});
server.Start();
```

Thoughts?

rogeralsing:
I think we need to start one level up, how do we init the actorsystem iteself in the asp.net core pipeline?

register it as a singleton on the di container.
do we need some form of config support on the system?

cpx86:
ActorSystem would need to be registered in the DI yes. Possibly the Remote as well, so it can be gracefully shut down.

[...]

PR #537, opened 2020-09-03
AspNet Remoting & Plugins

Damian-P:
This is my proposition for the remote refactoring : #411

We can choose between Asp.Net or the native Grpc.

With asp.net we can run a cluster on a raspberry pi now using dotnet core. :-)

There is also a fix in the PartitionPlacementActor. Actually when a grain is stopped and we ask a new activation, it is redeployed elsewhere. But when it is redeployed on the same node, the PartitionPlacementActor returns the previous address as it is never removed from the _myActors dictionary and never spawn the new actor. When an actor is terminated I remove this pid from _myActors to fix that.


src/Proto.Cluster.Extensions/HostedClusterService.cs
```diff
@@ -0,0 +1,50 @@
+﻿// -----------------------------------------------------------------------
```
alexeyzimarev, line 1: It's already in contrib for a couple of months https://github.com/AsynkronIT/protoactor-dotnet-contrib/pull/2/files#diff-05d243082ea68e8c49700f5bc7fc94d2

rogeralsing:
@Damian-P this looks super interesting!
One request from our side, could you maybe split this up in smaller parts?
e.g. 
placement actor bug fix
plugin support
etc.

Because 109 files in one go is a bit too heavy for my personal schedule right now.
Many thanks!

alexeyzimarev:
Playground projects could be unloaded for the period of refactoring and we fix them in the next round.

PR #538, opened 2020-09-04
PartitionPlacementActor-fix

Damian-P:
Actually when a grain is stopped and we ask a new activation, it is redeployed elsewhere. But when it is redeployed on the same node, the PartitionPlacementActor returns the previous address as it is never removed from the _myActors dictionary and never spawn the new actor. When an actor is terminated I remove this pid from _myActors to fix that.


rogeralsing:
Good catch!

PR #539, opened 2020-09-04
Proto Plugins

Damian-P:
Adds plugin support to the actor system

### DIFF
--- a/src/Proto.Actor/ActorSystem.cs
+++ b/src/Proto.Actor/ActorSystem.cs
@@ -15,6 +15,7 @@ namespace Proto
             DeadLetter = new DeadLetterProcess(this);
             Guardians = new Guardians(this);
             EventStream = new EventStream();
+            Plugins = new Plugins();
             var eventStreamProcess = new EventStreamProcess(this);
             ProcessRegistry.TryAdd("eventstream", eventStreamProcess);
         }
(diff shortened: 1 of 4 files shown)


rogeralsing:
And the plugin system is basically a state store attached to the ActorSystem, so that other parts of the system can ask for a specific plugin?

alexeyzimarev:
Would be nice to have some /// comments there. Otherwise, it's hard to understand what is it.

Damian-P:
Yes it's right, it's state store.

We can do extensions like this :

```csharp
        public static void StartRemote(this ActorSystem actorSystem)
        {
            var remote = actorSystem.Plugins.GetPlugin<IRemote>();
            remote.Start();
        }
        public static IRemote GetRemote(this ActorSystem actorSystem)
        {
            var remote = actorSystem.Plugins.GetPlugin<IRemote>();
            return remote;
        }

```
https://github.com/AsynkronIT/protoactor-dotnet/issues/411#issuecomment-601091505

Damian-P:
We can also use the Microsoft Dependency injection layer instead.

alexeyzimarev:
I am wondering if it is maybe too weak of abstraction? Also, it the actor system who depends on Remote or the other way around?

rogeralsing:
I'm on the fence with this one.
I get the idea, Akka has the Extended Actor System concept which is basically the same.

One of the things I personally have felt pretty good about with Proto.Actor is that it has a simple DI based design.

Remote uses System
Cluster uses Remote

There are some back and forth things in the cluster infra though.

Having registered plugins might in the end lead to design choices where one plugin asks for another.
Another consideration, I'd like to be able to keep the structure at least somewhat similar in the Go impl and other langs.

PR #540, opened 2020-09-04
Relocate the Address on the ActorSystem class

Damian-P:


rogeralsing:
Looks good to me!
cc @alexeyzimarev 

(IMO there is no completely natural fit where the address should go, but the system is better than the process registry.)

alexeyzimarev:
I am wondering if we can improve it as well by making `Address` a value object and pass it by reference. Now, every request to `Address` and `GetAddress` lead to an allocation although the data is stale.

rogeralsing:
I'll merge this. the address is not a critical path anyway. 
Many thanks @Damian-P !

PR #546, opened 2020-09-23
Cancellation token in actor context

Damian-P:
## Description

It would be nice to be able to stop an actor right away even if it is processing a long running task. Today we need to wait that the actor finishes processing the current message.
This PR enables it by passing a CancellationToken to the actor's context.

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/.github/workflows/build-master.yml
+++ b/.github/workflows/build-master.yml
@@ -22,9 +22,7 @@ jobs:
 
       - name: Run tests
         run: |
-          docker-compose -f tests/docker-compose.yml up -d
           dotnet test -c Release
-          docker-compose -f tests/docker-compose.yml down
         
   nuget:
     runs-on: windows-latest
--- a/src/Proto.Actor/ActorContext.cs
+++ b/src/Proto.Actor/ActorContext.cs
@@ -43,8 +43,8 @@ namespace Proto
         private static ILogger Logger { get; } = Log.CreateLogger<ActorContext>();
 
         public ActorSystem System { get; }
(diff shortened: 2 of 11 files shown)


rogeralsing:
Hello @Damian-P, I _really_ like this idea. 
It makes perfect sense to me.

My only objection to it:
https://github.com/AsynkronIT/protoactor-dotnet/pull/546/files#diff-7f6ae40b07d28b06255dc176688be953R46


This would lead to an allocation of a TaskCompletionSource for each and every actor, no matter if it is being used.
It also adds a few extra bytes to the ActorContext size in terms of the reference to it. which also affects mem profile and allocation speed of the actor context.
As Proto.Actor is extremely optimized for having small allocations on the default path.
e.g. this is shown by SpawnBenchmark which outperforms other actor implementations in .NET space.

I would rather see that this lives in the `ActorContextExtras.cs` where any additional, non default state lives.
Then having the IContext.CancellationToken lazily map to that.
e.g.
by default, the actor does not have a ActorContextExtras instance, it is only allocated on demand.

ActorContex could have a property definition like so:

```
class ActorContext : IContext .... {

   //EnsureExtras allocates the ActorContextExtras if needed.
   CancellationToken => EnsureExtras().CancellationToken;
```

Do you think you could add this adaptation to it?

Damian-P:
You're absolutely right. I moved that to ActorContextExtras.cs.

rogeralsing:
Many thanks!
This is great work

PR #553, opened 2020-10-04
WiP Config Cleanup

rogeralsing:
WiP PR on unifying the configurations of all modules.

Some reasoning behind this, the xxxConfig constructors accept all the arguments that has to be provided.
Everything else, can be set either directly as a property, or with the Withxxx fluent set methods.

This way, we still keep it explicit what has to be provided, and make it easy to provide the optionals.

Examples:

```csharp
var remoteConfig = new RemoteConfig(bindingHost, bindingPort)
        .WithAdvertisedHostname("192.168.0.23")
        .WithProtoMessages(MessagesReflection.Descriptor);

var clusterConfig = new ClusterConfig("mycluster", host, p, clusterProvider)
        .WithIdentityLookup(identity)
        .WithRemoteConfig(remoteConfig);

var cluster = new Cluster(system, clusterConfig)
```


rogeralsing:
Breaking changes in this one.

e.g.
KnownKinds from Remote, have been separated into Remote.RemoteKinds and Cluster.ClusterKinds.
The ClusterKinds are the ones that goes into the ClusterProvider.

The reason for this, is that even though they are pretty much identical, the parent actor, that owns and inits the kinds, are different between Remote and Cluster.

In Cluster mode, the IdentityLookup provider will have its own Placement actor, that owns the cluster spawned actors.
The cluster spawned actors also get initialized with a GrainInit message. 

Same concept, different implementations.

rogeralsing:
Another thing that needs to be settled, should Remote and Cluster configs be immutable, where the WithXXX returns a new copy.

This is how Props works.  so I think it would make sense. but needs some added code for this

PR #557, opened 2020-10-19
net5.0 upgrade

RagingKore:
## Description

Upgraded everything to `netcoreapp3.1` and `net5.0` without any breaking changes.

Also upgraded a lot of old nuget packages.


rogeralsing:
Out of sync and .NET 5.0 upgrade merged already, closing

PR #559, opened 2020-10-26
Local affinity processing

rogeralsing:


src/Proto.Cluster/ClusterHeartBeat.cs
```diff
@@ -36,7 +46,7 @@ public ClusterHeartBeat(Cluster cluster)
 
         public Task StartAsync()
         {
-            var props = Props.FromProducer(() => new ClusterHeartBeatActor());
+            var props = Props.FromProducer(() => new ClusterHeartBeatActor(_system));
```
rogeralsing: Do we need _system here? `context.System` exists once we are in the receive method

src/Proto.Cluster/Member/LocalAffinityPropsExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// Adds middleware which causes remote traffic against the actor to potentially poison it.
+        /// Useful with local affinity strategy to move partitioned workloads to the right node, even after a re-balance
+        /// </summary>
+        /// <param name="props"></param>
+        /// <param name="relocationFactor">Chance the actor is poisoned on remote traffic, 0-1</param>
+        /// <param name="hasLocalAffinity">Predicate on message envelope, to have local affinity only on partitioned messages</param>
+        /// <returns></returns>
+        /// <exception cref="ArgumentException"></exception>
+        public static Props WithPoisonOnRemoteTraffic(this Props props,
+            float relocationFactor = DefaultRelocationFactor, Predicate<MessageEnvelope>? hasLocalAffinity = null)
```
rogeralsing: ooh, this is very nice, per message affinity checks. 👍🏻

rogeralsing:
This looks great!

src/Proto.Cluster/ClusterHeartBeat.cs
mhelleborg: Cleaned up :)

src/Proto.Cluster/Member/LocalAffinityPropsExtensions.cs
mhelleborg: Nice to have if we have some low volume traffic from elsewhere :)

PR #581, opened 2020-11-11
Plugin poc

rogeralsing:
A continuation of @Damian-P's plugin PR.

This avoids using any dictionary lookups and instead goes for a fast path array lookup.
Same concept otherwise


src/Proto.Actor/ActorSystem.cs
```diff
@@ -25,6 +27,7 @@ public ActorSystem(ActorSystemConfig config)
             EventStream = new EventStream(config.DeadLetterThrottleInterval, config.DeadLetterThrottleCount);
             var eventStreamProcess = new EventStreamProcess(this);
             ProcessRegistry.TryAdd("eventstream", eventStreamProcess);
+            Extensions = new ActorSystemExtensions(this);
```
rogeralsing, line 30: Initialize the extensions instance for the system

src/Proto.Actor/Extensions.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace Proto
 {
-    public static class Extensions
+    public static class UtilExtensions
```
rogeralsing, line 12: Avoid name clash for these extension methods

src/Proto.Actor/Extensions/ActorSystemExtension.cs
```diff
@@ -0,0 +1,14 @@
+using System.Threading;
+
+namespace Proto.Extensions
+{
+    public static class ActorSystemExtension
+    {
+        private static int _nextId;
```
rogeralsing: counter for the unique plugin ID's

src/Proto.Actor/Extensions/ActorSystemExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+        private readonly ActorSystem _actorSystem;
+
+        public ActorSystemExtensions(ActorSystem actorSystem)
+        {
+            _actorSystem = actorSystem;
+        }
+
+        public T GetExtension<T>(int extensionId) where T: IActorSystemExtension
+        {
+            return (T)_extensions[extensionId];
```
rogeralsing: Get an extension by type and ID. a bit ugly to require both, but perf is important

src/Proto.Actor/Extensions/IActorSystemExtension.cs
```diff
@@ -0,0 +1,7 @@
+namespace Proto.Extensions
+{
+    public interface IActorSystemExtension
```
rogeralsing, line 5: Marker interface for plugins so you don't register whatever as a plugin

src/Proto.Cluster/Cluster.cs
```diff
@@ -22,6 +22,8 @@ public class Cluster
 
         public Cluster(ActorSystem system, ClusterConfig config)
         {
+            system.RegisterClusterPlugin(this);
```
rogeralsing: create a cluster plugin that knows about this cluster, register this plugin for the system

src/Proto.Cluster/ClusterPlugin.cs
```diff
@@ -0,0 +1,32 @@
+using Proto.Extensions;
+
+namespace Proto.Cluster
+{
+    public class ClusterPlugin : IActorSystemExtension
```
rogeralsing: The cluster plugin, this could basically be the cluster itself too if we want

src/Proto.Cluster/ClusterPlugin.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        public static readonly int PluginId = ActorSystemExtension.GetNextId();
+
+        public static void RegisterClusterPlugin(this ActorSystem system, Cluster cluster)
+        {
+            var plugin = new ClusterPlugin(cluster);
+            system.Extensions.RegisterExtension(PluginId, plugin);
+        }
+        
+        public static Cluster Cluster(this IContext self)
```
rogeralsing: Get the cluster instance from within any actor

src/Proto.Cluster/ClusterPlugin.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            Cluster = cluster;
+        }
+    }
+
+    public static class ClusterPluginExtensions
+    {
+        public static readonly int PluginId = ActorSystemExtension.GetNextId();
+
+        public static void RegisterClusterPlugin(this ActorSystem system, Cluster cluster)
```
rogeralsing: register the cluster plugin for the actor system

src/Proto.Actor/DependencyInjection/DIExtension.cs
```diff
@@ -0,0 +1,50 @@
...
+        //
+        // public Func<IActor> CreateActorFactory(Type actorType)
+        // {
+        //     throw new NotImplementedException();
+        // }
+
+        public Props PropsFor<TActor>() where TActor : IActor => Proto.Props.FromProducer(() => _services.GetService<TActor>()!);
+
+        public Props PropsFor(Type actorType) => Proto.Props.FromProducer(() => (IActor)_services.GetService(actorType)!);
+
```
mhelleborg: In addition to the types, perhaps we should add props by string?

src/Proto.Actor/Extensions/ActorSystemExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            var id = IActorSystemExtension<T>.Id;
+            return (T)_extensions[id];
+        }
+
+        public void Register<T>(IActorSystemExtension<T> extension) where T:IActorSystemExtension
+        {
+            var id = IActorSystemExtension<T>.Id;
+            //TODO, ensure capacity
+            _extensions[id] = extension;
```
mhelleborg: I don't think you need to optimize here when adding extensions. Just replace the _extensions array with a bigger one with all the old ones copied over. Then there is no set limit to how many plugins one can add, and you can use a smaller default size..

mhelleborg: Oh, and ofc lock when switching out the array :)

src/Proto.Actor/DependencyInjection/DIExtension.cs
mhelleborg: Oh, `IDependencyResolver` does not support it..

src/Proto.Actor/Extensions/ActorSystemExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        public void Register<T>(IActorSystemExtension<T> extension) where T:IActorSystemExtension
+        {
+            lock (_lockObject)
+            {
+                var id = IActorSystemExtension<T>.Id;
+                if (id >= _extensions.Length)
+                {
+                    var newSize = id * 2; //double size when growing
+                    Array.Resize(ref _extensions, newSize);
```
rogeralsing, line 30: lock and resize

PR #582, opened 2020-11-12
Rewritten external identity storage, added support for Redis

rogeralsing:


benchmarks/ClusterBenchmark/Program.cs
```diff
@@ -159,10 +160,11 @@ private static IClusterProvider ClusterProvider()
             }
         }
 
-        private static MongoIdentityLookup GetIdentityLookup()
+        private static IIdentityLookup GetIdentityLookup()
         {
             var db = GetMongo();
-            var identity = new MongoIdentityLookup("mycluster", db);
+            var identity = new IdentityStorageLookup(
+                new MongoIdentityStorage("mycluster", db.GetCollection<PidLookupEntity>("pids")));
```
rogeralsing, line 167: Should we make "pids" configurable?

mhelleborg: This is configurable, since it is passed in by the user

rogeralsing: doh, it's in the consumer code, my bad

rogeralsing:
Looks rock solid, I can't find anything that needs changing. 👍🏻

mhelleborg:
I have one more change I want to add, but it requires changing the IIdentityLookup interface. When retrieving a PID from IdentityLookup and getting a dead letter response, we should purge it..

mhelleborg:
@rogeralsing Should we update the IdentityLookup interface to support this?

rogeralsing:
Sure lets go for that

PR #585, opened 2020-11-13
Refactoring remote layer & GrpcNet

Damian-P:
As discussed in #411 I refactored the remote layer and abstracted the Remote class to give us the ability to choose between GrpcCore and GrpcNet.

The Cluster layer uses the extensions to access the remote layer.
Before creating the cluster instance with this PR, we need to add the Remote plugin to an actor system.


src/Proto.Remote.GrpcCore/RemoteConfig.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Proto.Remote
+{
+    public record RemoteConfig : RemoteConfigBase
+    {
+        protected RemoteConfig(string host, int port) : base(host, port)
+        {
+        }
+        
+        public static RemoteConfig BinToAllInterfaces(string advertisedHost, int port = 0) =>
```
rogeralsing: Old typo, BinTo -> BindTo

src/Proto.Remote.GrpcNet/RemoteConfig.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Proto.Remote
+{
+    public record RemoteConfig : RemoteConfigBase
+    {
+        protected RemoteConfig(string host, int port) : base(host, port)
+        {
+
+        }
+        public static RemoteConfig BinToAllInterfaces(string advertisedHost, int port = 0) =>
```
rogeralsing: Same typo, BinTo -> BindTo

rogeralsing:
This is great 🎉 

Some questions,
If someone uses the Grpc.NET version, and does so from an Asp.NET core app, is there a separate startup in that case or can it still safely do all the bootstrapping of configuring kestrel etc seen here? https://github.com/AsynkronIT/protoactor-dotnet/pull/585/files#diff-08dae864959e2426048096def86528a42ebf4988ddaf15e0b62096e79034d520R33

Have you done any benchmarks, e.g. RemoteBenchmarks comparing the two versions?

Damian-P:
It's the same startup.
You will find 3 different Remote classes:
- Remote on the GrpcCore side
- Remote on the GrpcNet side
- HostedRemote wich is used in the RemoteHostedService, so it's started with the AspNet app

For example this is a regular AspNet app :
```C#
// This method gets called by the runtime. Use this method to add services to the container.
// For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
public void ConfigureServices(IServiceCollection services)
{
    services.AddLogging(loggingBuilder =>
    {
        loggingBuilder.AddSeq(Configuration.GetSection("Seq"));
    });
    services.AddGrpc(options =>
    {
        options.MaxReceiveMessageSize = null;
        options.EnableDetailedErrors = true;
    });
    services.AddSingleton(sp => new ActorSystem().WithServiceProvider(sp));
    services.AddRemote((sp) =>
    {
        RemoteConfig remoteConfig;
        if (HostingEnvironment.IsDevelopment())
            remoteConfig = RemoteConfig.BindToLocalhost(80);
        else
            remoteConfig = RemoteConfig.BinToAllInterfaces(Environment.MachineName, 80);
        remoteConfig = remoteConfig.WithProtoMessages(Messages.ProtosReflection.Descriptor);
        remoteConfig = remoteConfig.WithRemoteKind("HelloActor", Props.FromProducer(() => ActivatorUtilities.GetServiceOrCreateInstance<HelloActor>(sp)));
        return remoteConfig;
    });
    ConsulProvider clusterProvider = new ConsulProvider(new ConsulProviderConfig(), c =>
    {
        c.Address = new Uri($"http://{Configuration.GetValue("ConsulHostname", "127.0.0.1")}:8500");
    });
    services.AddClustering((remoteConfig, sp) =>
        ClusterConfig.Setup("StabilityTestAsp", clusterProvider, new PartitionIdentityLookup(), remoteConfig)
        .WithClusterKind<HelloActor>(sp, "HelloActor")
        .RegisterHelloGrainKind(sp)
    );
    // services.AddGrains<Grains>((grains, sp) => grains.RegisterHelloGrainFactory(() => sp.GetRequiredService<IHelloGrain>()));
    services.AddTransient<IHelloGrain, HelloGrain>();
}

// This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILoggerFactory loggerFactory)
{
    Log.SetLoggerFactory(loggerFactory);
    if (env.IsDevelopment())
    {
        app.UseDeveloperExceptionPage();
    }

    app.UseRouting();

    app.UseEndpoints(endpoints =>
        {
            endpoints.MapGet("/",
                async context =>
                {
                    await context.Response.WriteAsync(
                        "Communication with gRPC endpoints must be made through a gRPC client. To learn how to create a client, visit: https://go.microsoft.com/fwlink/?linkid=2086909"
                    );
                }
            );
        }
    );
    app.UseProtoRemote();
}
```

src/Proto.Remote/RemoteProcess.cs
```diff
@@ -41,7 +39,7 @@ private void Send(object msg)
 
         private void SendMessage(object msg)
         {
-            _remote.SendMessage(_pid, msg, -1);
+            _endpointManager.SendMessage(_pid, msg, -1);
```
rogeralsing, line 42: Looking at this, makes me wonder if we somehow safely can bypass the `EnsureConnected` part in endpoint manager SendMessage/RemoteDeliver.
if that check is done once and then return some stateful object back to the RemoteProcess, which then holds on to that.
We could likely improve remote throughput by avoiding the dictionary lookup for address for each message.

Not a critical fix for this PR, but still worth thinking about.

Damian-P:
There are some performance benefits but nothing huge. I'm able to send over 2M messages /s in the Remote benchmark.
It looks like dotnet needs to warm up before getting such numbers. 

But the main benefit is to be able to run Proto.Actor anywhere (arm devices for example).

The GrpcNet backend can be easily wired to some Tracing tools like OpenTracing.

Damian-P:
Other thing enabled is this PR is the graceful shutdown. Every request has to be processed for that.
I have some code to enable this in the GO implementation as well.

src/Proto.Remote.GrpcCore/Remote.cs
```diff
@@ -0,0 +1,113 @@
...
+                        System.Address, ex.Message
+                    );
+                    _server.KillAsync().GetAwaiter().GetResult();
+                }
+                Started = false;
+                return Task.CompletedTask;
+            }
+        }
+
+        public void SendMessage(PID pid, object msg, int serializerId)
```
Damian-P: Is it only used in unit tests ?

rogeralsing: Might be after your changes, in the dev branch its used from a few more places

Skärmavbild 2020-11-13 kl  19 40 19

src/Proto.Remote.GrpcCore/Remote.cs
```diff
@@ -0,0 +1,113 @@
...
+using Grpc.Core;
+using Grpc.Health.V1;
+using Grpc.HealthCheck;
+using JetBrains.Annotations;
+using Microsoft.Extensions.Logging;
+
+namespace Proto.Remote
+{
+    [PublicAPI]
+    public class Remote : IRemote
```
mhelleborg: Namespace / class cannot clash, this will collide with the GrpcNet implementation..

rogeralsing:
Some thoughts. 
The effort to make this backwards compatible is impressive, but I do think there is an inconsistency compared to how other parts of PA works.
e.g. Persistence has providers. Cluster has providers.

One could argue that the IRemote also are providers, but this is my current gut feeling, that maybe the Remote object should be the same, and injected with e.g a TransportProvider or something like that.
I have no solid arguments for it rather than the consistency of how the API is shaped.

src/Proto.Remote.GrpcCore/Remote.cs
Damian-P: The idea was to be backward compatible. The client project will only have to reference one Implementation.
But I can change that to be SelfHostedGrpcCoreRemote and SelfHostedGrpcNetRemote or change the namespaces to be Proto.Remote.GrpcNet and Proto.Remote.GrpcCore.

mhelleborg: Doing it with the namespace could be enough I think. Perhaps GrpcNetRemote / GrpcCoreRemote for the classes?

mhelleborg:
One thing I am missing here is a test suite. Ideally we can abstract out a common spec which both providers can be run against.

Have a look at the way it is done for both cluster and identity, where you can run the same tests against multiple implementations. A separate fixture can be provided for each Remote type.

src/Proto.Remote.GrpcCore/GrpcCoreRemote.cs
```diff
@@ -0,0 +1,113 @@
...
+            lock (this)
+            {
+                if (!Started)
+                    return Task.CompletedTask;
+                try
+                {
+                    if (graceful)
+                    {
+                        _endpointManager.Stop();
+                        _server.ShutdownAsync().GetAwaiter().GetResult();
```
mhelleborg: This already returns a task, why not use await?

src/Proto.Remote/EndpointActor.cs
```diff
@@ -0,0 +1,299 @@
...
+        {
+            Logger.LogDebug("[EndpointActor] Connecting to address {Address}", _address);
+            try
+            {
+                _channel = _channelProvider.GetChannel(_address);
+            }
+            catch (Exception e)
+            {
+                Logger.LogError(e, "[EndpointActor] Error connecting to {_address}.", _address);
+                throw e;
```
mhelleborg: This changes the stacktrace, throw without e to keep it..

src/Proto.Remote.GrpcCore/GrpcCoreRemote.cs
Damian-P: We can't await in a lock block. I can to use a semaphore instead.

mhelleborg: You could go with a semaphore then. But if we never re-use the remote after shutdown, you could just set a flag that it is already shut down in a lock..
The rest could be done after the lock block?

PR #588, opened 2020-11-14
PoC logging

rogeralsing:
PoC for logging extension.

It contains a temp hack capturing the current static Log factory and copies that into the new LogExtension.

The LogExtension is then accessible from anywhere you have access to the ActorSystem.

**Some things to consider,** I really don't want to create a new logger per `ActorContext` as that would be very expensive.
Maybe this can be solved with some form of shared ActorContext logger.

There are some examples showing how this new feature is used for Remote, Cluster and MemberList.

Thoughts?

This could all also be done w/o any form of extensions, just having a LoggerFactory prop on the actor system, so maybe this PR overcomplicates things here.

### DIFF
--- a/src/Proto.Actor/ActorSystem.cs
+++ b/src/Proto.Actor/ActorSystem.cs
@@ -2,6 +2,7 @@ using System;
 using Google.Protobuf;
 using JetBrains.Annotations;
 using Proto.Extensions;
+using Proto.Logging;
 
 namespace Proto
 {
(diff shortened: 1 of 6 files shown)


alexeyzimarev:
When getting the `ActorSystem` instance in the producer function, each actor can create its own logger using the extension. But in reality, I would use the static Serilog logger in each actor type, not per actor instance as it's completely unnecessary. That's the weakness of the Microsoft logging abstraction tbh.

rogeralsing:
Going down the DI route instead, we could easily have a singleton Logger<T> open generics, which when fetched from the DI container yields a unique instance per T. e.g.  Logger<Deviceactor>

but, this would still eat cpu cycles resolving this. for each actor. I need to ponder this some more

PR #592, opened 2020-11-28
Some more Remote refactoring

Damian-P:
As @rogeralsing said in my last PR, the RemoteProcess can access the EndpointActor pid and then it sends directly to his EndpointActor.

The EndpointActor now sends directly to himself the various RemoteDeliver messages he's handling.*

I Added a RemoteTestFixture with the GoogleProtobuf Json serializer as default on the client node and another fixture with a custom serializer based on System.Text.Json on both nodes.

By removing some debug logs and configuring some tasks in the remote layer, the dotnet version is almost as fast as the go implementation.
Dotnet version over GrpcNet : 3 517 284 msg / sec
Dotnet version on GrpcCore : 3 249 613 msg / sec
Go version : 4 123 692 msg / sec

With the PR in the Go project I'm able to gracefully shutdown remotes between go and .net versions.


src/Proto.Remote/RemoteProcess.cs
```diff
@@ -23,35 +23,33 @@ public RemoteProcess(ActorSystem system, EndpointManager endpointManager, PID pi
 
         private void Send(object msg)
         {
+            object message;
+            var endpoint = _endpointManager.GetEndpoint(_pid.Address);
```
rogeralsing: Wouldn't it be possible to keep this `endpoint` reference around after the first lookup?
Meaning it only performs the dict lookup once.

It might have to be cleared out if the endpoint have been terminated and later recreated, but might be doable?

rogeralsing:
Awesome work. 👍🏻 
I think there might be a few more TPS to squeeze out if we cache the endpoint of the remote process.
each remote process only talks to one endpoint, so it should be doable I think.

>With the PR in the Go project I'm able to gracefully shutdown remotes between go and .net versions.

This is great! 🎉

src/Proto.Remote/RemoteProcess.cs
Damian-P: As you suggested, I'm now storing the endpoint pid in the remoteProcess. But it's done only if it's not null, keeping the current behavior when it tries to connect to a terminated endpoint.

To handle the case when the endpoint is terminated, I added a subscription to the DeadLetterEvent in the EndpointManager to handle the RemoteWatch and RemoteDeliver messages to translate them.

Issue #444, opened 2018-12-08
Client connection without callback port


sudsy:
I have been using Orleans server for several years now, but I was looking for an alternative and stumbled across this project.

I have been using Orleans in conjunction with thousands of AWS lambda processes to offload some computationally intensive workloads.

I have been looking at implementing Grpc with Orleans and having the lambda clients open a bidirectional streaming rpc with Orleans to allow messages to pass between them.

When I found your project I thought I might have found a better answer. I thought perhaps, remote actors were doing exactly that. Just opening a streaming connection to allow the rpc calls to happen.

Unfortunately (for my use case) it seems that the client is required to listen on a port to receive messages.

If I am wrong and there is a way to achieve bidirectional comms without a listening client port, please let me know.

If I am right, could we open a discussion about another way of interfacing with the server which uses the mechanism I have described above. I need to do a fair amount of work plugging gRPC to Orleans and it seems like it might be quicker to amend this project instead. I haven't had a deep dive into the architecture yet so perhaps you can comment on how plausible this approach might be.

potterdai:
@sudsy From my knowledge, the Cluster part of ProtoActor is indeed based on bidirectional message of GRPC. The client is needed in a similar way as Orleans, which is to access and send/receive messages from grains.

sudsy:
I had a look through the code and can see that there is some bidirectional communication going on and I can also see that Proto.Remote is used by Proto.Cluster to establish connectivity. It seems Proto.Remote needs a port open to communicate with other nodes. I can see in my network traces that is happening too.

I am going to put together a prototype of Proto.Client so I can use it on hosts that can't open an incoming port. Hopefully I will have something to discuss further soon.

sudsy:
@PotterDai I have put together a basic client that can connect without needing to open a tcp port for replies. The work I have done so far is in a branch called [proto_client in my fork](https://github.com/sudsy/protoactor-dotnet/tree/proto_client). All the work is kept in src/contrib. I haven't made a pull request at this stage because I wanted to discuss the approach before doing so.

My goals for this client were:

- Zero (or close to) impact on the current codebase and architecture
- Ability to support connection from AWS lambda processes (which can't open inbound ports) or other client processes
- A programming model that is natural to users of Proto.Actor
- The ability to register client actors to receive messages as well as Request Reply modes
- Support for 000's of client processes

[...]

PR #594, opened 2020-11-30
Allow client resolvers

sudsy:
## Description

I am expecting to have some time to work on the client project #444 again in the coming months and want to implement the changes in the core that will enable the work to progress.

In the end I decided that a client resolver collection would be the best way to deal with clients. This will permit clients to be registered with the actor system in the same way as remotes are registered.

Clients can be addressed with a scheme like:

 {
            address: "{clienthostaddress}",
            Id: "$client/{guid}/$2"
 }

The rest of the cluster does not need to know that it is communicating with a client.

## Purpose
This pull request is a:

## Checklist


rogeralsing:
Can you elaborate on this one?
When you say client, you mean another remote node?

For cluster we can spawn up nodes that can communicate with the cluster, while not joining themselves.
e.g.
Cluster.StartClientAsync() vs Cluster.SpawnMemberAsync()

Is this change part of making it possible to call back w/o opening a backwards connection to the calling client?

sudsy:
Yes It is part of making it possible to call back w/o opening a backwards connection to the calling client?

Perhaps we need different terminology here but for now I will use the following definitions.

ClientHost - an actor system that can accept client connections without a backwards connection to the calling client.

Client - a calling client that is connected to the ClientHost

The flow will go something like this:

- ClientHost is started just like a normal remote but it has an additional grpc service for accepting Client connections that is started on a different port (ClientHostProxy Port)

- Client connects to ClientHost Proxy Port, client is issued a guid and clientHost registers a clienthostresolver that will handle any messages addressed to:

{
address: "{clienthostaddress}",
Id: "$client/{guid}/*"
}

I have included * in the id above to indicate that it will handle anything that starts with $client/{guid}/

- The root context on the client is given the Id 

{
address: "{clienthostaddress}",
Id: "$client/{guid}/"
}

so that all actors on the client have an address of the form listed above

- When a client communicates with any other node, it always does so through this client proxy connection, with the ClientHost forwarding the messages to the destination.

- When other nodes send a message back to a client actor, they are unaware that they are sending to a client, the just send their messages to the clienthostaddress and the clienthostresolver collection takes care of mapping the ids to the correct client proxy connection.

rogeralsing:
Thanks for the explanation,
So:
```
address: "{clienthostaddress}",
Id: "$client/{guid}/"
```

the `$client/guid` part is used to match that this is should be looked up by the clienthost resolver, correct?

What if we remove that part, and instead have something in the address instead?
That way it should be possible to resolve this via the HostResolvers, just like the remoting layer does.

Address: "client://352353532@192.168.0.232"
Id: "MyActor"

something like that.

Also, the guid, why is that needed if the client has an address?
is it to disambiguate a restarted client that happens to be on the same host+port as a previous client?

sudsy:
IMHO an addressing scheme like Address: "client://352353532@192.168.0.232" looks more logical but I have some concerns about it. All nodes participating in the cluster would need to be aware of the addressing scheme. So, would need to have a host resolver on every single node to handle the resolution of the client:// scheme or some hardcoded awareness of the client:// scheme that re-routes the request to the appropriate client host. Each of the other implementations of protoactor would also need to be aware of the scheme too.

If we were going to make a change in all implementations, I would be in favour of a scheme like:

```
Address: "client://{guid}"
Id: "MyActor"
```
With the lookup of the client location to be handled by a directory (or maybe piggyback on some of the cluster directory functionality). That way, we could handle clients being disconnected and reconnecting to a different clienthost without changing the identity of the actors on the client.

While I am in favour of an approach like this, it's a bigger project than I had in mind for the coming months but I would be happy to work with someone on it.

An implementation like
```
address: "{clienthostaddress}",
Id: "$client/{guid}/"
```
is just the lightest touch implementation that I could come up with. The strength of this approach being that it is a simple implementation, the weakness of this approach is handling client's reconnecting to different clienthosts. (It could still be done, but it is a bit messy)

The reason for using a guid instead of something like 352353532 is so that the client can create it's own guid for an identity without having to request it's identity from the clienthost. The guid saves some startup time for the client (not mutch) but makes for bigger pids with a long guid instead of an integer identifier. I would be happy with either approach really.

rogeralsing:
I'm following now. I'm just dense.

```
{
address: "{clienthostaddress}", 
Id: "$client/{guid}/*"
}
```

In this case, the address is the address of the member node that received the client connection, correct?
So all other nodes only knows about this normal node, and any response to such pid would go to the the node that knows about the client, and this node would then send it back to the correct client.

This is actually brilliant!

rogeralsing:
I'm merging this in :-)

sudsy:
Yeah, that's correct. 

Thanks for the merge, glad to hear you like the approach.

PR #603, opened 2020-12-05
async throttler, dont block the threadpool

rogeralsing:
Small change to throttler, no hard blocking calling thread


mhelleborg:
👍

mhelleborg (approved):
Looks fine to me, SemaphoreSlim here seems like a good choice

PR #607, opened 2020-12-05
Rewritten IdentityStorageWorker to enable ordering without blocking on storage calls.

mhelleborg:


src/Proto.Cluster.Identity/IdentityStorageWorker.cs
```diff
@@ -32,51 +32,74 @@ public IdentityStorageWorker(IdentityStorageLookup storageLookup)
...
+                pidList.Add(context.Sender);
+                return Task.CompletedTask;
             }
 
+            _inProgress.Add(msg.ClusterIdentity, new List<PID>());
 
-            void Respond(PID? pid)
-            {
-                context.Respond(new PidResult
+            context.ReenterAfter(GetWithGlobalLock(context.Sender!, msg.ClusterIdentity, context.CancellationToken), task => {
```
rogeralsing, line 63: I like this 👍🏻 , let's consider if it can be made into some form of pattern/abstraction also.
IIRC we have(had?) a similar case in PartitionIdentityActor.

TaskCompleter<TKey,TValue,TSubscriber>  _taskCompleter = new();


_taskCompleter.Add (key,  onMissing: key => ...., onCompleted: (value, subscriberArr) => foreach(var s in subscriberArr) ..... )

something in that direction maybe?

mhelleborg: Not a bad idea. Perhaps we can make it more efficient in the base case as well, eliminating the list allocation?

PR #611, opened 2020-12-06
Revert Cluster Regressions

rogeralsing:
Emergency hot fix to revert proper behavior in public nugets.

After the latest optimizations of Remote and Identity modules, 
Something broke in terms of cluster stability.
Reverting both these projects makes everything work as expected again.


rogeralsing:
cc @mhelleborg  @Damian-P 

This is not intended as a permanent change, but we need to have nugets that work properly.
Continuing to debug what causes the incorrect behavior.

PR #617, opened 2020-12-09
use cluster request async with actor context

rogeralsing:
@mhelleborg 

PoC of using actor context as sender in Cluster.RequestAsync


mhelleborg (commented):
Excellent 🥇

PR #620, opened 2020-12-17
Masstransit counters PoC

rogeralsing:
PoC stealing the Performance Counter abstraction from @MassTransit.
Nothing implemented yet, just copy paste on counter infra code.

cc @alexeyzimarev @mhelleborg do you think this is something we can work on? adapting it to the PA infrastructure?

Some special cases, we have core, remote and cluster, (and persistence?) that could benefit from this.
each module would need its own counters.

cc @phatboyg for transparency

### DIFF
--- a/src/Proto.Actor/Monitoring/Performance/BasePerformanceCounters.cs
+++ b/src/Proto.Actor/Monitoring/Performance/BasePerformanceCounters.cs
@@ -1,0 +1,36 @@
+namespace MassTransit.Monitoring.Performance
+{
+    using System;
+    using System.Collections.Generic;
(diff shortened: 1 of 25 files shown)


rogeralsing:
`Counter`, is just the name and description/help of a counter.
Counter gets converted into a `CounterData`, which is the same + counter type. (why are those not the same?)
Then there is the actual `IPerformanceCounter` implementations, e.g Null and StatsD. 

Not sure how the relation is between the first two and the IPerformanceCounter.

My gut feeling is that this can be adapted, as on a conceptual level, we are also dealing with messages sent, failed messages etc.

PR #624, opened 2020-12-19
Global killswitch

rogeralsing:
Global Killswitch WiP.

This adds a CancellationTokenSource to the ActorSystem.
Whenever system.ShutdownAsync is called, the token is cancelled.
This way, other subsystems can check if the token is cancelled

There are a lot of moving parts, requests, heartbeats, endpoints etc that all should interact with this token.

It's a bit of trial and error on where to use the token and if it should just silently return, or throw in those cases.

at the very least, the shutdown process is clean now and will unregister whatever should be unregistered.

cc @Damian-P @mhelleborg @alexeyzimarev

### DIFF
--- a/src/Proto.Actor/ActorSystem.cs
+++ b/src/Proto.Actor/ActorSystem.cs
@@ -31,7 +31,7 @@ namespace Proto
             Root = new RootContext(this);
             DeadLetter = new DeadLetterProcess(this);
             Guardians = new Guardians(this);
-            EventStream = new EventStream(config.DeadLetterThrottleInterval, config.DeadLetterThrottleCount, Token);
+            EventStream = new EventStream(config.DeadLetterThrottleInterval, config.DeadLetterThrottleCount, Shutdown);
             var eventStreamProcess = new EventStreamProcess(this);
             ProcessRegistry.TryAdd("eventstream", eventStreamProcess);
             Extensions = new ActorSystemExtensions(this);
@@ -53,7 +53,7 @@ namespace Proto
 
         public ActorSystemExtensions Extensions { get; }
 
-        public CancellationToken Token => _cts.Token;
+        public CancellationToken Shutdown => _cts.Token;
 
         public Task ShutdownAsync()
         {
--- a/src/Proto.Actor/ActorSystem.cs
+++ b/src/Proto.Actor/ActorSystem.cs
@@ -31,7 +31,7 @@ namespace Proto
             Root = new RootContext(this);
             DeadLetter = new DeadLetterProcess(this);
             Guardians = new Guardians(this);
-            EventStream = new EventStream(config.DeadLetterThrottleInterval, config.DeadLetterThrottleCount);
+            EventStream = new EventStream(config.DeadLetterThrottleInterval, config.DeadLetterThrottleCount, Token);
             var eventStreamProcess = new EventStreamProcess(this);
             ProcessRegistry.TryAdd("eventstream", eventStreamProcess);
             Extensions = new ActorSystemExtensions(this);
--- a/src/Proto.Actor/ActorSystem.cs
+++ b/src/Proto.Actor/ActorSystem.cs
@@ -4,6 +4,8 @@
 // </copyright>
 // -----------------------------------------------------------------------
 using System;
+using System.Threading;
+using System.Threading.Tasks;
 using JetBrains.Annotations;
 using Proto.Extensions;
 
@@ -16,6 +18,7 @@ namespace Proto
         internal const string NoHost = "nonhost";
         private string _host = NoHost;
         private int _port;
+        private CancellationTokenSource _cts = new();
 
         public ActorSystem() : this(new ActorSystemConfig())
         {
@@ -49,6 +52,14 @@ namespace Proto
         public EventStream EventStream { get; }
 
         public ActorSystemExtensions Extensions { get; }
+
+        public CancellationToken Toke => _cts.Token;
+
+        public Task ShutdownAsync()
+        {
+            _cts.Cancel();
+            return Task.CompletedTask;
+        }
 
         public void SetAddress(string host, int port)
         {
--- a/src/Proto.Actor/ActorSystem.cs
+++ b/src/Proto.Actor/ActorSystem.cs
@@ -53,7 +53,7 @@ namespace Proto
 
         public ActorSystemExtensions Extensions { get; }
 
-        public CancellationToken Toke => _cts.Token;
+        public CancellationToken Token => _cts.Token;
 
         public Task ShutdownAsync()
         {
--- a/src/Proto.Actor/EventStream/EventStream.cs
+++ b/src/Proto.Actor/EventStream/EventStream.cs
@@ -7,6 +7,7 @@
 // ReSharper disable once CheckNamespace
 using System;
 using System.Collections.Concurrent;
+using System.Threading;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
 using Microsoft.Extensions.Logging;
@@ -20,11 +21,11 @@ namespace Proto
     {
         private readonly ILogger _logger = Log.CreateLogger<EventStream>();
 
-        internal EventStream() : this(TimeSpan.Zero, 0)
+        internal EventStream() : this(TimeSpan.Zero, 0, CancellationToken.None)
         {
         }
 
-        internal EventStream(TimeSpan throttleInterval, int throttleCount)
+        internal EventStream(TimeSpan throttleInterval, int throttleCount, CancellationToken ct)
         {
             var shouldThrottle = Throttle.Create(throttleCount, throttleInterval,
                 droppedLogs => _logger.LogInformation("[DeadLetter] Throttled {LogCount} logs.", droppedLogs)
@@ -32,7 +33,7 @@ namespace Proto
             Subscribe<DeadLetterEvent>(
                 dl =>
                 {
-                    if (shouldThrottle().IsOpen())
+                    if (!ct.IsCancellationRequested && shouldThrottle().IsOpen())
                     {
                         _logger.LogInformation(
                             "[DeadLetter] could not deliver '{MessageType}:{Message}' to '{Target}' from '{Sender}'",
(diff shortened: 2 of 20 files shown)


src/Proto.Actor/ActorSystem.cs
```diff
@@ -50,6 +53,14 @@ public ActorSystem(ActorSystemConfig config)
 
         public ActorSystemExtensions Extensions { get; }
 
+        public CancellationToken Token => _cts.Token;
```
rogeralsing: Naming? Token is not very semantic

src/Proto.Actor/ActorSystem.cs
```diff
@@ -50,6 +53,14 @@ public ActorSystem(ActorSystemConfig config)
...
+
+        public Task ShutdownAsync()
```
rogeralsing, line 58: Trigger the global killswitch

src/Proto.Actor/EventStream/EventStream.cs
```diff
@@ -20,19 +21,19 @@ public class EventStream : EventStream<object>
     {
         private readonly ILogger _logger = Log.CreateLogger<EventStream>();
 
-        internal EventStream() : this(TimeSpan.Zero, 0)
+        internal EventStream() : this(TimeSpan.Zero, 0, CancellationToken.None)
         {
         }
 
-        internal EventStream(TimeSpan throttleInterval, int throttleCount)
+        internal EventStream(TimeSpan throttleInterval, int throttleCount, CancellationToken ct)
```
rogeralsing, line 28: eventstream takes a CT. in practice it is the killsiwtch CT.

it's used to stop logging deadletter logs once it is triggered for a less noisy shutdown flow

src/Proto.Cluster/Cluster.cs
```diff
@@ -103,8 +103,9 @@ private async Task BeginStartAsync(bool client)
 
         public async Task ShutdownAsync(bool graceful = true)
         {
+            await System.ShutdownAsync();
```
rogeralsing, line 106: Not sure what is the optimal order here.
Maybe it should be the other way around. but this solves the problem for now

src/Proto.Cluster/ClusterHeartBeat.cs
```diff
@@ -78,12 +78,20 @@ private async Task HeartBeatLoop()
                         }
                         catch (TimeoutException)
                         {
+                            if (_cluster.System.Token.IsCancellationRequested)
+                            {
+                                return;
```
rogeralsing, line 83: silent exit if shutting down

src/Proto.Cluster/Identity/IdentityStorageLookup.cs
```diff
@@ -83,6 +85,10 @@ internal PID RemotePlacementActor(string address)
 
         public Task RemovePidAsync(PID pid, CancellationToken ct)
         {
+            if (_system.Token.IsCancellationRequested)
```
rogeralsing: silent exit if shutting down. this one was a major issue before.
RequestAsyncStrategy hitting this method when it detects that there is no activation

src/Proto.Cluster/Identity/IdentityStoragePlacementActor.cs
```diff
@@ -57,8 +71,13 @@ private Task Tick(IContext context)
             return Task.CompletedTask;
         }
 
-        private async Task Terminated(Terminated msg)
+        private async Task Terminated(IContext context, Terminated msg)
         {
+            if (context.System.Token.IsCancellationRequested)
+            {
+                return;
```
rogeralsing, line 78: silent exit on shutdown.

This one was also tricky.
When the placement actor is shutting down, e.g. from IdentityLookup.ShutdownAsync.
it also kills all of its children. which is what parent actors do.

BUT, each child was sending terminated events which we acted on and removed activation entries on

src/Proto.Cluster/Identity/IdentityStorageWorker.cs
```diff
@@ -47,9 +47,18 @@ public Task ReceiveAsync(IContext context)
...
 
-            if (_cluster.PidCache.TryGet(msg.ClusterIdentity, out var existing))
+            var clusterIdentity = msg.ClusterIdentity;
+        //    var ct = msg.CancellationToken;
+            //
+            // if (ct.IsCancellationRequested)
+            // {
+            //     _logger.LogError("CT already timed out....");
+            //     return Task.CompletedTask;
+            // }
```
rogeralsing: if IdentityWorker receives a message, where the CT already have timed out, what should we do with that?
discard? any future waiting for a response is already dead.

src/Proto.Cluster/Identity/IdentityStorageWorker.cs
```diff
@@ -170,6 +179,10 @@ await _storage.WaitForActivationAsync(clusterIdentity, ct)
             }
             catch (Exception e)
             {
+                if (_cluster.System.Token.IsCancellationRequested)
+                {
+                    return null;
```
rogeralsing, line 183: silent exit on shutdown

src/Proto.Cluster/RequestAsyncStrategy.cs
```diff
@@ -34,19 +34,25 @@ public class DefaultClusterContext : IClusterContext
...
         {
+            
              _logger.LogDebug("Requesting {ClusterIdentity} Message {Message}", clusterIdentity.ToShortString(), message
             );
             var i = 0;
             while (!ct.IsCancellationRequested)
             {
+                if (context.System.Token.IsCancellationRequested)
+                {
+                    return default;
```
rogeralsing, line 53: silent exit

src/Proto.Cluster.Consul/ConsulProvider.cs
```diff
@@ -109,12 +109,18 @@ public async Task ShutdownAsync(bool graceful)
             _logger.LogInformation("Shutting down consul provider");
             //flag for shutdown. used in thread loops
             _shutdown = true;
-            if (!graceful) return;
+
+            if (!graceful)
```
mhelleborg: Perhaps reverse this check?
If graceful, deregister etc.
And a single log statement at the end

src/Proto.Cluster/RequestAsyncStrategy.cs
```diff
@@ -63,6 +69,12 @@ public async Task<T> RequestAsync<T>(ClusterIdentity clusterIdentity, object mes
                 try
                 {
                     var pid = await _identityLookup.GetAsync(clusterIdentity, ct);
+                    
+                    if (context.System.Token.IsCancellationRequested)
+                    {
+                        return default;
```
rogeralsing, line 75: silent exit

src/Proto.Cluster/RequestAsyncStrategy.cs
```diff
@@ -80,6 +92,11 @@ public async Task<T> RequestAsync<T>(ClusterIdentity clusterIdentity, object mes
                         "Requesting {Identity}-{Kind} Message {Message} - Got PID {PID} from IdentityLookup",
                         clusterIdentity.Identity, clusterIdentity.Kind, message, pid
                     );
+                    
+                    if (context.System.Token.IsCancellationRequested)
+                    {
+                        return default;
```
rogeralsing, line 98: silent exit

src/Proto.Cluster/RequestAsyncStrategy.cs
```diff
@@ -96,6 +113,11 @@ public async Task<T> RequestAsync<T>(ClusterIdentity clusterIdentity, object mes
                 }
                 catch
                 {
+                    if (context.System.Token.IsCancellationRequested)
+                    {
+                        return default;
```
rogeralsing, line 118: silent exit

src/Proto.Cluster.Identity.MongoDb/MongoIdentityStorage.cs
```diff
@@ -153,21 +154,14 @@ public Task RemoveMemberIdAsync(string memberId, CancellationToken ct)
...
-                }
-
-                // ReSharper disable once MethodSupportsCancellation
-                await Task.Delay(i * 1000);
+                var res = await _asyncSemaphore.WaitAsync(_pids.Find(x => x.Key == key).Limit(1).SingleOrDefaultAsync(ct));
+                return res;
+            }
+            catch (MongoConnectionException x)
+            {
+                Logger.LogWarning(x, "Mongo connection failure, retrying");
```
mhelleborg, line 164: This is no longer true? Or has the retries been moved?

src/Proto.Cluster/Identity/IdentityStorageWorker.cs
```diff
@@ -20,14 +20,14 @@ internal class IdentityStorageWorker : IActor
         private readonly IIdentityStorage _storage;
 
         private readonly Dictionary<ClusterIdentity, Task<PID?>> _inProgress =
-            new Dictionary<ClusterIdentity, Task<PID?>>();
+            new();
```
mhelleborg: Same line ;)

src/Proto.Cluster/Identity/IdentityStorageWorker.cs
```diff
@@ -170,6 +179,10 @@ await _storage.WaitForActivationAsync(clusterIdentity, ct)
             }
             catch (Exception e)
```
mhelleborg, line 179: Were we going to add retries to GetWithGlobalLock? Better to retry a few times here than failing all waiting request and retrying those?

src/Proto.Cluster/Identity/IdentityStorageWorker.cs
mhelleborg: Seems reasonable to discard it then, perhaps return null to sender. We could also link the global cancellationtoken with these.

mhelleborg: And if we should log error, we should probably throttle it :)

src/Proto.Actor/ActorSystem.cs
mhelleborg: StoppingToken?

PR #628, opened 2020-12-24
Polly

rogeralsing:
No Polly yet.
Initial updates to allow IClusterContext to be pluggable in ClusterConfig.

I'm thinking we could and should(?) look into how Polly would look in a case like this

### DIFF
--- a/src/Proto.Cluster/ClusterContext.cs
+++ b/src/Proto.Cluster/ClusterContext.cs
@@ -1,0 +1,15 @@
+// -----------------------------------------------------------------------
+// <copyright file="RequestAsyncStrategy.cs" company="Asynkron AB">
+//      Copyright (C) 2015-2020 Asynkron AB All rights reserved
+// </copyright>
+// -----------------------------------------------------------------------
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Proto.Cluster
+{
+    public interface IClusterContext
+    {
+        Task<T?> RequestAsync<T>(ClusterIdentity clusterIdentity, object message, ISenderContext context, CancellationToken ct);
+    }
+}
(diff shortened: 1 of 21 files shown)


src/Proto.Cluster/ClusterContext.cs
```diff
@@ -0,0 +1,15 @@
...
+// </copyright>
+// -----------------------------------------------------------------------
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Proto.Cluster
+{
+    public interface IClusterContext
+    {
+        Task<T?> RequestAsync<T>(ClusterIdentity clusterIdentity, object message, ISenderContext context, CancellationToken ct);
```
mhelleborg, line 13: Should we have an overload which does not requires passing in an ISenderContext?

src/Proto.Cluster/Identity/IIdentityStorage.cs
```diff
@@ -32,7 +32,8 @@ public SpawnLock(string lockId, ClusterIdentity clusterIdentity)
             ClusterIdentity = clusterIdentity;
         }
     }
-
+    
+    //TODO: what is this? its not used atm..
```
mhelleborg, line 36: Should be cleaned up, but lets do the changes it after merging the identity changes

src/Proto.Cluster/DefaultClusterContext.cs
```diff
@@ -123,22 +115,22 @@ public async Task<T> RequestAsync<T>(ClusterIdentity clusterIdentity, object mes
...
-            PID cachedPid, string source, ISenderContext context)
+            object message, PID cachedPid, string source, ISenderContext context)
         {
             try
             {
                 var res = await context.RequestAsync<T>(cachedPid, message, TimeSpan.FromSeconds(5));
 
                 if (res is not null) return (ResponseStatus.Ok, res);
             }
+            //TODO: all catch logging here should be Debug level, or?
```
mhelleborg, line 133: If we implement cluster distributed PidCache invalidation we could keep in as info.. currently this will be noisy when using local affinity and adding / removing nodes.

mhelleborg (approved):

PR #629, opened 2020-12-25
do not merge

rogeralsing:
@mhelleborg 

Tried to get your changes back for the identity worker.
I still see strange blocking behavior in the cluster benchmark.
We need to find the cause before merging

### DIFF
--- a/src/Proto.Cluster/Identity/IdentityStorageWorker.cs
+++ b/src/Proto.Cluster/Identity/IdentityStorageWorker.cs
@@ -1,25 +1,26 @@
 using Proto.Utils;
 
+
+using System;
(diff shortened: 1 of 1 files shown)


rogeralsing:
It blocks up when member leaves.
I'm thinking it could potentially be something like this:

x requests are buffered waiting for PID to resolve. in the new InProgress list.
Pid is resolved, but shortly after becomes invalid due to member leave.
Now all those requests that got the buffered resolved pid has an incorrect PID.

Might not be that, but at the very least I see full CT timeout for those failed requests as opposed to immediate  retries on deadletter pids in Dev branch.

Haven't seen any obvious issues in the code so far.

PR #631, opened 2020-12-27
System ID

rogeralsing:
Moving ID from Cluster to ActorSystem.
Changed ID from Guid to string. this opens up for better compat with new cluster providers and other language implementations. and if we ever for some reason want to have named members.


rogeralsing:
Thoughts? should we roll with this?
It's a good start to get block-lists into remote

mhelleborg:
Looks good to me, merge away :)

mhelleborg (approved):

PR #638, opened 2020-12-29
decorator context

rogeralsing:
small PR to add DecoratorContext base class, to make it easier to create decorators


mhelleborg (approved):
Do you think we can point the dotnet team to https://kotlinlang.org/docs/reference/delegation.html so we don't need this? ;)

PR #706, opened 2021-01-17
WiP - Plugin experiments

rogeralsing:
Plugin experiments, just ignore for now


src/Proto.Cluster/Cache/CacheInvalidationExtensions.cs
```diff
@@ -32,7 +32,7 @@ public static Props WithPidCacheInvalidation(this Props props)
 
         public static Cluster WithPidCacheInvalidation(this Cluster cluster)
```
mhelleborg, line 33: Can we perhaps refactor so we are able to register cluster plugins against the cluster config?

It seems that might be cleaner, and can be initialized in the correct order on startup.

rogeralsing: `clusterConfig.WithExtension<TExt>` maybe?

mhelleborg: Could work. But it would not let you pass config to the extension?

PR #745, opened 2021-02-22
Moved build into dockerfiles for ClusterGrainHelloWorld example, and …

chezuma:
…updated docker-compose.

## Description

running run.sh in /examples/ClusterGrainHelloWorld, failed due to a static path requiring the example to be compiled with Core 3.1. 

moved the build of the cotainers into the dockerfiles with the correct base images for build, so that we do not rely on static paths.

## Purpose
This pull request is a:

### DIFF
--- a/examples/ClusterGrainHelloWorld/Node1/Dockerfile
+++ b/examples/ClusterGrainHelloWorld/Node1/Dockerfile
@@ -1,10 +1,10 @@
-FROM mcr.microsoft.com/dotnet/core/aspnet:3.1.0-alpine3.10
+FROM mcr.microsoft.com/dotnet/sdk:5.0 AS base
+COPY . .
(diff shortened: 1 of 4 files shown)


chezuma:
I have signed the licence/cla but it do not update, does this usually take some time?

alexeyzimarev:
> I have signed the licence/cla but it do not update, does this usually take some time?

Seems like you need to connect the email address:

> Andreas Kristiansen seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please add the email address used for this commit to your account.

chezuma:
Apparently I forgot to set the User.email property when committed the first commit. 

Therefor there are no link to my github user on that commit, the second commit have my thumbnail, so I guess it is ok :)

chezuma:
Because of my commit without email address the cla would never be approved

PR #773, opened 2021-03-09
Metrics again

rogeralsing:
I think this will work.
We can add cached known counters via the Metrics.KnownMetrics lookup.
Running this using the SpawnBenchmark shows no visible performance loss when using a noop spawn counter.


src/Proto.Actor/Metrics/ActorMetrics.cs
```diff
@@ -0,0 +1,18 @@
...
+//      Copyright (C) 2015-2021 Asynkron AB All rights reserved
+// </copyright>
+// -----------------------------------------------------------------------
+namespace Proto.Metrics
+{
+    public class ActorMetrics
+    {
+        public ActorMetrics(Metrics metrics)
+        {
+            SpawnCount = metrics.CreateCount(nameof(SpawnCount), new string[] { });
```
rogeralsing: The only known counter in this PoC

src/Proto.Actor/Metrics/Metrics.cs
```diff
@@ -0,0 +1,123 @@
...
+    {
+        private IConfigureMetrics[] _configurators;
+
+        internal TypeDictionary<object> KnownMetrics { get; }
+        
+        public Metrics(IConfigureMetrics[] configurators)
+        {
+            _configurators = configurators;
+            KnownMetrics = new TypeDictionary<object>();
+            KnownMetrics.Add<ActorMetrics>(new ActorMetrics(this));
```
rogeralsing: Add the ActorMetrics known counters to the fast lookup

src/Proto.Actor/Props.cs
```diff
@@ -117,7 +118,11 @@ public Props WithSenderMiddleware(params Func<Sender, Sender>[] middleware)
         public Props WithSpawner(Spawner spawner) =>
             this with {Spawner = spawner};
 
-        internal PID Spawn(ActorSystem system, string name, PID? parent) => Spawner(system, name, this, parent);
+        internal PID Spawn(ActorSystem system, string name, PID? parent)
+        {
+            system.Metrics.Get<ActorMetrics>().SpawnCount.Inc();
```
rogeralsing: Increment the SpawnCount on Spawn

rogeralsing:
cc @mhelleborg  @alexeyzimarev

src/Proto.Actor/Metrics/NoMetrics.cs
```diff
@@ -0,0 +1,68 @@
...
+            public void Inc(int count = 1, params string[]? labels) => _count += count;
+        }
+
+        private class NoHistogram : NoMetric, IHistogramMetric
+        {
+            protected internal NoHistogram(string name, string[] labelNames) : base(name, labelNames)
+            {
+            }
+
+            public void Observe(Stopwatch stopwatch, string[]? labels = null, int count = 1) => Observe(stopwatch.ElapsedMilliseconds * 1000, labels);
```
mhelleborg: For debugging purposes? No-op should maybe do nothing?

src/Proto.Actor/Metrics/ICountMetric.cs
```diff
@@ -0,0 +1,7 @@
+namespace Proto.Metrics
+{
+    public interface ICountMetric
+    {
+        void Inc(int count = 1, params string[]? labels);
```
mhelleborg: long instead of int? Prevents overflow in high-volume use cases, and will also align better with Prometheus

src/Proto.Actor/Metrics/IHistogramMetric.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+using System.Diagnostics;
+
+namespace Proto.Metrics
+{
+    public interface IHistogramMetric
+    {
+        void Observe(Stopwatch stopwatch, string[]? labels = null, int count = 1);
```
mhelleborg: int -> long, same as ICount

src/Proto.Actor/Metrics/NoMetrics.cs
mhelleborg: On second thought, some metrics might be valuable from a cluster perspective, even if we are not exporting them. Then we might want to expose them internally

rogeralsing:
Maybe useful metrics to copy?
https://kamon.io/docs/latest/instrumentation/akka/metrics/

rogeralsing:
And this https://developer.lightbend.com/docs/telemetry/current/instrumentations/akka/akka.html

rogeralsing:
Merging initial prerelease so we can try it out

PR #786, opened 2021-03-18
Dedupe cluster requests WIP

### LINKED ISSUE #778: Improved Cluster Requests
Cluster requests as currently implemented has some drawbacks. Each ClusterRequestAsync-call might make multiple calls to RequestAsync (during retries), which will be handled as separate messages by the virtual actor (if the calls made it over the network). 
Messages does not even have to be lost to make the requester retry. If the virtual actor is slow to respond, new messages will be sent.

I propose a way to prevent the drawback of messages being processed multiple times, while still allowing quick retries on the requester side.

For each ClusterRequest:
- Create a FutureProcess for the response.
- Send the request and any retries with the same FutureProcess as sender.
- On the virtual actor side, we could recognize that the sender PID is a future in middleware, and deduplicate on the ID, keeping successfully processed sender PID's in a short window.

/cc @rogeralsing

rogeralsing:


rogeralsing:
Looks great!

PR #791, opened 2021-03-22
Cluster remote experiments

rogeralsing:
Reduce noisiness on topology changes.

Any retriable action is changed from Error to Warn logging except for the last attempt.
This affects IdentityWorker and EndpointActor.

Removed stacktrace for known Error types, such as Unavailable gRPC status.
We know what that means, the remote node is unavailable, there is no value in seeing the entire stacktrace for gRPC there.

### DIFF
--- a/benchmarks/ClusterBenchmark/Program.cs
+++ b/benchmarks/ClusterBenchmark/Program.cs
@@ -120,7 +120,7 @@ namespace ClusterExperiment1
                         var id = "myactor" + rnd.Next(0, ActorCount);
                         semaphore.Wait(() => {
                                 return cluster.RequestAsync<HelloResponse>(id, "hello", new HelloRequest(),
-                                    CancellationTokens.WithTimeout(20)
+                                    CancellationTokens.WithTimeout(20*1000)
                                 ).ContinueWith(task => { Console.Write(task.Result is null ? "X" : "."); }
                                 );
                             }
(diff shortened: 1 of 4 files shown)


benchmarks/ClusterBenchmark/Program.cs
```diff
@@ -119,7 +120,7 @@ private static void RunFireForgetClient()
                         var id = "myactor" + rnd.Next(0, ActorCount);
                         semaphore.Wait(() => {
                                 return cluster.RequestAsync<HelloResponse>(id, "hello", new HelloRequest(),
-                                    new CancellationTokenSource(TimeSpan.FromSeconds(5)).Token
+                                    CancellationTokens.WithTimeout(20*1000)
```
rogeralsing, line 125: Extend timeout in cluster benchmarks to allow for retries inside the IdentityWorker.
If the request times out faster than the identity worker retries, you get null response

src/Proto.Cluster/Identity/IdentityStorageWorker.cs
```diff
@@ -131,9 +134,18 @@ private void RespondToWaitingRequests(IContext context, ClusterIdentity clusterI
                     else
                     {
                         //we have the lock, spawn and return
-                        (result, spawnLock) = await SpawnActivationAsync(activator, spawnLock, CancellationTokens.WithTimeout(5000));
+                        (result, spawnLock) = await SpawnActivationAsync(activator, spawnLock, CancellationTokens.WithTimeout(1000));
                     }
                 }
+                catch (OperationCanceledException e)
```
rogeralsing, line 140: Warning level logging for timeouts.
The system is not broken because we got a timeout in a retry loop

src/Proto.Remote/Endpoints/EndpointSupervisorStrategy.cs
```diff
@@ -42,7 +42,7 @@ public EndpointSupervisorStrategy(string address, RemoteConfigBase remoteConfig,
         {
             if (ShouldStop(rs))
             {
-                Logger.LogError(reason,
+                Logger.LogError(
```
rogeralsing, line 45: gRPC stack traces adds no value here

PR #826, opened 2021-03-27
Clusterkinds member strategy

rogeralsing:


rogeralsing:
Updates.
Brought back the MemberStrategyBuilder on ClusterConfig..
Made the ClusterKind able to have a null builder and return a null strategy

The memberlist will use the strategy for the ClusterKind, if the ClusterKind exists and have a strategy.
Otherwise, it will use the member strategy builder like before.

We need to be able to get a member strategy for kinds that exist on other nodes, as it is the caller node that invokes the strategy and selects the activator node.
If we don't have this, a node that doesn't have a specific kind will never be able to invoke such a virtual actor unless it is already activated by someone else.

TLDR; the old behavior is still there, but if you want, you can provide a member strategy per registered Kind.

So we are backward compatible, and bring in a new nice to have feature.

A longer discussion would be, should really the calling node be the one that selects the activator node for kinds it doesn't know anything about.
Probably not, not sure how it should work instead, so we are keeping this behavior for now

PR #844, opened 2021-03-30
SqlServerProvider - Creating database schema failed

MarkEwer:
## Description

When using the `SqlServerProvider` for persistence, the option to create the database schema failed because the connection to SQL Server was not open. I added code to open the connection.

Additionally, I made some minor changes to the `InMemoryProvider` used during testing so that it would work better for my unit testing needs. I have added these changes as well to share them with the community.

## Purpose
This pull request is a:

### DIFF
--- a/src/Proto.Persistence.SqlServer/SqlServerProvider.cs
+++ b/src/Proto.Persistence.SqlServer/SqlServerProvider.cs
@@ -234,6 +234,8 @@ namespace Proto.Persistence.SqlServer
 
             using var command = new SqlCommand(sql, connection);
 
+            connection.Open();
+
             command.ExecuteNonQuery();
         }
     }
--- a/tests/Proto.Persistence.Tests/InMemoryProvider.cs
+++ b/tests/Proto.Persistence.Tests/InMemoryProvider.cs
@@ -30,15 +30,17 @@ namespace Proto.Persistence.Tests
 
         public Task<long> GetEventsAsync(string actorName, long indexStart, long indexEnd, Action<object> callback)
         {
+            var lastIndex = 0l;
             if (_events.TryGetValue(actorName, out var events))
             {
                 foreach (var e in events.Where(e => e.Key >= indexStart && e.Key <= indexEnd))
                 {
+                    lastIndex = e.Key;
                     callback(e.Value);
                 }
             }
 
-            return Task.FromResult(0L);
+            return Task.FromResult(lastIndex);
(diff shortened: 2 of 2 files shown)


src/Proto.Persistence.SqlServer/SqlServerProvider.cs
```diff
@@ -234,6 +234,8 @@ private void ExecuteNonQuery(string sql)
 
             using var command = new SqlCommand(sql, connection);
 
+            connection.Open();
```
MarkEwer, line 237: This was the main problem for me. When I first initialized the persistence provider I got an exception that the connection was not open.  The rest of the places in this class that we use the SQL Connection object we open it explicitly but, for some reason, this one didn't have a call to open the connection.  This change seems to correct the issue.

tests/Proto.Persistence.Tests/InMemoryProvider.cs
```diff
@@ -30,15 +30,17 @@ public Task<(object Snapshot, long Index)> GetSnapshotAsync(string actorName)
...
             {
                 foreach (var e in events.Where(e => e.Key >= indexStart && e.Key <= indexEnd))
                 {
+                    lastIndex = e.Key;
                     callback(e.Value);
                 }
             }
 
-            return Task.FromResult(0L);
+            return Task.FromResult(lastIndex);
```
MarkEwer, line 43: I don't think this causes any side effects since the `InMemoryProvider` was only used in the unit tests. But, this helped me to write a better test for my domain model object because the return value shows the key given to the event in the storage provider.

tests/Proto.Persistence.Tests/InMemoryProvider.cs
```diff
@@ -47,7 +49,9 @@ public Task<long> PersistEventAsync(string actorName, long index, object @event)
 
             events.Add(index, @event);
 
-            return Task.FromResult(0L);
+            long max = events.Max(x => x.Key);
```
MarkEwer, line 52: I don't think this causes any side effects since the InMemoryProvider was only used in the unit tests. But, this helped me to write a better test for my domain model object.

MarkEwer:
~I have signed the CLA but it still says that I need to for some reason. Maybe you have a manual check involved to verify this or something?~ I just needed to add my work email to my GitHub account.

rogeralsing:
Many thanks for the fix!

PR #846, opened 2021-03-31
Task clock

rogeralsing:
Instead of creating cancellation token sources for every request, we can use a shared clock that pulses out a new task every x seconds. consumers can then pick up the current task from the clock and use that as a shared timeout.

e.g.
we can pulse every 1 sec and emit a task that delays for 5 seconds.
the next second, it pulses another 5 sec task.

meaning that we have ~5 delay tasks scheduled at any given point in time.

```
|----
 |----
  |----
   |----
    |----
     |---- when this starts the first task is done
```

### DIFF
--- a/src/Proto.Cluster/DefaultClusterContext.cs
+++ b/src/Proto.Cluster/DefaultClusterContext.cs
@@ -24,8 +24,9 @@ namespace Proto.Cluster
         private readonly ILogger _logger;
         private readonly PidCache _pidCache;
         private readonly ShouldThrottle _requestLogThrottle;
-        
-        public DefaultClusterContext(IIdentityLookup identityLookup, PidCache pidCache, ILogger logger, ClusterContextConfig config)
+        private readonly TaskClock _clock;
+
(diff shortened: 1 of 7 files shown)


rogeralsing:
Pretty sure this approach can be used in a lot of other places. e.g. Id provider timeouts

src/Proto.Cluster/DefaultClusterContext.cs
```diff
@@ -36,6 +37,8 @@ public DefaultClusterContext(IIdentityLookup identityLookup, PidCache pidCache,
                 _config.RequestLogThrottlePeriod,
                 i => _logger.LogInformation("Throttled {LogCount} TryRequestAsync logs", i)
             );
+            _clock = new TaskClock(config.ActorRequestTimeout, TimeSpan.FromSeconds(1), killSwitch);
```
mhelleborg, line 40: Perhaps make sure that the chosen timeout is at least ActorRequestTimeout, by adding the updateInterval to the timeout delays.

PR #847, opened 2021-03-31
SqlServerProvider - Create custom schema if `AutoCreateTables`=true

MarkEwer:
## Description

When the `SqlServerProvider` for data persistence is used it contains an option to create the database tables needed for storage the first time it is used. However, if you specify that the tables should be located in a _schema_ that does not already exist in the database then the provider will throw an exception because it cannot create the table if the schema is not present.

In this pull request, I have added a new method to the startup code for the `SqlServerProvider` class that will create the requested schema if it does not already exist in the database.

## Purpose
This pull request is a:

## Checklist


MarkEwer:
I'd like to add some documentation on how to configure and use the `SqlServerProvider` class now that I think I have figured it out. Where does this project put the developer documentation?  Should I just add a `readme.md` inside the `Proto.Persistence.SqlServer` project?

rogeralsing:
👍🏻

PR #852, opened 2021-04-02
proper hash calculation

rogeralsing:
* Reverted the interlocked inc/dec of concurrent hashtable
* Fix hash calculation

### DIFF
--- a/src/Proto.Actor/HashedConcurrentDictionary.cs
+++ b/src/Proto.Actor/HashedConcurrentDictionary.cs
@@ -15,6 +15,7 @@ namespace Proto
     
     class HashedConcurrentDictionary<TKey, TValue>
     {
+        //power of two
         private const int HashSize = 1024;
         private readonly Dictionary<TKey,TValue>[] _partitions = new Dictionary<TKey,TValue>[HashSize];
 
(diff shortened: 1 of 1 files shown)


src/Proto.Actor/HashedConcurrentDictionary.cs
```diff
@@ -32,7 +34,7 @@ internal HashedConcurrentDictionary()
 
         private Dictionary<TKey,TValue> GetPartition(TKey key)
         {
-            var hash = key.GetHashCode() & (0x7FFFFFFF % HashSize);
+            var hash = key.GetHashCode() & HashMask;
```
rogeralsing, line 37: The previous code can't be correct, modulus of this bit mask and then bitwise and of the hashcode.
The new code assumes hash size is a power of two, then creates a mask for all bits below
e.g.
1111111111

This guarantees a positive result, less than HashSize with less computations.

src/Proto.Actor/HashedConcurrentDictionary.cs
```diff
@@ -58,36 +60,17 @@ public bool TryGetValue(TKey key, out TValue value)
...
         {
             var p = GetPartition(key);
 
             lock (p)
             {
                 if (p.Remove(key))
                 {
-                    _count--;
-                    return true;
+                    Interlocked.Decrement(ref _count);
```
rogeralsing, line 71: bring back interlocked

PR #853, opened 2021-04-02
Value tasks

rogeralsing:
* Value Tasks in ActorContext.
* Vastly simplified code in Mailbox and Receive loop
* Performance Improvement 🚀 💥 

InProcessBenchmark max on this machine:
Before:  47mln msg/sec
After:  58mln msg/sec


rogeralsing:
https://github.com/AsynkronIT/protoactor-dotnet/blob/a5c66c54eb242c77301e2ea36e354777dfc94111/src/Proto.Actor/Mailbox/Mailbox.cs#L153-L195

PR #866, opened 2021-04-07
Local affinity 2.0

mhelleborg:
New way to do local affinity relocation.

Uses new proxy activator to re-spawn actors directly after relocation. Prevents following requests from having to cold-start actors, and should make the relocation process smoother.


src/Proto.Cluster/ClusterExtension.cs
```diff
@@ -31,22 +31,33 @@ public static Task<T> ClusterRequestAsync<T>(this IContext context, string ident
         {
             var cluster = context.System.Extensions.Get<Cluster>();
```
rogeralsing: context.System.Cluster();

src/Proto.Cluster/ClusterExtension.cs
```diff
@@ -31,22 +31,33 @@ public static Task<T> ClusterRequestAsync<T>(this IContext context, string ident
...
-                    (ctx, env) => {
-                        return env.Message switch
-                        {
-                            Started => HandleStart(baseReceive, ctx, env),
-                            Stopped => HandleStopped(baseReceive, ctx, env),
-                            _       => baseReceive(ctx, env)
-                        };
+            return props
+                .WithContextDecorator(context => {
+                        context.SetMetadata(clusterIdentity);
```
rogeralsing: This is where the recursion happens?

mhelleborg: Yes, since setting metadata requires extras to be instantiated

rogeralsing:
OnCreate vs OnInit vs OnIncarnate

In actor context we call it "Incarnate Actor", for current cluster we have ClusterInit.
Maybe Incarnate is just too weird of a word to use?

Not important from a technical PoV, but consistency matters. so lets ponder this a bit

rogeralsing:
Depending on how much work left here, it might be worth it to extract the new GetMetadata bits and Props OnCreate to a separate PR. just to get some control on each part

rogeralsing:
Another reflection, GetMetadata, is it only metadata we ever going to store there?
e.g. if we store a persistence provider / Redis state store or something like that

Just a thought that there might be bigger use cases lurking here

mhelleborg:
> Another reflection, GetMetadata, is it only metadata we ever going to store there?
> e.g. if we store a persistence provider / Redis state store or something like that
> 
> Just a thought that there might be bigger use cases lurking here

I think you are right, we should perhaps update the interfaces to support a wider use case. It will definitely look weird to pull a persistence provider from the metadata store, even if it would have worked.

mhelleborg:
> OnCreate vs OnInit vs OnIncarnate
> 
> In actor context we call it "Incarnate Actor", for current cluster we have ClusterInit.
> Maybe Incarnate is just too weird of a word to use?
> 
> Not important from a technical PoV, but consistency matters. so lets ponder this a bit

Perhaps go with OnInit, as that matches the ClusterInit concept we already have?

src/Proto.Cluster/ClusterExtension.cs
```diff
@@ -27,26 +27,30 @@ public static Cluster Cluster(this ActorSystem system)
...
-                        return env.Message switch
-                        {
-                            Started => HandleStart(baseReceive, ctx, env),
-                            Stopped => HandleStopped(baseReceive, ctx, env),
-                            _       => baseReceive(ctx, env)
-                        };
-                    }
-            );
+            return props
+                .WithOnInit(context => context.Set(clusterIdentity))
```
rogeralsing, line 42: I really like how this turned out, super smooth and aligns well with rest of API

rogeralsing:
Looks great overall.
I'm thinking if we have other migration strategies. e.g. not checking sender but checking memory pressure or similar.
Would classes like LocalAffinityOptions for throttling migrations, be called something else then?
Just so we keep the options open going forward.

TLDR; LocalAffinity might be a strategy of some form of RelocationStrategy or something like that?

mhelleborg:
> Looks great overall.
> I'm thinking if we have other migration strategies. e.g. not checking sender but checking memory pressure or similar.
> Would classes like LocalAffinityOptions for throttling migrations, be called something else then?
> Just so we keep the options open going forward.
> 
> TLDR; LocalAffinity might be a strategy of some form of RelocationStrategy or something like that?

We could generalize the throttle options at least, those can be re-used anywhere we need to throttle throughput.

I'm not sure how the RelocationStrategy interface should look like yet, but it sounds like a bigger rewrite including rethinking how IMemberStrategy fits into it. We could make use of the new metrics available for the cluster, propagated via gossip.

I think that is a new PR though :)

PR #870, opened 2021-04-10
Refactored ClusterInit, receiver per kind instead of per activation

mhelleborg:
Identity will now be retrieved from the actor-local context store, eliminating the need for a specific receiver per identity.


rogeralsing (approved):
Looks great!

rogeralsing:
Pulling this in

PR #888, opened 2021-04-12
Extension methods for sends, requests, stops

rogeralsing:


rogeralsing:
Removing RequestAsync breaks context decorators. they need it to be able to apply logic before and after task completes

rogeralsing:
OK closing this. it should not have been extension methods as this messes up interception via decorators.

Default impl interface methods should work better

PR #905, opened 2021-04-16
System.Text.Json remote Serializer

Damian-P:
## Description

This PR allows to register a serializer that serializes registered IMessage types as protobuf and other types with System.Text.Json.

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/src/Proto.Remote/Proto.Remote.csproj
+++ b/src/Proto.Remote/Proto.Remote.csproj
@@ -9,6 +9,7 @@
     <PackageReference Include="Grpc.Core" Version="2.36.4" />
     <PackageReference Include="Grpc.HealthCheck" Version="2.37.0" />
     <PackageReference Include="Grpc.Tools" Version="2.36.4" PrivateAssets="All" />
+    <PackageReference Include="System.Text.Json" Version="5.0.2" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\Proto.Actor\Proto.Actor.csproj" />
--- a/src/Proto.Remote/RemoteConfigExtensions.cs
+++ b/src/Proto.Remote/RemoteConfigExtensions.cs
@@ -7,6 +7,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.Json;
 using Google.Protobuf.Reflection;
 using Grpc.Core;
 using JetBrains.Annotations;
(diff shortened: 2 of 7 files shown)


rogeralsing:
Very cool! have you done any benchmarks if this affects e.g. remote benchmarks anything?
I'll try to review this fully during later today

Damian-P:
When switching to the SystemTextJsonSerializer there is no noticeable performance impact on my computer (around 3M msg/s in the remote benchmark).
If the protobuf messages are not registered, then the performances in the remote benchmark are very poor comparing to the protobuf format. It gives me around 300k msg/s.

In the cluster benchmark, when not registering the protobuf messages it gives me 23k msg/s and 29k msg/s when they are registered.

The idea is to provide an abstract class with default protobuf serialization to serialize and deserialize all registered protobuf messages (especially the internal ones) and the user can then implement a custom serialization for other messages.

rogeralsing:
Some thoughts here:
I'm a bit confused about the type hierarchy.

Why is it SystemTextJsonSerializer `: ProtobufSerializer`
and not `: ISerializer` ?

* It seems like it does not rely on anything in the base class
* it also seems to be a json serializer that can serialize any custom CLR type, not associated with the Protobuf Json format

Is it just a bit of bad naming of the base class, e.g. it's a ProtoActorSerializer, not a ProtobufSerializer ?

Damian-P:
The base class relies on the implementation.

There is an explicit interface implementation that calls the protected abstract methods.
```c#
        object ISerializer.Deserialize(ByteString bytes, string typeName)
        {
            if (_serialization.TypeLookup.TryGetValue(typeName, out var parser))
                return parser.ParseFrom(bytes);
            return Deserialize(bytes, typeName);
        }
```
```c#
        protected abstract object Deserialize(ByteString bytes, string typeName);
```

The same concept applies to `GetTypeName()` and `Serialize()` so when the call is made to the serializer to deserialize a message, it tries to do his work in the interface implementation on the ProtobufSerializer class and if it can't handle the type then it passes the bytestring to the abstract method, here to the SystemTextJsonSerializer methods.

I have done this that way to keep the compatibility.

Users could then provide their own implementation of the ProtobufSerializer base class for the messages that the default serializer don't handle.

Of course we can rename the base class as you wish but it uses Protobuf so I though that it was a good name.
We can rename the DefaultProtobufSerializer to ProtoActorSerializer.

With this base class we can make the ISerializer internal and force users to rely on it to keep the internal messages being serialized with protobuf.

PR #912, opened 2021-04-20
Fix actor context

Damian-P:
## Description

When using an asynchronous method in the Receive method, exceptions that are thrown are ignored.

## Purpose
This pull request is a:

## Checklist


Damian-P:
I see that this IsCompleted property is used as well in Futures.cs and DefaultClusterContext.cs, maybe it's worth checking.

rogeralsing:
Good catch!

PR #913, opened 2021-04-20
Added ability to use reentrancy in Grains.

mhelleborg:
In case the grains require reentrancy, the ability to override a callback-based version of the method is added.
Override on any method which require reentrant behavior.

Added Grain errors on invalid grain requests (either parameter or method index incorrect)

### DIFF
--- a/protobuf/ProtoGrainGenerator/Template.cs
+++ b/protobuf/ProtoGrainGenerator/Template.cs
@@ -11,8 +11,10 @@ namespace ProtoBuf
 using System;
 using System.Threading;
 using System.Threading.Tasks;
+using Google.Protobuf;
 using Proto;
 using Proto.Cluster;
+
 
 namespace {{CsNamespace}}
 {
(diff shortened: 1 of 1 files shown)


protobuf/ProtoGrainGenerator/Template.cs
```diff
@@ -117,29 +132,25 @@ public async Task ReceiveAsync(IContext context)
...
-                            catch (Exception x)
-                            {
-                                var grainErrorResponse = new GrainErrorResponse
-                                {
-                                    Err = x.ToString()
-                                };
-                                context.Respond(grainErrorResponse);
+                        {   
+                            if(r is {{InputName}} input){
+                                await _inner.{{Name}}(input, res => context.Respond( new GrainResponseMessage(res) ), onError);
```
rogeralsing: `res => context.Respond( new GrainResponseMessage(res) )`

Could be a private instance method if we want to get rid of the closure capture we talked about. 

```
 await _inner.{{Name}}(input, respond, onError);

....

private void respond<T>(T response)
{
    _context.respond(response);
}
```

mhelleborg: Not a bad idea, fixing

PR #933, opened 2021-04-23
JsonSerialization

Damian-P:
## Description

This PR adds System.Text.Json serializer as a fallback.
We can configure the options in RemoteConfig.
Not registered protobuf types are serialized with json serializer but a warning is logged.


rogeralsing:
Very nice! 👍🏻

PR #941, opened 2021-04-25
ClusterRequest overload using ClusterIdentity

mhelleborg:
Added ClusterIdentity overload to Cluster-request. Slightly more efficient, but allows for later optimizations with direct references to kind and cached Pid.

Added Cluster-request microbenchmarks


src/Proto.Cluster/Cluster.cs
```diff
@@ -148,6 +148,12 @@ public async Task ShutdownAsync(bool graceful = true)
 
         public Task<T> RequestAsync<T>(string identity, string kind, object message, ISenderContext context, CancellationToken ct) =>
             ClusterContext.RequestAsync<T>(new ClusterIdentity {Identity = identity, Kind = kind}, message, context, ct)!;
+        
+        public Task<T> RequestAsync<T>(ClusterIdentity clusterIdentity, object message, CancellationToken ct) =>
```
rogeralsing: Move these to be extension methods maybe?
Keeping the API surface of the core types smaller generally makes it easier to evolve.

PR #950, opened 2021-05-01
Shared futures

rogeralsing:


src/Proto.Actor/Configuration/ActorSystemConfig.cs
```diff
@@ -28,6 +28,9 @@ public record ActorSystemConfig
         public bool DeadLetterRequestLogging { get; set; } = true;
         public bool DeveloperSupervisionLogging { get; init; }
 
+        public bool UseSharedFutures { get; init; } = true;
+        public int SharedFutureSize { get; init; } = 5000;
+
```
rogeralsing, line 33: WithSharedFutureSize & WithUseSharedFutures saknas?

src/Proto.Actor/Future/Futures.cs
```diff
@@ -33,13 +30,12 @@ internal FutureProcess(ActorSystem system) : base(system)
 
             if (!absent) throw new ProcessNameExistException(name, pid);
 
-            Pid = pid;
+            Pid = pid.WithRequestId(1);
```
rogeralsing: What happens if this is 0?
micro optimization, 0 is not transmitted in protobuf. saving one field when passing the pid remote

src/Proto.Actor/Future/Futures.cs
```diff
@@ -51,10 +47,7 @@ await using (cancellationToken.Register(() => _tcs.TrySetCanceled()))
             }
             catch
             {
-                if (!_system.Metrics.IsNoop)
-                {
-                    _metrics!.FuturesTimedOutCount.Inc(new[] {System.Id, _system.Address});
-                }
+                _metrics?.FuturesTimedOutCount.Inc(new[] {System.Id, System.Address});
```
rogeralsing: is metrics changed to be null instead of Noop provider as default?
If not, we don't want to touch the metrics API if the noop provider is present.
as the Inc allocates a new object array each call

src/Proto.Actor/Future/Futures.cs
```diff
@@ -69,10 +62,7 @@ protected internal override void SendUserMessage(PID pid, object message)
             }
             finally
             {
-                if (!_system.Metrics.IsNoop)
-                {
-                    _metrics!.FuturesCompletedCount.Inc(new[] {System.Id, System.Address});
-                }
+                _metrics?.FuturesCompletedCount.Inc(new[] {System.Id, System.Address});
```
rogeralsing, line 99: same here

src/Proto.Actor/Future/Futures.cs
```diff
@@ -88,10 +78,7 @@ protected internal override void SendSystemMessage(PID pid, object message)
 
             _tcs.TrySetResult(default!);
 
-            if (!_system.Metrics.IsNoop)
-            {
-                _metrics!.FuturesCompletedCount.Inc(new[] {System.Id, System.Address});
-            }
+            _metrics?.FuturesCompletedCount.Inc(new[] {System.Id, System.Address});
```
rogeralsing: And here

src/Proto.Actor/Future/Futures.cs
mhelleborg: I think metrics actually warrants some redesign overall. The abstraction here should not pass any system level parameters at all, but expose a parameterless .Inc() method.

This will simplify usage, and internally we can pre-assign the system parameters and not do it every call..

mhelleborg: And No-ops could be called without checking if it actually is a no-op.

src/Proto.Cluster/ClusterExtension.cs
```diff
@@ -105,10 +106,10 @@ bool TryGetRef(MessageEnvelope envelope, out PidRef pidRef)
                     {
                         var pid = envelope.Sender;
 
-                        if (pid is not null && int.TryParse(pid.Id[1..], out var id) &&
+                        if (pid is not null && pid.RequestId > 0 && int.TryParse(pid.Id[1..], out var id) &&
```
rogeralsing, line 108: >int.TryParse(pid.Id[1..], out var id)

We parse the $123 id for the future and what do we do with the 123 part then?

src/Proto.Cluster/PidCache.cs
```diff
@@ -25,6 +25,12 @@ public bool TryGet(ClusterIdentity clusterIdentity, out PID pid)
         {
             if (clusterIdentity is null) throw new ArgumentNullException(nameof(clusterIdentity));
 
+            if (clusterIdentity.CachedPid is { } identityCachedPid)
+            {
+                //If the PID is already cached using ClusterIdentity, we can skip the lookup altogether
+                pid = identityCachedPid;
```
rogeralsing, line 31: is this invalidated somewhere? so we don't get get stuck with an invalid pid at some point

src/Proto.Cluster/PidCache.cs
```diff
@@ -45,22 +54,30 @@ public bool TryUpdate(ClusterIdentity clusterIdentity, PID newPid, PID existingP
...
             return _cacheDict.TryRemove(clusterIdentity, out _);
         }
 
         public bool RemoveByVal(ClusterIdentity clusterIdentity, PID pid)
         {
-            var key = clusterIdentity;
-            if (_cacheDict.TryGetValue(key, out var existingPid) && existingPid.Id == pid.Id &&
+            if (clusterIdentity.CachedPid?.Equals(pid) == true)
+            {
+                clusterIdentity.CachedPid = null;
```
rogeralsing, line 75: I suppose this answers my previous question

src/Proto.Cluster/ClusterExtension.cs
mhelleborg: It becomes part of the PidRef, a cheaper way to store the Pid than as a string.

mhelleborg: And if it is not possible to parse, it is not an unnamed Future-pid, which is what we are checking for in the filter.

src/Proto.Actor/Future/Futures.cs
mhelleborg: The advantage with having requestId 1 is that we can classify it as a single request in the deduplication filter. The disadvantage is ofc that there are more bytes transferred, but a pretty negligible difference..

src/Proto.Actor/Future/Futures.cs
rogeralsing: There are plenty of other cases where that is relevant though. e.g. message type, actor type etc.
We still need a pattern that is intact for this. in the current form, the Noop is what we have

mhelleborg: You would still pass the specific metric values at the call point, I was just thinking we did not need to pass the values that are common for the entire system. It would have made the interfaces a little cleaner IMO

rogeralsing:
Looks great! fantastic work on this one!
Merging

PR #962, opened 2021-05-04
Updated IIdentityStorage with ClusterIdentity on pid removal.

mhelleborg:
Removes need for secondary indexing on pid.id


rogeralsing:
👍🏻

PR #985, opened 2021-05-17
Fix serialization of message implementing IRootSerializable

jakubgrabny:
## Description

Proto generator creates grain client with method proxy which wraps request in `GrainRequestMessage`. `GrainRequestMessage` includes inner serialization method however its outcome was never used in `EndpointActor`. Serialization of `GrainRequestMessage` with JsonSerializer resulted in serialization exception with message: "Serialization and deserialization of 'System.Type' instances are not supported and should be avoided since they can lead to security issues. Path: $.Columns.DataType."

## Purpose
This pull request is a:

## Checklist


rogeralsing:
Good catch!, merging this

PR #986, opened 2021-05-17
Shredded Futures

rogeralsing:
Requests:	19 509 755
Successful:	19 509 300
Failures:	0
Throughput:	325 060 requests/sec -> 650 120 msg/sec

### DIFF
--- a/benchmarks/ClusterBenchmark/Program.cs
+++ b/benchmarks/ClusterBenchmark/Program.cs
@@ -243,7 +243,7 @@ namespace ClusterExperiment1
                     for (var i = 0; i < batchSize; i++)
                     {
                         var id = "myactor" + rnd.Next(0, actorCount);
-                        var request = SendRequest(cluster, id, ct , ctx);
+                        var request = SendRequest(cluster, id, CancellationToken.None , ctx);
 
                         requests.Add(request);
                     }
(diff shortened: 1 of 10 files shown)


benchmarks/ClusterBenchmark/Program.cs
```diff
@@ -228,14 +234,16 @@ async Task RunBatch(Random? rnd, Cluster cluster)
...
                 try
                 {
                     var ct = CancellationTokens.FromSeconds(20);
 
+                    var ctx = cluster.System.Root.Batch(batchSize,ct);
                     for (var i = 0; i < batchSize; i++)
                     {
                         var id = "myactor" + rnd.Next(0, actorCount);
-                        var request = SendRequest(cluster, id, ct);
+                        var request = SendRequest(cluster, id, CancellationToken.None , ctx);
```
rogeralsing: unsure on what this is needed for if we use a batch context that already have a ct?

src/Proto.Actor/Future/Futures.cs
```diff
@@ -55,7 +55,8 @@ internal FutureProcess(ActorSystem system) : base(system)
 
             if (!absent) throw new ProcessNameExistException(name, pid);
 
-            Pid = pid.WithRequestId(1);
+            pid.RequestId = 1;
```
rogeralsing, line 58: seems to have some slight impact. not pushing GC too hard by creating new instances and ignoring the old

benchmarks/ClusterBenchmark/Program.cs
mhelleborg: The batch future should actually check if it is the same token as set on the parent batch process. So theoretically the same token passed to each future should not affect the outcome.

PR #1016, opened 2021-06-01
Heartbeat configuration

pblachut:
## Description

Adds possibility to configure heartbeatinterval and heartbeattimeout. Currently heartbeattimeout is hardcoded.


rogeralsing:
Looks great! 
We need the CLA signed before merging (see pending check for PR)

Issue #1024, opened 2021-06-07
New API allowing message headers to be sent and consumed more easily from application code


mhelleborg:
Use cases for headers from application level code is not very well supported with the current request abstractions.

This could be improved upon. Let's use this issue to discuss possible ways to support it.

rogeralsing:
Partially solved, we need to make this possible to use for ClusterContext also.

mhelleborg:
Fixed for cluster requests in #1411

PR #1025, opened 2021-06-07
Added ability to access and send message headers in requests

mhelleborg:
Alternative to solve #1024, and support richer use cases for headers. Breaking change for future-tasks, as they no longer unpack message envelopes.

### DIFF
--- a/src/Proto.Actor/Context/ActorContext.cs
+++ b/src/Proto.Actor/Context/ActorContext.cs
@@ -65,8 +65,6 @@ namespace Proto.Context
             if (_messageOrEnvelope is not null) EnsureExtras().Stash.Push(_messageOrEnvelope);
         }
 
-        public void Respond(object message, MessageHeader header) => Respond(new MessageEnvelope(message, null, header));
-        
         public void Respond(object message)
         {
             if (Sender is not null)
--- a/src/Proto.Actor/Context/ActorContext.cs
+++ b/src/Proto.Actor/Context/ActorContext.cs
@@ -65,6 +65,8 @@ namespace Proto.Context
             if (_messageOrEnvelope is not null) EnsureExtras().Stash.Push(_messageOrEnvelope);
         }
 
+        public void Respond(object message, MessageHeader header) => Respond(new MessageEnvelope(message, null, header));
+        
         public void Respond(object message)
         {
             if (Sender is not null)
@@ -115,7 +117,7 @@ namespace Proto.Context
             if (duration == ReceiveTimeout) return;
 
             ReceiveTimeout = duration;
-            
+
             EnsureExtras();
             _extras!.StopReceiveTimeoutTimer();
 
@@ -159,7 +161,7 @@ namespace Proto.Context
                     break;
             }
         }
-        
+
         public void Request(PID target, object message, PID? sender)
         {
             var messageEnvelope = new MessageEnvelope(message, sender);
@@ -284,7 +286,7 @@ namespace Proto.Context
                 var res = System.Config.DiagnosticsSerializer(Actor!);
                 diagnosticsString += res;
             }
-            
+
             processDiagnosticsRequest.Result.SetResult(diagnosticsString);
 
             return default;
@@ -422,7 +424,7 @@ namespace Proto.Context
         private Task HandleAutoRespond(IAutoRespond autoRespond)
         {
             // receive normally
-            var res =  Actor!.ReceiveAsync(_props.ContextDecoratorChain is not null ? EnsureExtras().Context : this);
+            var res = Actor!.ReceiveAsync(_props.ContextDecoratorChain is not null ? EnsureExtras().Context : this);
             //then respond automatically
             var response = autoRespond.GetAutoResponse();
             Respond(response);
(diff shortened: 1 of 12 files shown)


mhelleborg:
We could possibly drop the MessageHeader parameter to `RequestWithHeadersAsync`, since it could just take a MessageEnvelope instead.  It would be less obvious to newcomers, but then again, they likely do not care about headers :)

src/Proto.Actor/Context/ActorContext.cs
```diff
@@ -65,6 +65,8 @@ public void Stash()
             if (_messageOrEnvelope is not null) EnsureExtras().Stash.Push(_messageOrEnvelope);
         }
 
+        public void Respond(object message, MessageHeader header) => Respond(new MessageEnvelope(message, null, header));
```
rogeralsing: 👍🏻

src/Proto.Actor/Context/IContext.cs
```diff
@@ -33,6 +33,13 @@ public interface IContext : ISenderContext, IReceiverContext, ISpawnerContext, I
         /// </summary>
         /// <param name="message">The message to send</param>
         void Respond(object message);
+        
+        /// <summary>
+        ///     Sends a response to the current Sender, including message header
+        /// </summary>
+        /// <param name="message">The message to send</param>
+        /// <param name="header"></param>
+        void Respond(object message, MessageHeader header) => Respond(new MessageEnvelope(message, null, header));
```
rogeralsing, line 42: 👍🏻

src/Proto.Actor/Context/ISenderContext.cs
```diff
@@ -102,23 +102,65 @@ public static class SenderContextExtensions
...
+        )
+        {
+            using var future = self.GetFuture();
+            var messageEnvelope = new MessageEnvelope(message, future.Pid, headers);
+            self.Send(target, messageEnvelope);
+            var result = await future.GetTask(cancellationToken);
+
+            var messageResult = MessageEnvelope.UnwrapMessage(result);
+
+            switch (messageResult)
```
rogeralsing, line 129: We have this switch a bit all over the place afaik. can it be moved to a shared method?

src/Proto.Actor/Context/ISenderContext.cs
```diff
@@ -102,23 +102,65 @@ public static class SenderContextExtensions
...
+                        $"Unexpected message. Was type {messageResult.GetType()} but expected {typeof(T)}"
+                    );
+            }
+        }
+
+        internal static async Task<T> RequestAsync<T>(this ISenderContext self, PID target, object message, CancellationToken cancellationToken)
         {
             using var future = self.GetFuture();
-            var messageEnvelope = new MessageEnvelope(message, future.Pid);
+            var messageEnvelope = message is MessageEnvelope envelope ? envelope.WithSender(future.Pid) : new MessageEnvelope(message, future.Pid);
```
rogeralsing, line 144: Is this to allow us to explicitly pass MessageEnvelopes as a message? (I assume yes, but just to be 100% :) )

src/Proto.Actor/Context/ISenderContext.cs
```diff
@@ -102,23 +102,65 @@ public static class SenderContextExtensions
...
             self.Send(target, messageEnvelope);
             var result = await future.GetTask(cancellationToken);
 
-            switch (result)
+            var messageResult = MessageEnvelope.UnwrapMessage(result);
+
+            switch (messageResult)
```
rogeralsing, line 150: See above comment on repeated code

src/Proto.Actor/Context/ISenderContext.cs
```diff
@@ -102,23 +102,65 @@ public static class SenderContextExtensions
...
+
+            var messageResult = MessageEnvelope.UnwrapMessage(result);
+
+            switch (messageResult)
+            {
+                case DeadLetterResponse:
+                    throw new DeadLetterException(target);
+                case null:
+                case T:
+                    return ((T) messageResult!, MessageEnvelope.UnwrapHeader(result));
```
rogeralsing, line 133: My bad regarding repeated code, this bit is ofc different.

rogeralsing:
>We could possibly drop the MessageHeader parameter to RequestWithHeadersAsync

I like the consistency, if we get message + headers back from the task, it makes sense to pass in message + headers into the request

src/Proto.Actor/Context/ISenderContext.cs
mhelleborg: It is indeed. Added a test for it as well

src/Proto.Actor/Messages/MessageEnvelope.cs
```diff
@@ -35,6 +38,16 @@ public MessageEnvelope(object message, PID? sender, MessageHeader? header = null
         public MessageEnvelope WithHeader(MessageHeader header) =>
             this with {Header = header};
 
+        public MessageEnvelope MergeHeader(MessageHeader header)
```
rogeralsing, line 41: 👍🏻

PR #1026, opened 2021-06-08
Gossip play 2

rogeralsing:
Removing old n*n broadcast in favor for gossip notifications


src/Proto.Cluster/Cluster.cs
```diff
@@ -81,9 +81,10 @@ public Cluster(ActorSystem system, ClusterConfig config)
 
         public async Task StartMemberAsync()
         {
+            //gossiper must be started whenever any topology events starts flowing
+            await Gossip.StartAsync();
```
rogeralsing, line 85: This was one of the bugs. the gossip actor ofc needs to be started before registration to process events

PR #1028, opened 2021-06-11
API to get cluster state

rogeralsing:
API and test to showcase shared cluster state


src/Proto.Cluster/Gossip/GossipActor.cs
```diff
@@ -24,11 +25,30 @@ public class GossipActor : IActor
...
+        private Task OnGetGossipStateKey(IContext context, GetGossipStateRequest getState)
+        {
+            var entries = ImmutableDictionary<string, Any>.Empty;
+            var key = getState.Key;
+            
+            foreach (var (memberId, memberState) in _state.Members)
+            {
+                if (memberState.Values.TryGetValue(key, out var value))
+                {
+                    entries = entries.SetItem(memberId, value.Value);
```
rogeralsing, line 43: for each member in state, try pick the given state key, and add it to the result

PR #1093, opened 2021-07-13
Generate void / parameterless grain methods

### LINKED ISSUE #1092: Allow for generation of void / parameterless grain methods
Even grain methods with no real return values or parameters currently needs to pass a surrogate value to and from the methods. This is part of the proto service rpc spec they are generated from.

In order to make the grains themselves nicer to work with, these can be omitted from the generated classes, and just present normal void / parameterless methods.

Google.Protobuf.WellKnownTypes.Empty can be used as a marker type for the parameter / return values, and it should be an idiomatic way to represent this from the proto side.

mhelleborg:
Closes #1092 
Generate void / parameterless grain methods when return value or parameter is Google.Protobuf.WellKnownTypes.Empty.

Enabled regression tests for codegen.

### DIFF
--- a/tests/Proto.Cluster.CodeGen.Tests/ProtoGrainGenerationTests.cs
+++ b/tests/Proto.Cluster.CodeGen.Tests/ProtoGrainGenerationTests.cs
@@ -1,7 +1,6 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using FluentAssertions;
 using Google.Protobuf.Reflection;
 using ProtoBuf.Reflection;
 using Xunit;
@@ -33,7 +32,7 @@ namespace Proto.Cluster.CodeGen.Tests
             }
 
             var expectedOutput = File.ReadAllText(expectedOutputFile).Trim();
-            res.Single().Text.Trim().Should().Be(expectedOutput);
+            Assert.Equal(expectedOutput, res.Single().Text.Trim());
         }
     }
 }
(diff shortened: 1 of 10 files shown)


tests/Proto.Cluster.CodeGen.Tests/ProtoGrainGenerationTests.cs
```diff
@@ -35,6 +32,8 @@ public void CanFindImportedNamespaces()
                 _testOutputHelper.WriteLine(codeFile.Text);
             }
 
+            var expectedOutput = File.ReadAllText(expectedOutputFile).Trim();
+            res.Single().Text.Trim().Should().Be(expectedOutput);
```
rogeralsing: it might be a bit too strict to test for exact whitespace, indentation etc.
touching either the template or expected output will fail the tests.
XUnit is pretty good at pointing out where strings differ though, so maybe not a big issue

mhelleborg: The idea was to have a guarantee against regressions. It will require us to change the expected result any time we change the generator output, but that is kind of the point :)

It would be nice to not have it so dependent on exact white space etc, do you have a suggestion for a better approach? I did consider something like parsing it with Roslyn and just checking the AST, but that is more work for something that might not give a better outcome overall.

rogeralsing: I was thinking something like remove all WS and compare the output. if that passes, but not the raw strings, then we know its just a whitespace issue, which would make it easier to diagnose at least. but might be overkill

PR #1108, opened 2021-07-30
Removes the check and logging of k8s port matching

### LINKED ISSUE #1107: Kubernetes provider throws ArgumentNullException when running with sidecar containers in pod
image

This code does not take into account there being more than one container in a pod - nor any of those containers not having port definitions. However this method is only used in `KuberentesProvider::RegisterMemberInner` to log a warning if null:

image

I think this code can be removed without impact - alternatively the `FindPort` extension could be modified to be more resilient.

## Expected Behavior

When I have sidecars running alongside the application without port definitions, then the cluster should be able to start as normal

## Actual Behavior
 An `ArgmentNullException` is thrown and the nodes are unable to start the cluster

## Steps to Reproduce the Problem

  1. Create a basic ActorSystem with KubernetesProvider cluster config
  2. Attempt to call `StartMemberAsync()` within a pod that has at least one container without port definitions
  
_Note: the container without ports must be the first element returned in `pod.Spec.Containers`_  

## Specifications

  - Version: 0.21.0
  - Platform: mcr.microsoft.com/dotnet/core/aspnet:3.1
  - Subsystem:

## StackTrace
<details>
System.ArgumentNullException: Value cannot be null. (Parameter 'source')
  Module "System.Linq.ThrowHelper", in ThrowArgumentNullException
  Module "System.Linq.Enumerable", in TryGetFirst
  Module "System.Linq.Enumerable", in FirstOrDefault
  Module "Proto.Cluster.Kubernetes.KubernetesExtensions", in FindPort
  Module "Proto.Cluster.Kubernetes.KubernetesProvider", in RegisterMemberInner
  Module "System.Runtime.ExceptionServices.ExceptionDispatchInfo", in Throw
  Module "System.Runtime.CompilerServices.TaskAwaiter", in ThrowForNonSuccess
  Module "Proto.Utils.Retry", in Try
</details>

aaronjamesford:
## Description

Closes #1107 

Removes the check and logging of k8s declared ports vs the port of the actor system

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/src/Proto.Cluster.Kubernetes/KubernetesExtensions.cs
+++ b/src/Proto.Cluster.Kubernetes/KubernetesExtensions.cs
@@ -22,15 +22,6 @@ namespace Proto.Cluster.Kubernetes
         private static string cachedNamespace;
 
         /// <summary>
-        ///     Find the container port for a given pod than matches the given port.
-        /// </summary>
-        /// <param name="pod">Kubernetes Pod object</param>
-        /// <param name="port">Port to find in container ports</param>
-        /// <returns></returns>
-        internal static V1ContainerPort FindPort(this V1Pod pod, int port)
-            => pod.Spec.Containers[0].Ports.FirstOrDefault(x => x.ContainerPort == port);
-
-        /// <summary>
         ///     Replace pod labels
         /// </summary>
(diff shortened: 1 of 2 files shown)


rogeralsing:
Many thanks, merging this!

Issue #431, opened 2018-09-27
Get rid of static/singletons ?


jrouaix:
Hello @rogeralsing, @cpx86 

Can I ask why you design such static/singletons stuffs ?

I was the Remote unit test, confirming that the only way to test such thing is to run another process.

Don't you think it could be better if we get rid if all these singletons and encapsulate them into some root "ActorSystem" instance, so we can start mutliple ActorSystem in a unit test more easily ? 

Would it then possible to have multiple GRPC endpoint (associated with distinct/isolated ActorSystems) again in order to easier Remote tests ?

cpx86:
I've thought about changing that as well. I think Roger and I have discussed it at some point, but tbh I can't remember why we decided not to change it 😅 

I'd say that if we do change it then it should be for solving an important use case, e.g. if we want to have support for isolating actors or exposing multiple GRPC endpoints within a single process, and not out of language/design purism.

rogeralsing:
TLDR;

if you have instances of remoting, then the serialization layer needs to understand what instance it should deserialize its messages for, context aware messages.
e.g. an PID. 

This puts an enormous burden on the serializer and forces you to use custom serialization schemes.

rogeralsing:
Thinking about it, now with the new API with context.Send. this might not be an issue as the context will know the context so to say :-)

jrouaix:
Ok, it's on my "will do if time" list.

cpx86:
@rogeralsing Not sure if you remember exactly what you were thinking 256 days ago 😄 But how would we go about this more precisely? Should we just move e.g. the static logger factory to the context instance? In that case what about remoting? Would we have a "remote context" inheriting from the root context but adding the remote-specific instances like the server, serialization config, internal processes, etc?

rogeralsing:
Let's see if I remember the reasoning here.

Deserilizing ActorRef in Akka.NET was a pain as it needs to be attached to the ActorSystem it belonged to.
So instead, ProtoActor went for a simpler approach where PID is like a phone number, one dont need to know anything more about its content or context.

This lead to the static API.
PID.Tell/Send would know where to forward the message by the processregistry.

But, now that Tell has moved onto the context, e.g RootContext, ActorContext.
We already have all the information needed for passing the message to the correct instances.

e.g. if the process registry was not static, but an instance.
as long as we could share that instance between RootContexts and ActorContexts, everything would be fine.
The process registry would be sort of the ActorSystem in Akka.
[...]

PR #1115, opened 2021-08-06
WiP: Cluster Singletons - Keep alive

rogeralsing:
WiP for #431 

scans all keys in the cluster gossip state. finds all of the keys for currently active singletons.
compares it to the list of actors that should exist.
any actor that doesn't exist, gets reactivated via Touch message


yeongjonglim:
Hi @rogeralsing, I am looking to have this feature as well, is there any chance this PR will be merged any time soon?

PR #1123, opened 2021-08-11
Modify the syntax for `Proto.Cluster.CodeGen` to be compatible with C# version 8.0

gsw945:
Modify the syntax to be compatible with C# version 8


rogeralsing:
👍🏻

PR #1173, opened 2021-09-07
Deprecated ClusterInit

### LINKED ISSUE #1172: Obsolete ClusterInit
The initial design for cluster actor initialization called for receiver middleware which injected a special `ClusterInit` message including the actor identity and kind, as well as a reference to the cluster itself.

With the improvements of the plugin system and the highly performant context store, this is no longer necessary. 

Instead of listening to ClusterInit, the actor can just use the `Started` lifecycle message to bootstrap the actor. Both `.Cluster()` and `.ClusterIdentity()` is available directly from the `IContext` parameter.

mhelleborg:
Closes #1172, deprecates `ClusterInit` as virtual actor initialization method.


rogeralsing:
LGTM!

PR #1254, opened 2021-10-21
Improve Partition Identity performance

mhelleborg:
Rewrite of both HashRing and Partition Identity handover.

Breaking change with regards to Rendezvous algorithm for identity ownership, so can not run with older versions of Proto.Actor.

### DIFF
--- a/src/Proto.Actor/Router/HashRing.cs
+++ b/src/Proto.Actor/Router/HashRing.cs
@@ -1,4 +1,4 @@
-// -----------------------------------------------------------------------
+﻿// -----------------------------------------------------------------------
 // <copyright file="HashRing.cs" company="Asynkron AB">
 //      Copyright (C) 2015-2020 Asynkron AB All rights reserved
 // </copyright>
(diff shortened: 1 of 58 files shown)


src/Proto.Cluster/Partition/MemberHashRing.cs
```diff
@@ -0,0 +1,22 @@
...
+// -----------------------------------------------------------------------
+using System.Collections.Generic;
+using Proto.Router;
+
+namespace Proto.Cluster.Partition
+{
+    public class MemberHashRing : HashRing<Member>
+
+    {
+        public MemberHashRing(IReadOnlyCollection<Member> nodes) : base(nodes, member => member.Address, MurmurHash2.Hash, 50)
```
mhelleborg, line 14: Replicacount could perhaps be tuned, not sure what would be the best tradeoff here..

src/Proto.Cluster/Partition/PartitionIdentityRelocationWorker.cs
```diff
@@ -0,0 +1,176 @@
...
+            Logger.LogError("Relocation timed out");
+            _onRelocationComplete.TrySetResult(false);
+            return Task.CompletedTask;
+        }
+
+        private Task OnDeadLetterResponse(DeadLetterResponse response, IContext context)
+        {
+            if (_waitingRequests.Remove(response.Target))
+            {
+                Logger.LogError("Unreachable node: {Address}", response.Target.Address);
```
mhelleborg: How is the best way to handle network partitions between members? If this is from a member which should still be in the cluster, we would be in an inconsistent state.

mhelleborg: More likely than network partition is that the member has already shut down though, which means that the node will need to rebalance again, and we will end up in a correct state.

src/Proto.Cluster/Partition/PartitionIdentityActor.cs
```diff
@@ -93,40 +93,12 @@ private async Task OnClusterTopologyInner(ClusterTopology msg, IContext context)
 
             requestMsg.Members.AddRange(members);
```
cpx86: A bit of a nit-pick, but this could be `msg.Members` and the `members` var could be scrapped, right?

mhelleborg: Yes, would be a little cleaner. Fixing :)

src/Proto.Actor/Router/HashRing.cs
```diff
@@ -9,33 +9,91 @@
...
+        }
+
+        public T GetNode(string key)
+        {
+            if (_ring.Length == 0) return default!;
+            if (_replicaCount == _ring.Length) return _ring[0].value;
+
+            var hash = _hash(key);
+
+            var result = Array.BinarySearch(_hashes, hash);
```
rogeralsing, line 48: Nice!

src/Proto.Cluster/Partition/MemberHashRing.cs
rogeralsing: My understanding of the stickiness is that the more replicas, the bigger the risk of not being sticky. e.g. more parts of the hash ring are moving.
So some config might be nice

src/Proto.Cluster/Partition/PartitionIdentityRelocationWorker.cs
rogeralsing: Any member that become invisible to other members due to network partitioning, would be evicted from the cluster.
We have a fairly different approach than e.g. Akka which tries to "fix" things.
But in the era of let it crash, containers and pods.
Just let that node shut down, and restart

rogeralsing (commented):
All in all, looks great. having config for hashring replicas would be nice though

src/Proto.Cluster/Partition/MemberHashRing.cs
mhelleborg: The one potential problem with making it configurable is that two different configs can have different answers for which member an identity belongs to. If both configurations exists in a cluster, the behavior would be incorrect, and you could end up with lots of weird failure cases.  It might be slightly less error prone to choose a constant default value. 
The downside would be for someone who would like to tune the  distribution / stickyness. An option could be to make the partition identity check if the other nodes have the same settings as itself, and otherwise fail fast on startup.

cpx86: > make the partition identity check if the other nodes have the same settings

The same problem would occur if we were to change the constant replica count between versions, so maybe this is something we'd want to handle either way?

This got me thinking, is there any way to handle a change to the replica count gracefully? If all nodes must be in agreement on the replica count, then as long as there's one node alive then the value can't change. Meaning the only way to change it would be to shut down the whole cluster first, right?

mhelleborg: That is actually a very good point. The first node active in the cluster could be the master for replica count, and additional nodes would use that instead of any local settings. Then it can be tuned without worrying about it being out of sync across the cluster.

cpx86: A question then is what happens when the leader decides to change the replica count. Would the cluster become in sync quickly enough to avoid disruptions? Or would it be necessary to have some sort of coordinated pause until all nodes use the same value?

mhelleborg: I definitely don't think that needs to be dynamic at all. Keep the initial value from starting the cluster, no problem with updates.

rogeralsing: Is this a likely scenario though?
Under what conditions would a node decide that hey, let's tweak the replicas of the hash ring?

cpx86: Let me rephrase, because I botched my previous comment a bit 😅 

If 1) the replicaCount value changes (either because the const value changed between Proto.Cluster versions, or because it was changed via config) and 2) nodes always check what the current value in the cluster is (controlled by the oldest node), then wouldn't it mean the only way to change the value is to shut down all nodes? Even if you bring it down to a single-node cluster, that node would already be configured to use the old value. Or am I missing something?

mhelleborg: No, I think you summarized it :) The cluster will have to be shut down if you want to change replica count, if we go for this solution where we allow the first node to decide the setting.
It then later being removed from the cluster cannot change it, as that would break the behavior between the rest of the nodes.

rogeralsing:
@mhelleborg , have this been tested with the cluster benchmark ?
otherwise I can give that part a go later

mhelleborg:
> @mhelleborg , have this been tested with the cluster benchmark ? otherwise I can give that part a go later

Not yet. I do have  a few more changes I think should be included, but they should not affect performance. You can have a go testing it, but don't merge it yet :)

mhelleborg:
In order to better handle compatibility between versions and potentially different configs, I think a handshake between nodes on startup would be helpful. Then they can verify that they are running with compatible configurations and within a compatible version range.  
Instead of running with potentially undefined behavior, it could instead stop before causing any problems.

src/Proto.Cluster/Gossip/Gossiper.cs
```diff
@@ -97,9 +127,148 @@ private async Task GossipLoop()
             }
         }
 
+        public class ConsensusCheckBuilder<T>
```
mhelleborg: Not married to the name..

PR #1259, opened 2021-10-27
Dependabot should ignore CouchbaseNetClient 3.x 

cpx86:
CouchbaseNetClient 3.x is a breaking version, so it can't be automatically updated by Dependabot.


cpx86:
Forgot to change to dev branch before making the change 😅

PR #1262, opened 2021-10-30
RequestReenter and ClusterRequestReenter

rogeralsing:
Sugar API to make it easier to combine request and reenter together

### DIFF
--- a/src/Proto.Cluster/ClusterExtension.cs
+++ b/src/Proto.Cluster/ClusterExtension.cs
@@ -37,6 +37,13 @@ namespace Proto.Cluster
             var task = context.System.Cluster().RequestAsync<T>(identity, kind, message, context, ct);
             context.ReenterAfter(task, callback);
         }
+        
+        public static void ClusterRequestReenter<T>(this IContext context, ClusterIdentity clusterIdentity, object message, Func<Task<T>, Task> callback, CancellationToken ct)
+        {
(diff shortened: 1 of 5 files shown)


src/Proto.Cluster/ClusterExtension.cs
```diff
@@ -30,6 +30,13 @@ public static Cluster Cluster(this IContext context)
         public static Task<T> ClusterRequestAsync<T>(this IContext context, string identity, string kind, object message, CancellationToken ct) =>
             //call cluster RequestAsync using actor context
             context.System.Cluster().RequestAsync<T>(identity, kind, message, context, ct);
+        
+        public static void ClusterRequestReenter<T>(this IContext context, string identity, string kind, object message, Func<Task<T>, Task> callback, CancellationToken ct)
```
mhelleborg, line 34: Use the ClusterIdentity overload?

mhelleborg: Then it could possibly have the same name as the other RequestReenter. In addition re-using ClusterIdentity is potentially more performant, as it can cache the activation pid internally.

rogeralsing: Added CI overload. but don't like the same name overload as we make the distinction everywhere else based on names. e.g. having RequestAsync and ClusterRequestAsync, but only RequestReenter seems inconsistent

PR #1289, opened 2021-11-17
Client remote

Damian-P:
## Description
Here is a proposition to address this issue : #444 .
This PR adds ClientRemote which connects to ServerRemotes using bi-directional streaming.

When a ClientRemote sends a message to a server node, it translates sender PID in the message envelope  ($client/GUID/id) to a proxy PID ( serverAddress:serverPort/$client/GUID/id). This way the server can forward theses messages to other nodes and they can respond to the client using this node as a Proxy.

There is a breaking change, every nodes would use only one grpc service method which handles connection, disconnection and message batches. This way we can identify a connection in the EndpointReader. The compatibility would be broken with the go version until it's aligned with our proto file.

```protobuf
message RemoteMessage {
  oneof message_type {
    MessageBatch message_batch = 1;
    ConnectRequest connect_request = 2;
    ConnectResponse connect_response = 3;
    DisconnectRequest disconnect_request = 4;
  }
}

message MessageBatch {
  repeated string type_names = 1;
  repeated actor.PID targets = 2; // Can be a proxy pid
  repeated MessageEnvelope envelopes = 3;
  repeated actor.PID senders = 4; // Can be a proxy pid
}

message MessageEnvelope {
  int32 type_id = 1;
  bytes message_data = 2;
  int32 target = 3;
  int32 sender = 4; // Index starting at 1 allows to pass messages without a sender (sender = 0)
  int32 serializer_id = 5;
  MessageHeader message_header = 6;
  uint32 request_id = 7;
}

message MessageHeader {
  map<string, string> header_data = 1;
}

message ConnectRequest {
  oneof connection_type {
    ClientConnection client_connection = 1;
    ServerConnection server_connection = 2;
  }
}

message DisconnectRequest {

}

message ClientConnection {
  string SystemId = 1;
}

message ServerConnection {
  string SystemId = 1;
  string Adress = 2;
}

message ConnectResponse {
  //int32 default_serializer_id = 1; Obsoleted
  string member_id = 2;
  bool blocked = 3;
}

service Remoting {
  rpc Receive (stream RemoteMessage) returns (stream RemoteMessage) {}
  rpc ListProcesses(ListProcessesRequest) returns (ListProcessesResponse) {}
  rpc GetProcessDiagnostics(GetProcessDiagnosticsRequest) returns (GetProcessDiagnosticsResponse) {}
}
```

## Purpose
This pull request is a:

## Checklist


sudsy:
This looks very interesting. I will put aside some time next week to review this.

rogeralsing:
Fantastic! Will dive into this later today

sudsy:
This looks good, but I'm sure I will have more comments once I have been able to run it up in my infrastructure.

I can see that the client resolvers #594 have been removed, but that isn't an issue for me any more, I ended up using an alternative strategy for my clients that doesn't depend upon the client resolver. 

One thing that has always been an issue with my client implementation is that the clients are short lived and generally much less stable. I think the following questions represent one of the more annoying edge cases, but unfortunately it is fairly common.

- Client connection dies mid stream - what happens next? Does the client try to reconnect? If so, perhaps it couldn't connect back to the original host and instead connects to another host in the cluster. 
- If the client reconnects to another host in the cluster can the processes in the cluster send messages to the old address and still find it?
- If hosts can't find the client after reconnect, have they been notified that the client is effectively dead through a termination message?
- After client reconnect, can the client continue to send messages to pid's that it may have cached on the client side and expect a response (ie, will the proxy addressing update properly without the client needing to be aware)

These scenarios are hard to write tests for, but I would want to be very clear on what the approach is under this type of scenario before deploying against my workload. It may be that reconnecting to a different host is just not supported, but we need to be explicit about that. It certainly has implications for how you advertise the address of the host service to clients (limiting how you might load balance them).

One other possible scenario that has been in the back of my mind. Once clients are properly supported, it opens up the possibility of exposing the client service over [grpc-web](https://docs.microsoft.com/en-us/aspnet/core/grpc/browser?view=aspnetcore-6.0#configure-grpc-web-in-aspnet-core).

I can imagine a scenario where you might want to expose just the client endpoint over grpc-web (for authenticated, but public internet access vie a future protoactor-js implementation), but keep the Remote <-> Remote communications on a different port with different security requirements. I couldn't really see whether that would be supported with this refactor. It looks like it just needs to be a seperate service to support this scenario. Perhaps this is something to defer to a future refactor.

rogeralsing:
>Client connection dies mid stream - what happens next? Does the client try to reconnect? If so, perhaps it couldn't connect back to the original host and instead connects to another host in the cluster.
>If the client reconnects to another host in the cluster can the processes in the cluster send messages to the old address and still find it?
>If hosts can't find the client after reconnect, have they been notified that the client is effectively dead through a termination message?
>After client reconnect, can the client continue to send messages to pid's that it may have cached on the client side and expect a response (ie, will the proxy addressing update properly without the client needing to be aware)

Excellent questions!

I will have to digest these and see where I stand myself on how this could/should work
cc @mhelleborg

rogeralsing:
Merged here https://github.com/asynkron/protoactor-dotnet/pull/1303

PR #1290, opened 2021-11-21
OpenTelemetry tracing

mhelleborg:
## Description

Ported Proto.OpenTracing to use System.Diagnostics ActivitySource with  OpenTelemetry conventions and span propagation.

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/src/Proto.OpenTelemetry/OpenTelemetryDecorators.cs
+++ b/src/Proto.OpenTelemetry/OpenTelemetryDecorators.cs
@@ -81,7 +81,7 @@ namespace Proto.OpenTelemetry
         public static void Send(PID target, object message, ActivitySetup sendActivitySetup, Action send)
         {
             using var activity =
-                OpenTelemetryHelpers.BuildStartedScope(Activity.Current?.Context ?? default, nameof(Send), message, sendActivitySetup);
+                OpenTelemetryHelpers.BuildStartedActivity(Activity.Current?.Context ?? default, nameof(Send), message, sendActivitySetup);
 
             try
             {
@@ -100,7 +100,7 @@ namespace Proto.OpenTelemetry
         internal static void Request(PID target, object message, ActivitySetup sendActivitySetup, Action request)
         {
             using var activity =
-                OpenTelemetryHelpers.BuildStartedScope(Activity.Current?.Context ?? default, nameof(Request), message, sendActivitySetup);
+                OpenTelemetryHelpers.BuildStartedActivity(Activity.Current?.Context ?? default, nameof(Request), message, sendActivitySetup);
 
             try
             {
@@ -119,7 +119,7 @@ namespace Proto.OpenTelemetry
         internal static void Request(PID target, object message, PID? sender, ActivitySetup sendActivitySetup, Action request)
         {
             using var activity =
-                OpenTelemetryHelpers.BuildStartedScope(Activity.Current?.Context ?? default, nameof(Request), message, sendActivitySetup);
+                OpenTelemetryHelpers.BuildStartedActivity(Activity.Current?.Context ?? default, nameof(Request), message, sendActivitySetup);
 
             try
             {
@@ -144,7 +144,7 @@ namespace Proto.OpenTelemetry
         internal static async Task<T> RequestAsync<T>(PID target, object message, ActivitySetup sendActivitySetup, Func<Task<T>> requestAsync)
         {
             using var activity =
-                OpenTelemetryHelpers.BuildStartedScope(Activity.Current?.Context ?? default, nameof(Request), message, sendActivitySetup);
+                OpenTelemetryHelpers.BuildStartedActivity(Activity.Current?.Context ?? default, nameof(Request), message, sendActivitySetup);
 
             try
             {
@@ -163,7 +163,7 @@ namespace Proto.OpenTelemetry
         internal static void Forward(PID target, object message, ActivitySetup sendActivitySetup, Action forward)
         {
             using var activity =
-                OpenTelemetryHelpers.BuildStartedScope(Activity.Current?.Context ?? default, nameof(Forward), message, sendActivitySetup);
+                OpenTelemetryHelpers.BuildStartedActivity(Activity.Current?.Context ?? default, nameof(Forward), message, sendActivitySetup);
 
             try
             {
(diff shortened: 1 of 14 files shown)


src/Proto.OpenTelemetry/OpenTelemetryDecorators.cs
```diff
@@ -0,0 +1,207 @@
...
+            => OpenTelemetryMethodsDecorators.Receive(envelope, _receiveActivitySetup, () => base.Receive(envelope));
+    }
+
+    static class OpenTelemetryMethodsDecorators
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void Send(PID target, object message, ActivitySetup sendActivitySetup, Action send)
+        {
+            using var activity =
+                OpenTelemetryHelpers.BuildStartedScope(Activity.Current?.Context ?? default, nameof(Send), message, sendActivitySetup);
```
marcinbudny: How about using activity names like `proto.send {messageType}` etc. This way it is easy to distinguish proto spans from e.g. spans related to message brokers.

src/Proto.OpenTelemetry/OpenTelemetryHelpers.cs
```diff
@@ -0,0 +1,54 @@
...
+            string verb,
+            object message,
+            ActivitySetup activitySetup,
+            ActivityKind activityKind = ActivityKind.Internal
+        )
+        {
+            var messageType = message?.GetType().Name ?? "Unknown";
+
+            var name = $"{verb} {messageType}";
+            var tags = new[] {new KeyValuePair<string, object?>(ProtoTags.MessageType, messageType)};
```
marcinbudny, line 30: How about including `net.host.name` tag with value from `Dns.GetHostName()`? Then we know which host the sending / receiving actor is on. Or should this be added to all outgoing traces by some middleware on tracing pipe?

BTW the name `net.host.name` comes from [SemanticConventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/README.md)

src/Proto.OpenTelemetry/OpenTelemetryHelpers.cs
```diff
@@ -0,0 +1,54 @@
...
+            var activity = ActivitySource.StartActivity(name, activityKind, parent, tags);
+
+            activitySetup.Invoke(activity, message!);
```
marcinbudny: Check activity for null?

src/Proto.OpenTelemetry/OpenTelemetryHelpers.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Runtime.CompilerServices;
+using OpenTelemetry;
+using OpenTelemetry.Context.Propagation;
+
+namespace Proto.OpenTelemetry
+{
+    static class OpenTelemetryHelpers
+    {
+        private static readonly ActivitySource ActivitySource = new ActivitySource(ProtoTags.ActivityName);
+        private static readonly TextMapPropagator Propagator = new TraceContextPropagator();
```
marcinbudny: `Propagators.DefaultTextMapPropagator`?

It includes baggage as well https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry/Sdk.cs#L32

src/Proto.OpenTelemetry/OpenTelemetryExtensions.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System.Diagnostics;
+
+namespace Proto.OpenTelemetry
+{
+    public delegate void ActivitySetup(Activity? activity, object message);
+
+    public static class OpenTelemetryExtensions
```
marcinbudny, line 8: How about extension that adds source to `TracerProviderBuilder`

```csharp
public static TracerProviderBuilder AddProtoActorInstrumentation(this TracerProviderBuilder builder) 
    => builder.AddSource(ProtoTags.ActivityName);
```

src/Proto.OpenTelemetry/ProtoTags.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace Proto.OpenTelemetry
+{
+    public static class ProtoTags
+    {
+        public const string ActivityName = "Proto.Actor";
```
marcinbudny: `ActivitySourceName`?

src/Proto.OpenTelemetry/OpenTelemetryHelpers.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public static void DefaultSetupActivity(Activity _, object __)
+        {
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static Activity? BuildStartedScope(
```
marcinbudny: `BuildStartedActivity` ?

src/Proto.OpenTelemetry/OpenTelemetryHelpers.cs
mhelleborg: We already know from the PID which host it is on, so not sure that is needed. I don't want to unnecessarily bloat the traces

src/Proto.OpenTelemetry/OpenTelemetryHelpers.cs
mhelleborg: Sure

src/Proto.OpenTelemetry/OpenTelemetryHelpers.cs
mhelleborg: ActivitySetup already has activity as nullable. Not sure if there are any point to calling it if it is actually null though

src/Proto.OpenTelemetry/OpenTelemetryDecorators.cs
mhelleborg: Certainly, we can make the operation names a bit more namespaced

src/Proto.OpenTelemetry/OpenTelemetryExtensions.cs
mhelleborg: Good idea, added

src/Proto.OpenTelemetry/OpenTelemetryDecorators.cs
```diff
@@ -0,0 +1,207 @@
...
+            IContext context,
+            ActivitySetup sendActivitySetup,
+            ActivitySetup receiveActivitySetup
+        ) : base(context)
+        {
+            var actorType = context.Actor.GetType().Name;
+            var self = context.Self.ToString();
+            _sendActivitySetup = (activity, message) => {
+                activity?.SetTag(ProtoTags.ActorType, actorType);
+                activity?.SetTag(ProtoTags.ActorPID, self);
```
mhelleborg, line 51: I'm not 100% sure we need ActorPID here. The PID is already present in the trace, but depending on the span type, it is either  SenderPid or TargetPid. ActorPid is only to normalize this to a single value

PR #1300, opened 2021-11-24
RemoteProcess PidCache timeout

mhelleborg:
## Description

Background cleanup of RemoteProcess  PidCache items. Runs on an interval and clears RemoteProcesses which has not been used recently.

Introduces 

```
ClusterConfig.RemotePidCacheTimeToLive = TimeSpan.FromMinutes(15);
ClusterConfig.RemotePidCacheClearInterval = TimeSpan.FromSeconds(15);
```

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/src/Proto.Remote/RemoteProcess.cs
+++ b/src/Proto.Remote/RemoteProcess.cs
@@ -4,6 +4,8 @@
 //   </copyright>
 // -----------------------------------------------------------------------
 
+using System.Diagnostics;
+
 namespace Proto.Remote
 {
     public class RemoteProcess : Process
@@ -11,11 +13,13 @@ namespace Proto.Remote
         private readonly EndpointManager _endpointManager;
         private readonly PID _pid;
         private PID? _endpoint;
+        private long _lastUsedTick;
 
         public RemoteProcess(ActorSystem system, EndpointManager endpointManager, PID pid) : base(system)
         {
             _endpointManager = endpointManager;
             _pid = pid;
+            _lastUsedTick = Stopwatch.GetTimestamp();
         }
 
         protected internal override void SendUserMessage(PID _, object message) => Send(message);
@@ -57,6 +61,9 @@ namespace Proto.Remote
             }
 
             System.Root.Send(_endpoint, message);
+            _lastUsedTick = Stopwatch.GetTimestamp();
         }
+
+        internal long LastUsedTick => _lastUsedTick;
     }
 }
--- a/src/Proto.Actor/PID.cs
+++ b/src/Proto.Actor/PID.cs
@@ -39,6 +39,8 @@ namespace Proto
             return _process;
         }
 
+        internal Process? CurrentRef => _process;
+
         internal void SendUserMessage(ActorSystem system, object message)
         {
             var reff = Ref(system) ?? system.ProcessRegistry.Get(this);
--- a/src/Proto.Cluster/Cluster.cs
+++ b/src/Proto.Cluster/Cluster.cs
@@ -31,7 +31,7 @@ namespace Proto.Cluster
 
             system.Extensions.Register(this);
             system.Metrics.Register(new ClusterMetrics(system.Metrics));
-            
+
             //register cluster messages
             var serialization = system.Serialization();
             serialization.RegisterFileDescriptor(ClusterContractsReflection.Descriptor);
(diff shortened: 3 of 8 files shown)


src/Proto.Remote/RemoteProcess.cs
```diff
@@ -4,18 +4,22 @@
...
+using System.Diagnostics;
+
 namespace Proto.Remote
 {
     public class RemoteProcess : Process
     {
         private readonly EndpointManager _endpointManager;
         private readonly PID _pid;
         private PID? _endpoint;
+        private long _lastUsedTick;
```
DenizPiri, line 16: This mixes the usage of Stopwatch.GetTimestamp and TimeSpan ticks. As far as I know, that won't be valid for all cases. Simply storing DateTime instance assigned from DateTime.UtcNow, or using DateTime.UtcNow.Ticks should be fast and reliable.

mhelleborg: I was hoping to avoid the allocation overhead of that :)

Fixed the cutoff calculation instead.

Thanks for spotting it :)

DenizPiri:
This solves the issue only for RemoteProcess instances. When the cluster is in use, all actors on the local machine get ActorProcess instances.

In order to solve the issue described in https://github.com/asynkron/protoactor-dotnet/issues/1301 fully, this approach could be better:
- PidCache Dictionary should hold a struct containing of LastUsedTime + PID instead of just the PID.
- At each TryGet call, it should update that LastUsedTime.
- It should clear up the whole dict as it is done in the first commit.
- RemoteProcess should stay as it was before.
  - Technically that Stopwatch.GetTimestamp() call could effect the throughput.

Would be nice to somehow plug the PidCache cleanup looping task into the metrics system. If any production users would actually have an issue with that in terms of performance, it could be ideal to switch to a different data structure to cleanup that dictionary in an easier way.

src/Proto.Remote/RemoteProcess.cs
DenizPiri: The current solution looks good!

Hmm, I think DateTime.UtcNow doesn't allocate, as the DateTime is a struct. Date literally contains only a LONG member inside, no different than storing Ticks as long. https://source.dot.net/#System.Private.CoreLib/DateTime.cs,143

mhelleborg:
@DenizPiri 

I don't think we really need to care about last use of processes that are not remote.

The memory leak you found is caused by PartitionIdentity not clearing the PidCache when the cluster actor is terminated, I'll fix that instead :) I think the virtual actor middleware is the right place to solve it, as we have all the info needed. Just a matter of adding it to the stop handler.

src/Proto.Remote/RemoteProcess.cs
mhelleborg: You are correct, it does not allocate :)  It is however a little slower than the stopwatch, since it handles leap seconds etc. We don't need the absolute time here, just how long it was since it was used, so I think the stopwatch ticks is the cheapest way to handle that.

DenizPiri:
@mhelleborg 
I understand, so you are thinking of a solution like the CacheInvalidationExtension?

mhelleborg:
@DenizPiri A lot simpler than the remote cache invalidation. Pushed a commit now, have a look.

DenizPiri:
@mhelleborg Oh, I see. Yes, this looks good. 👍

Later today, I will try my leaking sample with trunk Proto.Actor with this PR merged in.

DenizPiri:
@mhelleborg I ran my leak demo code (https://github.com/DenizPiri/ProtoActorClusterPIDCacheMemoryLeak) on your branch, and it works perfectly now! My demo code only focuses on local Actors, doesn't use a real cluster. But, either the RemoteProcess timeout mechanism you add and/or PidCacheInvalidation extension should take care of that too.

On the left is the same code against proto.actor 0.25.0, on the right remote-pidcache-timeouts branch:
image
🎉🎉🎉🎉🎉

I believe this change is safe to merge now. However, I still think it would be great to surround the call to RemoveIdleRemoteProcessesOlderThan with a timer metric. IHistogramMetric.Observe(Stopwatch) should be fine I guess.

rogeralsing:
Merging, we can add metrics separately

DenizPiri:
@rogeralsing Would it be possible to get 0.26.0 or 0.25.1 out with this fix merged-in? We would like to deploy this to production and would like to avoid doing custom builds.

PR #1318, opened 2021-12-14
System.Diagnostics.Metrics + OpenTelemetry metrics

marcinbudny:
## Description

Rewrite metrics implementation to use abstractions now available in .NET. Add extensions to publish metrics via OpenTelemetry  compatible exporters.

Introduced breaking changes:
* Ubiquitous.Metrics abstraction removed in favor of builtin metrics abstraction
* metric configuration on ActorSystem changed to `ActorSystemConfig.WithMetrics(bool recordMetrics = true)`
* some metrics changed names
* some metrics changed types (specifically `protoactor_actor_mailbox_length` is now a histogram, since gauge made no sense when we have a set of multiple different values - we can only trackt the distribution)

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/src/Proto.Actor/EventStream/DeadLetter.cs
+++ b/src/Proto.Actor/EventStream/DeadLetter.cs
@@ -34,7 +34,8 @@ namespace Proto
         public PID? Sender { get; }
         public MessageHeader Header { get; }
 
-        public override string ToString() => $"DeadLetterEvent: [ Pid: {Pid}, Message: {Message.GetType()}:{Message}, Sender: {Sender}, Headers: {Header} ]";
+        public override string ToString()
+            => $"DeadLetterEvent: [ Pid: {Pid}, Message: {Message.GetType()}:{Message}, Sender: {Sender}, Headers: {Header} ]";
     }
 
     public class DeadLetterProcess : Process
@@ -46,7 +47,12 @@ namespace Proto
         protected internal override void SendUserMessage(PID pid, object message)
         {
             if (System.Metrics.Enabled)
-                ActorMetrics.DeadletterCount.Add(1, new("id", System.Id), new("address", System.Address), new("messagetype", message.GetType().Name));
+            {
+                ActorMetrics.DeadletterCount.Add(1,
+                    new("id", System.Id), new("address", System.Address),
+                    new("messagetype", MessageEnvelope.UnwrapMessage(message)?.GetType().Name ?? "<null>")
+                );
+            }
 
             var (msg, sender, header) = MessageEnvelope.Unwrap(message);
             System.EventStream.Publish(new DeadLetterEvent(pid, msg, sender, header));
--- a/src/Proto.Actor/EventStream/DeadLetter.cs
+++ b/src/Proto.Actor/EventStream/DeadLetter.cs
@@ -45,7 +45,9 @@ namespace Proto
 
         protected internal override void SendUserMessage(PID pid, object message)
         {
-            System.Metrics.Get<ActorMetrics>().DeadletterCount.Inc(new[] {System.Id, System.Address, message.GetType().Name});
+            if (!System.Metrics.IsNoop)
+                System.Metrics.Get<ActorMetrics>().DeadletterCount.Add(1, new("id", System.Id), new("address", System.Address), new("messagetype", message.GetType().Name));
+
             var (msg, sender, header) = MessageEnvelope.Unwrap(message);
             System.EventStream.Publish(new DeadLetterEvent(pid, msg, sender, header));
             if (sender is null) return;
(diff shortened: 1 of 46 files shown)


src/Proto.Actor/EventStream/DeadLetter.cs
```diff
@@ -45,7 +45,9 @@ public DeadLetterProcess(ActorSystem system) : base(system)
 
         protected internal override void SendUserMessage(PID pid, object message)
         {
-            System.Metrics.Get<ActorMetrics>().DeadletterCount.Inc(new[] {System.Id, System.Address, message.GetType().Name});
+            if (!System.Metrics.IsNoop)
+                ActorMetrics.DeadletterCount.Add(1, new("id", System.Id), new("address", System.Address), new("messagetype", message.GetType().Name));
```
mhelleborg: This could end up with messagetype 'MessageEnvelope', which is probably not what we want :)

src/Proto.Actor/Metrics/Metrics.cs
```diff
@@ -1,42 +1,17 @@
...
-        public ICountMetric CreateCount(string name, string description, params LabelName[] labelNames)
-            => _metrics.CreateCount(name, description, labelNames);
-
-        public IGaugeMetric CreateGauge(string name, string description, params LabelName[] labelNames)
-            => _metrics.CreateGauge(name, description, labelNames);
+        public readonly bool IsNoop;
 
-        public IHistogramMetric CreateHistogram(string name, string description, params LabelName[] labelNames)
-            => _metrics.CreateHistogram(name, description, labelNames);
+        public ProtoMetrics(bool recordMetrics) => IsNoop = !recordMetrics;
```
mhelleborg: .Enabled? It is almost exclusively used with !, that could perhaps read better if it was not defined as a negative

alexeyzimarev: I'd say more, if there's no listener for the metric, it won't do anything. So, I don't really think checking if metrics are enabled or not is overkill as the `Meter` thing does it anyway. Might be useful to measure it anyway before making a decision to skip this check or not.

alexeyzimarev: Ok, it's not _exactly_ like that, but metrics report to their subscribers, so when there are no subscribers, there's still some overhead, but it's minimal. The question is if we encourage people to measure their system, checking if metrics are enabled is pointless.

marcinbudny: I kept the IsNoop flag checking from original implementation assuming this is a micro-optimization. If someone is not interested in metrics, they don't have to pay the price of allocations and method calls related to that. Bear in mind that many of the metrics are on the hot path, e.g. measuring the actor's Receive handler. There are many places in proto.actor code that do micro-optimizations, like [here](https://github.com/asynkron/protoactor-dotnet/blob/620e61ddc739b6d34ddb127481169c9d76437d28/src/Proto.Actor/Context/ActorContext.cs#L432)

I can change the name to Enabled though.

src/Proto.Actor/EventStream/DeadLetter.cs
marcinbudny: Added message unwrapping in user message handler. AFAIK system messages will not be wrapped?

src/Proto.Actor/EventStream/DeadLetter.cs
```diff
@@ -46,7 +47,12 @@ public DeadLetterProcess(ActorSystem system) : base(system)
         protected internal override void SendUserMessage(PID pid, object message)
         {
             if (System.Metrics.Enabled)
-                ActorMetrics.DeadletterCount.Add(1, new("id", System.Id), new("address", System.Address), new("messagetype", message.GetType().Name));
+            {
+                ActorMetrics.DeadletterCount.Add(1,
+                    new("id", System.Id), new("address", System.Address),
+                    new("messagetype", MessageEnvelope.UnwrapMessage(message)?.GetType().Name ?? "<null>")
```
mhelleborg: Put it after `var (msg, sender, header) = MessageEnvelope.Unwrap(message);`, no need to unwrap it twice :)

src/Proto.Actor/EventStream/DeadLetter.cs
mhelleborg: System messages will not be wrapped, but that is handled by MessageEnvelope.Unwrap. If the message is not a MessageEnvelope, that will just return the exact same message as was passed in

src/Proto.Actor/EventStream/DeadLetter.cs
marcinbudny: fixed

src/Proto.Actor/EventStream/DeadLetter.cs
```diff
@@ -46,15 +46,16 @@ public DeadLetterProcess(ActorSystem system) : base(system)
...
         protected internal override void SendUserMessage(PID pid, object message)
         {
+            var (msg, sender, header) = MessageEnvelope.Unwrap(message);
+
             if (System.Metrics.Enabled)
             {
                 ActorMetrics.DeadletterCount.Add(1,
                     new("id", System.Id), new("address", System.Address),
-                    new("messagetype", MessageEnvelope.UnwrapMessage(message)?.GetType().Name ?? "<null>")
+                    new("messagetype", msg)
```
mhelleborg: This is the raw object, we should still get the type name ;)

marcinbudny: fixed

PR #1330, opened 2021-12-25
Gossip consensus abstractions

mhelleborg:
Adds the ability to register new consensus checks on demand.

This is useful when the cluster needs to agree on an event happening on all nodes, ex the topology is in consensus or all nodes are in a certain state.

Initial draft only implements full consensus since that is what is used internally.

### DIFF
--- a/src/Proto.Cluster/Gossip/Consensus.cs
+++ b/src/Proto.Cluster/Gossip/Consensus.cs
@@ -1,0 +1,78 @@
+﻿// -----------------------------------------------------------------------
+// <copyright file="ClusterConsensus.cs" company="Asynkron AB">
+//      Copyright (C) 2015-2021 Asynkron AB All rights reserved
+// </copyright>
+// -----------------------------------------------------------------------
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Proto.Cluster.Gossip
+{
+    public interface IConsensusHandle<T> : IDisposable
+    {
+        Task<(bool consensus, T value)> TryGetConsensus(CancellationToken ct);
+        Task<(bool consensus, T value)> TryGetConsensus(TimeSpan maxWait, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Clear any current consensus to prevent stale reads
+        /// </summary>
+        public void TryResetConsensus();
+    }
+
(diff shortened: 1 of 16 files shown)


src/Proto.Cluster/Gossip/Consensus.cs
```diff
@@ -0,0 +1,78 @@
...
+                    return (true, t.Result);
+            }
+
+            return (false, default);
+        }
+        
+        public async Task<(bool consensus, T value)> TryGetConsensus(TimeSpan maxWait, CancellationToken cancellationToken)
+        {
+            var t = _consensus.Task;
+            await Task.WhenAny(t, Task.Delay(maxWait, cancellationToken)).ConfigureAwait(false);
```
rogeralsing: We should probably introduce some abstraction for these WhenAny + Delay.
Because if the first task `t` is already completed, there will be delay tasks spawned that doesn't get cancelled.
_potentially_ in some bussy section of a system which spams TryGetConsensus, this could lead to a lot of tasks floating around.

Not really caused by this PR as we have had this a bit of everywhere.

src/Proto.Cluster/Gossip/Gossiper.cs
```diff
@@ -97,9 +127,148 @@ private async Task GossipLoop()
...
+                            return default;
+                        }
+
+                        var valueTuples = memberStates.Select(mapToValue);
+                        // ReSharper disable PossibleMultipleEnumeration
+                        var result = valueTuples.Select(it => it.value).HasConsensus();
+
+                        if (context.System.Config.DeveloperSupervisionLogging)
+                        {
+                            Logger.LogDebug("{SystemId}, consensus {Consensus}: {Values}", context.System.Id, result.Item1, valueTuples
```
rogeralsing, line 192: Check if debug logging is enabled. otherwise the groupby linq will always run

src/Proto.Cluster/Gossip/Consensus.cs
mhelleborg: We do have `Proto.Utils.TaskExtensions.WithTimeout`, but it does not quite solve the same problem. It also throws TimeoutException which is not what we want here.

How would the ideal abstraction look for something like this? We want to fall back to a default value if maxwait expires. But we also want to be able to cancel the operation if it is no longer relevant..

PR #1343, opened 2022-01-12
Threading fixes

DenizPiri:
## Description

This PR fixes many minor issues with threading.
Also fixes an issue with RoundRobinRouterState where it would fail beyond int.MaxValue messages.

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/src/Proto.Actor/Future/SharedFuture.cs
+++ b/src/Proto.Actor/Future/SharedFuture.cs
@@ -70,7 +70,15 @@ namespace Proto.Future
         private PID Pid { get; }
         public bool Stopping { get; private set; }
 
-        public int RequestsInFlight => (int) (_createdRequests - _completedRequests);
+        public int RequestsInFlight {
+            get {
+                // Read completedRequests first and createdRequests later so that we will
+                // never read the 2 vars in an order that would result in completedRequests > createdRequests.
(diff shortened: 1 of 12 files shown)


rogeralsing:
Many thanks @DenizPiri ! this is great!

src/Proto.Actor/Future/SharedFuture.cs
```diff
@@ -70,7 +70,15 @@ internal SharedFutureProcess(ActorSystem system, int size) : base(system)
         private PID Pid { get; }
         public bool Stopping { get; private set; }
 
-        public int RequestsInFlight => (int) (_createdRequests - _completedRequests);
+        public int RequestsInFlight {
+            get {
+                // Read completedRequests first and createdRequests later so that we will
+                // never read the 2 vars in an order that would result in completedRequests > createdRequests.
+                long completed = Interlocked.Read(ref _createdRequests);
```
swimtver, line 77: looks like here should be _completedRequests.

rogeralsing: Yes, you are correct. thanks for spotting it

DenizPiri: Seems like this is already handled by mhelleborg. Resolving this.

Issue #1316, opened 2021-12-13
Partition Identity Push Pull strategy


rogeralsing:
Mind notes after discussion with @mhelleborg 

a push-pull based strategy for partition identity could work as follows:

Activator actor sends placement information to owner node in batches. triggered by topology changes.
The activator computes the identity owner and starts batching information to it.
If there is a new topology change, the process starts over.
Once all activations have been sent. the process stops.

Identity actor consumes placement information sent from the activator actor. making sure itself and the activator both have the same topology hash. if not, the message is discarded.

That is the easy part.

The tricky part:

if a PidRequest comes into the identity owner actor, and the identity owner have not yet received the placement information from the activator actor.
Then the identity owner needs to query the cluster for this information.

this could potentially be done by broadcasting, via some gossip strategy, or via token-ring style passing a message through the entire ring of nodes.

the latter would be easy to implement, pass the request around one node to another, as long as the topology hash is the same.
if not. respond back to the original sender that the process needs to restart.

it all comes down to how do we in the best way locate the activation without too much latency and without too much chatter.

rogeralsing:
> Activator actor sends placement information to owner node in batches. triggered by topology changes.

This one might be more tricky than I thought.

say identity node is on topology v1 and activator node on topology v2

activator starts sending placements, identity receives them, but identity is on v1 and discards them.
then the placements are lost as the activator will not resend until a new topology.

PR #1355, opened 2022-01-16
PartitionIdentity handover rewrite

mhelleborg:
## Description

Ref #1316, this takes the idea to do push handovers and runs with it..

In addition to being able to push activations (thus ensuring the lowest latency and a single pass over the activations) I have added the ability to do delta handover for both pull and push. 

This takes a last known good  topology which we have previously rebalanced as a baseline. The handover then checks if the activation owner would have moved between the two topologies, and only sends the ones which have a new owner. 
If the owner is unchanged, it adds the fact that it was skipped to the last handover message, enabling the owner to check that it's local state matches 100% with the activating node.

If there is a mismatch detected (could be caused by terminations not received yet), the owner will do a full handover from that member. This should ensure consistency if there is any drift between members.

All of this is made possibly by "stopping the world". When a topology change is detected by a member, it will stop all new activations, and wait for all current activations to complete. After they have completed, this is sent as gossip state across the cluster. When a member has received from all other members the fact that there are no more in-flight activations, the topology rebalance is started. 

Even though there are no new activations while the rebalance happens, all existing activations will be returned by it's owner if present. This is also an improvement over previous iterations.

Network-wise these are breaking changes, and requires the full cluster to upgrade to the new version. All members also need to use the same mode for partition-rebalance.

## Purpose
This pull request is a:

## Checklist


tests/Proto.Cluster.PartitionIdentity.Tests/HandoverSinkTests.cs
```diff
@@ -0,0 +1,245 @@
...
+        Id = Guid.NewGuid().ToString(),
+        Port = 50001
+    };
+
+    [Fact]
+    public void CompletesOnEmptyHandover()
+    {
+        var topology = CreateTopology(TestMember1, TestMember2);
+        var receivedCount = 0;
+        var sink = new Partition.HandoverSink(topology, handover => receivedCount += handover.Actors.Count);
```
rogeralsing, line 41: Is this called concurrently? Interlocked increments might be needed

rogeralsing: Even if assign is atomic, reading then adding then assigning is not

mhelleborg: It's only used within actors, so no need for additional synchronization :)

PR #1361, opened 2022-01-20
Fix ability to ignore receive timeout with INotInfluenceReceiveTimeout

mhelleborg:
## Description

☝️

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/src/Proto.Actor/Context/ActorContext.cs
+++ b/src/Proto.Actor/Context/ActorContext.cs
@@ -341,10 +341,11 @@ namespace Proto.Context
                 System.DeadLetter.SendUserMessage(Self, msg);
                 return default;
             }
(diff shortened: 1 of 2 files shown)


src/Proto.Actor/Context/ActorContext.cs
```diff
@@ -341,13 +341,11 @@ private ValueTask InternalInvokeUserMessageAsync(object msg)
                 System.DeadLetter.SendUserMessage(Self, msg);
                 return default;
             }
+            var influenceReceiveTimeout = MessageEnvelope.UnwrapMessage(msg) is not INotInfluenceReceiveTimeout;
```
rogeralsing: this introduces an unwrap in a hot-path.
What are the effects on in process benchmark?

mhelleborg: That is a good point, will do a minor change to fix it

PR #1364, opened 2022-01-22
PoC abstract gossip mechanism

rogeralsing:


src/Proto.Cluster/Gossip/ConsensusChecks.cs
```diff
@@ -12,7 +12,7 @@ namespace Proto.Cluster.Gossip
 {
     
     
-    internal record ConsensusCheck(string Id, Action<GossipState, ImmutableHashSet<string>, IContext> Check, string[] AffectedKeys);
+    internal record ConsensusCheck(string Id, Action<GossipState, ImmutableHashSet<string>> Check, string[] AffectedKeys);
```
rogeralsing: Remove all actor stuff from consensus

src/Proto.Cluster/Gossip/GossipActor.cs
```diff
@@ -4,84 +4,57 @@
...
     {
-        private readonly TimeSpan _gossipRequestTimeout;
         private static readonly ILogger Logger = Log.CreateLogger<GossipActor>();
-        private long _localSequenceNo;
-        private GossipState _state = new();
-        private readonly Random _rnd = new();
-        private ImmutableDictionary<string, long> _committedOffsets = ImmutableDictionary<string, long>.Empty;
-        private ImmutableHashSet<string> _activeMemberIds = ImmutableHashSet<string>.Empty;
-        private Member[] _otherMembers = Array.Empty<Member>();
-        private readonly ConsensusChecks _consensusChecks = new();
```
rogeralsing, line 29: Remove all consensus state from the ConsensusActor

src/Proto.Cluster/Gossip/GossipActor.cs
```diff
@@ -4,84 +4,57 @@
...
+        private readonly TimeSpan _gossipRequestTimeout;
+        private readonly GossipFoo _foo;
```
rogeralsing: We need a name for this, the actual gossip implementation without actor code.
Gossip Foo is not a good name :D

src/Proto.Cluster/Gossip/GossipActor.cs
```diff
@@ -4,84 +4,57 @@
...
-            return Task.CompletedTask;
-        }
+        private Task OnClusterTopology(ClusterTopology clusterTopology) =>
+            _foo.OnClusterTopology(clusterTopology);
 
-        private Task OnAddConsensusCheck(IContext context, AddConsensusCheck msg)
-        {
-            _consensusChecks.Add(msg.Check);
+        private Task OnAddConsensusCheck(AddConsensusCheck msg) =>
+            _foo.OnAddConsensusCheck(msg);
```
rogeralsing: Delegate most calls to the internal consensus logic

PR #1368, opened 2022-01-24
Consensus refactoring. New ReenterAfterCancellation-abstraction.

mhelleborg:


src/Proto.Actor/Context/ContextExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Threading.Tasks;
+
+namespace Proto.Context
+{
+    public static class ContextExtensions
+    {
+        /// <summary>
+        /// Calls the callback on token cancellation. If CancellationToken is non-cancellable, this is a noop.
+        /// </summary>
+        public static CancellationTokenRegistration ReenterAfterCancellation(this IContext context, CancellationToken token, Action onCancelled)
```
mhelleborg: An alternative to having it as an extension is to add it directly to context. Then we could perhaps add cleanup for contexts with callbacks still waiting. With this model that is on the user.

tests/Proto.Cluster.Tests/GossipCoreTests.cs
```diff
@@ -66,8 +67,11 @@ void SendState(MemberStateDelta memberStateDelta, Member targetMember, InstanceL
             }
 
             var first = environment.Values.First().Gossip;
-            
-            var handle = RegisterConsensusCheck<ClusterTopology, ulong>("topology", tp => tp.TopologyHash, first);
+
+            var checkDefinition = Gossiper.ConsensusCheckBuilder<ulong>.Create("topology", (ClusterTopology tp) => tp.TopologyHash);
```
rogeralsing, line 71: Nice

rogeralsing:
Looks great!

mhelleborg:
Great, merging :)

mhelleborg:
Wrong button 😄

PR #1413, opened 2022-02-02
Declare known hosts parameter in seed node cluster provider

pblachut:
Seed node cluster provider doesn't call JoinSeed during starting member. In the examples, one of the node is selected to join.

PR introduces knownHost parameter in seed node cluster provider and calls join seed for them during startup.

Join seed method already checks if a given address is self address that is why it is not checked in cluster provider.


pblachut:
Closed due to fact that more discussions need to be involved here. PR in the current shape is not complete.

PR #1448, opened 2022-02-08
Apply Dotnetoss feedback

rogeralsing:


rogeralsing:
No real noticeable performance improvements, but lower levels of allocation

PR #1449, opened 2022-02-08
Display deserialization exceptions together with the error message.

DenizPiri:
## Description

Currently, if there is an error during deserialization, an error message is displayed, but the underlying exception is not shown. This change passes the exception to the logger.

## Purpose
This pull request is a:


rogeralsing:
👍🏻

PR #1455, opened 2022-02-10
Mark Proto.Opentracing and Proto.Remote.GrpcCore as obsolete

pblachut:
## Description
This PR marks Proto.Opentracing and Proto.Remote.GrpcCore as obsolete.


rogeralsing:
Looks good

PR #1458, opened 2022-02-12
Fix ActorSystemExtensions.Get for the cases of having many extensions.

DenizPiri:
## Description

In case anyone would ever have more than 10 extensions, this fixes a potential problem for that case.

## Purpose
This pull request is a:


rogeralsing:
Just woke up here so I might be missing something. but the array is resized together with the creation of the ID for any such plugin:

https://github.com/asynkron/protoactor-dotnet/blob/dev/src/Proto.Actor/Extensions/ActorSystemExtensions.cs#L42-L48

Are there cases where this is not enough?

DenizPiri:
That's right but that assumes that every extension will be registered. Which defies the point of Get function. In case the extension is not registered, id is increased beyond 10, but the array is not resized.

PR #1462, opened 2022-02-13
Don't execute older Continuations on older Actor instance after the actor is recreated.

DenizPiri:
## Description
Before this change, a `ReenterAfter(Task.Delay(5000), () => SaveInternalStateToDB());` line followed immediately by a `Restart` by the supervisor, that SaveInternalStateToDB() would execute on the previous Actor instance, as a result saving the old state into DB.

With this change each Continuation stores the Actor instance it was created for, if it is different when the Continuation message is received, it simply ignores it and logs this.

I would overall consider this a critical bug, given that, in production scenarios, it could cause issues that would be very very hard to debug.

## Purpose
This pull request is a:

## Checklist


rogeralsing:
Very nice catch!

PR #1482, opened 2022-02-21
Wait for partition rebalance retries

mhelleborg:
If for some reason ClusterTopology is not updated correctly the partition rebalance retries will currently be done in a hot loop. This adds a wait period of 200 ms before it tries again. Should prevent log flooding.

### DIFF
--- a/src/Proto.Cluster/Partition/PartitionIdentityRebalanceWorker.cs
+++ b/src/Proto.Cluster/Partition/PartitionIdentityRebalanceWorker.cs
@@ -19,6 +19,9 @@ namespace Proto.Cluster.Partition
     /// </summary>
     class PartitionIdentityRebalanceWorker : IActor, IDisposable
     {
+        private const string ReasonTimeout = "Timeout";
+        private const string ReasonDeadletter = "DeadLetter";
+
         private static readonly ILogger Logger = Log.CreateLogger<PartitionIdentityActor>();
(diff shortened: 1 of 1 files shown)


src/Proto.Cluster/Partition/PartitionIdentityRebalanceWorker.cs
```diff
@@ -94,7 +94,7 @@ private Task OnPartitionCompleted(PartitionCompleted response, IContext context)
         private Task OnPartitionFailed(PartitionFailed response, IContext context)
         {
             Logger.LogWarning("[PartitionIdentity] Retrying member {Member}, failed with {Reason}", response.MemberAddress, response.Reason);
-            StartRebalanceFromMember(_request!, context, response.MemberAddress);
+            context.ReenterAfter(Task.Delay(200, _cancellationToken), () => StartRebalanceFromMember(_request!, context, response.MemberAddress));
```
rogeralsing: Can we have some form of RequestAsync initialization here?
e.g. we send a request and if we get a proper answer back, we continue with the rebalance.
if we get timeout. do exponential backoff?

mhelleborg: Alternative suggestion. We already have the reason for the failure (deadletter or timeout). If it's a deadletter it makes sense to wait before a retry, but if it is a timeout it would make sense to retry instantly..

PR #1510, opened 2022-03-12
Fix PartitionActivator leaking memory in case of cluster actors being stopped

DenizPiri:
## Description

Before this fix, when PartitionActivator is in use and many cluster actors are created and destroyed (for example: because of timeout), PartitionActivator's `Dictionary<ClusterIdentity, PID> _actors` map were forever growing in size. Also the PidCache were not cleaned correctly. This fix solves these problems.

I verified the fix to be working by making a sample app (which manages to increase memory usage by 1GB/s, all while having no alive virtual actors). PartitionIdentity codes do not suffer from the same issue at all.

## Purpose
This pull request is a:


DenizPiri:
Added another fix comment for cleaning up PidCache upon cluster topology changes. Because of the way PartitionActivator works, we can clean PidCache correctly, whereas this is not possible to do in PartitionIdentity.

PR #1601, opened 2022-05-25
System props

rogeralsing:
PoC, allow configuring system actor props


src/Proto.Actor/Context/RootContext.cs
```diff
@@ -70,6 +70,25 @@ public PID SpawnNamed(Props props, string name)
             throw;
         }
     }
+    
+    public PID SpawnNamedSystem(Props props, string name)
```
rogeralsing: this should be moved somewhere else to not clutter RootContext.
Maybe an extension method

src/Proto.Actor/Configuration/ActorSystemConfig.cs
```diff
@@ -47,6 +48,15 @@ public record ActorSystemConfig
...
     /// </summary>
     public Func<Props, Props> ConfigureProps { get; init; } = props => props;
+    
+    /// <summary>
+    ///     Allows ActorSystem-wide augmentation of system Props
+    ///     All system props are translated via this function
+    /// </summary>
+    public Func<string, Props, Props> ConfigureSystemProps { get; init; } = (_,props) => props
+        .WithDeadlineDecorator(TimeSpan.FromSeconds(1))
+        .WithLoggingContextDecorator()
```
mhelleborg: If we should use a logging decorator by default, should we reduce the log level for messages to trace?
This could quickly become very noisy

src/Proto.Actor/Context/RootContext.cs
```diff
@@ -70,6 +70,25 @@ public PID SpawnNamed(Props props, string name)
...
+                ? System.Guardians.GetGuardianPid(props.GuardianStrategy)
+                : null;
+            
+            //augment props with system actor specific settings
+            props = System.ConfigureSystemProps(name, props);
+            return props.Spawn(System, name, parent);
+        }
+        catch (Exception x)
+        {
+            Logger.LogError(x, "RootContext Failed to spawn child actor {Name}", name);
```
mhelleborg: Throw wrapped in inner exception instead?

It's also spawned in root context, not a child actor

src/Proto.Actor/Context/RootContext.cs
mhelleborg: `SpawnNamedSystemActor` ?

src/Proto.Actor/Configuration/ActorSystemConfig.cs
rogeralsing: Done, I've set the loglevel to None. but using Information for system/infra messages, and Error for anything that throws

PR #1609, opened 2022-05-26
PubSub API changes

marcinbudny:
## Description

This is the first part of PubSub API adjustments before v1.0. 
Following changes are included:
* `PubSubManager` is now an actor system extension called `PubSubExtension`
* Introduced `IPublisher` abstraction that handles publishing batches. Current implementation sends them directly to `TopicActor`
* `Producer` renamed to `BatchingProducer`. It takes `IPublisher` as dependency.
* `BatchingProducer` can now be stopped by disposing it
* `BatchingProducer` will now correctly set results on delivery reports in case of a publishing error (and it will stop in that case)
* `PubSubBatchSize` removed from cluster config - it is now passed to the constructor of `BatchingProducer`
* `BatchingProducer` now has ability to limit the size of internal queue to prevent unbounded queue growth
* enable direct publishing a message or message batch to a topic, without the heavyweight `BatchingProducer` (`Publish` and `PublishBatch` extensions on `IPublisher`).
* `BatchingProducer.PublishBatch` made private
* Topic Kind is automatically registered with noop KV storage, unless user provides own registration
* Topic Kind is now `$topic` and delivery kind is `$pubsub-delivery` to avoid conflicts with user's kinds (**warning:** this prevents rolling updates on existing applications using pub sub)
* lower logging level to Debug
* added PubSub tests
* changed ClusterPubSub example to be a more straightforward use case
* fixed unsubscribe extensions on `Cluster` that did not work

Out of scope of this PR:
* retries in the `BatchingProducer`
* ack mode for message publishing, e.g. "ack when delivered to topic" vs "ack when delivered to subscribers"
* facilities to share `BatchingProducer` instances between multiple publishers (e.g. actors)

## Purpose
This pull request is a:

## Checklist


src/Proto.Cluster/DefaultClusterContext.cs
```diff
@@ -202,7 +202,6 @@ IFuture future
 
             if (!context.System.Shutdown.IsCancellationRequested)
                 if (Logger.IsEnabled(LogLevel.Debug)) Logger.LogDebug("TryRequestAsync timed out, PID from {Source}", source);
-            _pidCache.RemoveByVal(clusterIdentity, pid);
```
rogeralsing, line 205: What is the effect of removing this? looks like we could get stuck with an invalid pid in the pid cache, or is it removed elsewhere now?

marcinbudny: This change was commited by mistake, reverting.

PR #1611, opened 2022-05-26
ActorLoggingContext shouldn't log Terminated and Touch messages.

DenizPiri:
## Description

Without this change, we get a lot of spam logs of Terminated messages from PartitionActivatorActor (the same would apply to PartitionIdentityActor too).

The reason for ignoring Touch is that we use that to do health checks on certain internal Proto.Actor's actors. That also causes log spam.

## Purpose
This pull request is a:


DenizPiri:
In actual deployments, we solved this temporarily by setting the min log level of SystemActors log source to Warning. However, that is not an ideal solution, as it fully disables this feature.

Issue #1604, opened 2022-05-25
Spawn worker actors as system actors


rogeralsing:
e.g. Kubernetes monitor actor, should be spawned as a system actor
SystemContext.SpawnNamedSystem 

This ensures the correct logging and supervision is applied

rogeralsing:
Done

Issue #1603, opened 2022-05-25
Make system actor names consistent


rogeralsing:
Some are prefixed with $, some other are not.
Let's stick with the $ prefix and use for all system actors

$invalidator 
activator 
partition-activator 
$proxy-activator 
pubsub-delivery 
partition-activator 
$proxy-activator 
gossip 
pubsub-delivery 
server_seed

rogeralsing:
Done

PR #1649, opened 2022-06-03
More system actors and more consistent system actor naming

woksin:
## Description

This PR intends to solve issues #1604 and #1603.

* kubernetes-cluster-monitorand and IdentityStorageWorker now spawned as system actors
* Add '$' prefix to system actor names   
## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/src/Proto.Cluster/Identity/IdentityStorageLookup.cs
+++ b/src/Proto.Cluster/Identity/IdentityStorageLookup.cs
@@ -6,7 +6,7 @@ namespace Proto.Cluster.Identity;
 public class IdentityStorageLookup : IIdentityLookup
 {
     private const string WorkerActorName = "$identity-storage-worker";
-    private const string PlacementActorName = "placement-activator";
+    private const string PlacementActorName = "$placement-activator";
     private bool _isClient;
     private string _memberId = string.Empty;
     private PID _placementActor = null!;
--- a/src/Proto.Cluster/Identity/IdentityStorageLookup.cs
+++ b/src/Proto.Cluster/Identity/IdentityStorageLookup.cs
@@ -5,6 +5,7 @@ namespace Proto.Cluster.Identity;
 
 public class IdentityStorageLookup : IIdentityLookup
 {
+    private const string WorkerActorName = "$identity-storage-worker";
     private const string PlacementActorName = "placement-activator";
     private bool _isClient;
     private string _memberId = string.Empty;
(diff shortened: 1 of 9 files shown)


src/Proto.Cluster/Identity/IdentityStorageLookup.cs
```diff
@@ -36,7 +37,7 @@ public async Task SetupAsync(Cluster cluster, string[] kinds, bool isClient)
         await Storage.Init();
 
         var workerProps = Props.FromProducer(() => new IdentityStorageWorker(this));
-        _worker = _system.Root.Spawn(workerProps);
+        _worker = _system.Root.SpawnNamedSystem(workerProps, WorkerActorName);
```
woksin, line 40: I'm not sure about whether this should be a system actor. As far as I understand this worker lives alongside the IdentityStorageLookup and stops when that is stopped hence in that case it makes sense for me that this is spawn as a system actor.

src/Proto.Remote/Endpoints/EndpointManager.cs
```diff
@@ -196,7 +196,7 @@ internal IEndpoint GetClientEndpoint(string systemId)
     private void SpawnActivator()
     {
         var props = Props.FromProducer(() => new Activator(_remoteConfig, _system));
-        ActivatorPid = _system.Root.SpawnNamedSystem(props, "activator");
+        ActivatorPid = _system.Root.SpawnNamedSystem(props, "$activator");
```
rogeralsing: I see that this change breaks the remoting tests as other places still rely on the old name.
The name could be turned into a const as for the other named actors above, e.g. $server_seed and $client_seed

woksin: I didn't get to run the tests on my machine as I don't have the 3.1 sdk 😃 I can fix that. Though it's not comforting knowing that there are such "couplings" in the system with dependencies on actor names. Have it been a discussion point to reduce that implicit coupling by either introducing an abstraction between or having something akin to a centralized "name registry" (don't think I like this 😀 ), or something completely different

woksin: [82187d7](https://github.com/asynkron/protoactor-dotnet/pull/1649/commits/82187d75f8c571cab0383940087e223bb2c9ace1) This should resolve it

rogeralsing:
Very nice!
Merging this

PR #1679, opened 2022-06-21
Identity lookup improvements + test fix

marcinbudny:
## Description

* Wait for the partition identity lookup to stop when calling `Cluster.ShutdownAsync()`
* Fix a problem with shutdown process hanging when using PartitionIdentityLookup with mode = Pull
* Fixes for the partition identity test and include it in CI

## Purpose
This pull request is a:

## Checklist


src/Proto.Cluster/Partition/PartitionManager.cs
```diff
@@ -72,15 +72,17 @@ public void Setup()
         }
     }
 
-    public void Shutdown()
+    public async Task Shutdown()
```
rogeralsing: ShutdownAsync for naming consistency

src/Proto.Cluster/PartitionActivator/PartitionActivatorManager.cs
```diff
@@ -67,12 +68,15 @@ public void Setup()
         }
     }
 
-    public void Shutdown()
+    public async Task Shutdown()
```
rogeralsing: same here

PR #1700, opened 2022-07-04
inline everything in cluster context

rogeralsing:


src/Proto.Cluster/DefaultClusterContext.cs
```diff
@@ -218,26 +233,32 @@ private async ValueTask RemoveFromSource(ClusterIdentity clusterIdentity, PidSou
         }
     }
 
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
     private static (ResponseStatus Ok, T?) ToResult<T>(PidSource source, ISenderContext context, object result)
```
marcinbudny, line 237: Not used anymore and can be removed.

PR #1713, opened 2022-07-13
Change Pub-Sub guarantees to at-most-once

marcinbudny:
## Description

After testing the pub-sub functionality on production it became clear that the aim for at-least-once delivery is not very well aligned with soft real-time, in-memory processing approach promoted by proto.actor. Additionally, ensuring at-least-once delivery has proven to increase complexity of the code and configuration significantly. This PR switches to at-most-once guarantee to align better with promoted use cases for the pub-sub.

Reliable stream processing is still relevant use case, and we may introduce a dedicated functionality that supports it in the future. For now we suggest using external infrastructure like Kafka or RabbitMQ in conjunction with proto.actor to ensure at-least-once processing.

## Purpose
This pull request is a:

## Checklist


src/Proto.Cluster/DefaultClusterContext.cs
```diff
@@ -21,7 +21,7 @@ public class DefaultClusterContext : IClusterContext
 
     private readonly PidCache _pidCache;
     private readonly ShouldThrottle _requestLogThrottle;
-    private readonly TaskClock _clock;
+    private readonly TaskClock? _clock;
```
mhelleborg: This is not used for dotnet 6 and greater? Perhaps just add a conditional block instead of making it nullable?

src/Proto.Cluster/PubSub/BatchingProducer.cs
```diff
@@ -68,7 +67,7 @@ private async Task PublisherLoop(CancellationToken cancel)
                     {
                         if (!produceMessage.Cancel.IsCancellationRequested)
                         {
-                            batch.Envelopes.Add(produceMessage.Message);
+                            batch.Batch.Envelopes.Add(produceMessage.Message);
```
mhelleborg: Is the top level one a batch or a batchbuilder? Might look at naming here

src/Proto.Cluster/DefaultClusterContext.cs
marcinbudny: Done

src/Proto.Cluster/PubSub/BatchingProducer.cs
marcinbudny: Changed name to `batchWrapper`

PR #1717, opened 2022-07-15
CI improvements, target framework cleanup

marcinbudny:
## Description

This PR aims to improve developer experience with the CI testing. 
* split test into slow and fast path, where tests in fast path are not run for each target framework in parallel because it is faster to run the side by side, skipping the setup overhead
* remove net5.0 target from tests and builds
* all examples target only net6.0
* remove mongo container setup

Other approaches evaluated:
* caching restored nuget packages (the cache provides almost no improvement while introducing additional complexity)
* building once, then reusing build results for testing (it takes a long time to upload and download build artifacts)

## Purpose
This pull request is a:

## Checklist


mhelleborg (approved):
Looks great, and agree with the decision to drop 5.0 👍

PR #1727, opened 2022-07-26
Cluster request timeout now throws TimeoutException

### LINKED ISSUE #1712: Make Cluster.RequestAsync throw exception instead of returning null
Currently, Cluster.RequestAsync will return null for timeouts, this is an old design based on the fact that exceptions are expensive in .NET. it is however non-obvious and problematic to communicate to users.

Instead, I think we should use proper exceptions here.
This could also have a setting in ClusterConfig, so that it can be set to the old behavior for compatibility.

marcinbudny:
## Description

Closes #1712 

Cluster requests now throw TimeoutException on timeout. This is a breaking change. Old behavior can be enabled via `ClusterConfig.WithLegacyRequestTimeoutBehavior()`

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/benchmarks/ClusterMicroBenchmarks/ConcurrentSpawnBenchmark.cs
+++ b/benchmarks/ClusterMicroBenchmarks/ConcurrentSpawnBenchmark.cs
@@ -96,7 +96,19 @@ public class ConcurrentSpawnBenchmark
             tasks[i] = _cluster.RequestAsync<Terminated>(id, PoisonPill.Instance, cts.Token);
         }
 
-        Task.WhenAll(tasks).GetAwaiter().GetResult();
+        try
+        {
+            Task.WhenAll(tasks).GetAwaiter().GetResult();
+        }
+        catch (TimeoutException)
+        {
+            // ignore
+        }
+        catch (Exception e)
+        {
+            Console.WriteLine(e);
+            throw;
+        }
     }
 
     public enum IdentityLookup
--- a/src/Proto.Cluster/DefaultClusterContext.cs
+++ b/src/Proto.Cluster/DefaultClusterContext.cs
@@ -112,7 +112,7 @@ public class DefaultClusterContext : IClusterContext
                             return t1;
                         }
 
-                        if (untypedResult == null) // TODO: what if request response is actually null?
+                        if (untypedResult == null) // timeout, actual valid response cannot be null 
                         {
                             return TimeoutOrThrow();
                         }
--- a/src/Proto.Cluster/DefaultClusterContext.cs
+++ b/src/Proto.Cluster/DefaultClusterContext.cs
@@ -27,6 +27,7 @@ public class DefaultClusterContext : IClusterContext
     private readonly ActorSystem _system;
     private static readonly ILogger Logger = Log.CreateLogger<DefaultClusterContext>();
     private readonly int _requestTimeoutSeconds;
+    private readonly bool _legacyTimeouts;
 
     public DefaultClusterContext(Cluster cluster)
     {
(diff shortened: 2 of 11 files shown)


benchmarks/ClusterMicroBenchmarks/ConcurrentSpawnBenchmark.cs
```diff
@@ -96,7 +96,19 @@ public void RemoveActivations()
             tasks[i] = _cluster.RequestAsync<Terminated>(id, PoisonPill.Instance, cts.Token);
         }
 
-        Task.WhenAll(tasks).GetAwaiter().GetResult();
+        try
+        {
+            Task.WhenAll(tasks).GetAwaiter().GetResult();
```
mhelleborg, line 101: This will be a change in behaviour.. but not sure it's a problem in this context

tests/Proto.Cluster.Tests/ClusterTests.cs
```diff
@@ -57,6 +57,14 @@ public async Task HandlesSlowResponsesCorrectly()
         response.Message.Should().Be(msg);
     }
 
+    [Fact]
+    public async Task ThrowsWhenTimeoutExceeded()
+        => await Members.First()
+            .Invoking(m => m.RequestAsync<Pong>(CreateIdentity("slow-test"), EchoActor.Kind,
+                    new SlowPing {Message = "hi", DelayMs = 5000}, new CancellationTokenSource(2000).Token
```
mhelleborg: The timeout here might be even lower, to provide for a faster test?

tests/Proto.Cluster.Tests/LegacyTimeoutTests.cs
```diff
@@ -0,0 +1,43 @@
...
+public class LegacyTimeoutTests
+{
+    [Fact]
+    public async Task ReturnsNullOnTimeoutInLegacyMode()
+    {
+        await using var fixture = new Fixture(1);
+        await fixture.InitializeAsync();
+
+        var response = await fixture.Members.First().RequestAsync<Pong>(CreateIdentity("slow-test"), EchoActor.Kind,
+            new SlowPing {Message = "hi", DelayMs = 5000}, new CancellationTokenSource(2000).Token
```
mhelleborg: Shorter timeout?

src/Proto.Cluster/DefaultClusterContext.cs
```diff
@@ -110,15 +112,14 @@ public DefaultClusterContext(Cluster cluster)
                             return t1;
                         }
 
-                        if (typeof(T) == typeof(MessageEnvelope))
+                        if (untypedResult == null) // TODO: what if request response is actually null?
```
rogeralsing: It cannot be. null is not an allowed message

benchmarks/ClusterMicroBenchmarks/ConcurrentSpawnBenchmark.cs
marcinbudny: Not sure what you mean. The previous version ignored timeouts and the catch I added does is meant to reproduce this behavior with the new version. 

Overall the behavior of Benchmark.NET is a bit weird when exception is thrown in the cleanup method. It seems to ignore it but then benchmark  crashes on super weird issues.

mhelleborg: The block ignores it, but it does not wait for the remaining tasks? Previously each call "suppressed" the timeout, now it should throw on the first failing call?

mhelleborg: Anyways, it's a benchmark, not critical path :)

Not too concerned if the semantics changed a little there

marcinbudny: Task.WhenAll waits for all tasks even if one of the fails. Am I missing something here?

tests/Proto.Cluster.Tests/ClusterTests.cs
marcinbudny: Moved to separate class and experimented with timeout length. It seems that for netcoreapp3.1 target (where we use TaskClock) 4 seconds is a min for this no to fail in CI.

tests/Proto.Cluster.Tests/LegacyTimeoutTests.cs
marcinbudny: It seems that for netcoreapp3.1 target (where we use TaskClock) 4 seconds is a min for this no to fail in CI.

benchmarks/ClusterMicroBenchmarks/ConcurrentSpawnBenchmark.cs
marcinbudny: Ping @mhelleborg 
Should I just resolve this?

PR #1734, opened 2022-08-09
Prefixing code-generated actor's Kind property with proto package

elnur-m:
## Description

I've prefixed code-generated actor's Kind property with the .proto file's package name. The purpose behind this is to be able to create grains with the same name in different C# namespaces. One example would be creating an Account grain in App.Security namespace and another Account grain in App.Finance namespace.

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/tests/Proto.Cluster.CodeGen.Tests/ProtoGrainGenerationTests.cs
+++ b/tests/Proto.Cluster.CodeGen.Tests/ProtoGrainGenerationTests.cs
@@ -18,28 +18,8 @@ public class ProtoGrainGenerationTests
 
     [Theory]
     [InlineData("foo.proto", "ExpectedOutput.cs")]
-    public void CanGenerateGrains(string protoDefinitionFile, string expectedOutputFile)
-    {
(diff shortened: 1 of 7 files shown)


tests/Proto.Cluster.CodeGen.Tests/ProtoGrainGenerationTests.cs
```diff
@@ -37,6 +37,27 @@ public void CanGenerateGrains(string protoDefinitionFile, string expectedOutputF
         Assert.Equal(expectedOutput, res.Single().Text.Trim());
     }
         
+    [Theory]
+    [InlineData("foo_packageless.proto", "ExpectedOutputPackageless.cs")]
```
mhelleborg, line 21: You could just move this line to the first Theory, and skip the second test?

mhelleborg (approved):

rogeralsing:
IMO we should use protobuf package name rather than .net namespace for this for future compatibility with Go cluster.

PR #1750, opened 2022-09-06
Single node cluster provider / lookup

mhelleborg:
## Description

Added single node 'cluster' provider, allowing Proto.Cluster to be used in single applications with the cluster / virtual actor abstractions. 

This makes it easy to get started, and allows for the same programming model in a single service use case as in a clustered environment. It will not work with multiple members, but is meant to simplify local testing, and use cases where you only ever have a single instance of the application.

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/src/Proto.Cluster/SingleNode/SingleNodeActivatorActor.cs
+++ b/src/Proto.Cluster/SingleNode/SingleNodeActivatorActor.cs
@@ -1,0 +1,215 @@
+﻿// -----------------------------------------------------------------------
+// <copyright file="SingleNodeActivatorActor.cs" company="Asynkron AB">
+//      Copyright (C) 2015-2022 Asynkron AB All rights reserved
+// </copyright>
+// -----------------------------------------------------------------------
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Proto.Cluster.SingleNode;
+
+class SingleNodeActivatorActor : IActor
+{
+    private static readonly ILogger Logger = Log.CreateLogger<SingleNodeActivatorActor>();
+
+
+    private readonly Cluster _cluster;
+    private readonly Dictionary<ClusterIdentity, PID> _actors = new();
+    private readonly HashSet<ClusterIdentity> _inFlightIdentityChecks = new();
+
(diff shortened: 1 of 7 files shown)


src/Proto.Cluster/SingleNode/SingleNodeActivatorActor.cs
```diff
@@ -0,0 +1,215 @@
...
+    private readonly Cluster _cluster;
+    private readonly Dictionary<ClusterIdentity, PID> _actors = new();
+    private readonly HashSet<ClusterIdentity> _inFlightIdentityChecks = new();
+
+    public SingleNodeActivatorActor(Cluster cluster) => _cluster = cluster;
+
+    private Task OnStarted(IContext context)
+    {
+        var self = context.Self;
+        _cluster.System.EventStream.Subscribe<ActivationTerminated>(e => _cluster.System.Root.Send(self, e));
```
marcinbudny: Could this be simplified by using the `Subscribe<TMsg>(ISenderContext context, params PID[] pids)` overload?

src/Proto.Cluster/SingleNode/SingleNodeActivatorActor.cs
```diff
@@ -0,0 +1,215 @@
...
+
+        if (!_actors.ContainsKey(msg.ClusterIdentity))
+            return Task.CompletedTask;
+
+        if (Logger.IsEnabled(LogLevel.Trace))
+            Logger.LogTrace("[SingleNode] Terminating {Pid}", msg.Pid);
+
+        _actors.Remove(msg.ClusterIdentity);
+
+        // Broadcast ActivationTerminated to all nodes so that PidCaches gets
```
marcinbudny, line 93: Perhaps change the comment to reflect the fact that there is a single node? Also just send the event to the EventStream directly, skipping the broadcast (to make the code easier to reason about).

src/Proto.Cluster/SingleNode/SingleNodeActivatorActor.cs
mhelleborg: Yes, fixed :)

src/Proto.Cluster/SingleNode/SingleNodeActivatorActor.cs
mhelleborg: Cleared up, thanks 😄

PR #1889, opened 2022-12-13
Implemented Proto cluster provider for Azure Container Apps

gurkanguran:
## Description

Implementation of Proto cluster provider for Azure Container Apps. It implements IClusterProvider and does the following things to fulfill the Gossip Protocol requirements. 

1. It does cluster monitoring by fetching the list of members and updating cluster topology. This is done by checking the tags on all the replicas of all revisions which have the following criteria:

- Revision should be active.
- Revision should have more than 0 traffic weight.

2. It registers the running replica as a member only if the revision of the replica has the criteria defined above.

3. It deregisters the member when the running replica is shut down.

I have not tested it yet, so please do not merge until I do proper testing. 
Please review it in the meantime and let me know if I did not implement the Proto Cluster provider in the correct way. 

## Purpose
This pull request is a:

## Checklist

- [ ] I have added the necessary documentation (if appropriate)


gurkanguran:
@rogeralsing Could you review it when you have time? Please do not merge until I properly test it.

rogeralsing:
This looks great!

Some small tweaks needed before merging:
* The project needs to reference the Proto.Cluster project instead of the Proto.Cluster Nuget.
* The project also needs to use the same settings / props as other projects so that it becomes part of the Nuget build/versioning pipeline.

rogeralsing:
Let me know if it needs more testing.
I think it all looks good now. we only need the CLA signed before merge

gurkanguran:
Hi @rogeralsing,

I found some issues during testing and fixed all of them except one which is registering members with the right host IP address. I could not find a way to access the IP address of the replica which is needed when we update cluster topology with the members with host address and port during cluster monitoring.

I described the issue here as well: https://github.com/microsoft/azure-container-apps/issues/531

Please let me know if you have any ideas. Thanks!

rogeralsing:
Is there any information to extract here ?
https://learn.microsoft.com/en-us/dotnet/orleans/deployment/deploy-to-azure-container-apps

gurkanguran:
> Is there any information to extract here ? https://learn.microsoft.com/en-us/dotnet/orleans/deployment/deploy-to-azure-container-apps

I checked how Orleans implements this as they have a similar membership protocol to Proto Cluster. The IP address of the silo can be passed via config by the client. However, this is not the case in their example, they don't pass the IP address but only the port: https://github.com/Azure-Samples/Orleans-Cluster-on-Azure-Container-Apps/blob/main/Silo/Program.cs

If it's not passed during the service registration of Orleans, then the library decides the IP address itself by analyzing all network interfaces on the node and selecting the smallest one as you can see here: 
https://github.com/dotnet/orleans/blob/main/src/Orleans.Core/Configuration/ConfigUtilities.cs

gurkanguran:
@rogeralsing Fixed all the issues and completed testing.

It works fine now, distribution of the actors throughout members is established with the connection of members with correct IP addresses. Orleans cluster implementation helped with finding the correct IP address.

I also needed to update how the tags are managed. As replicas are not azure resources, they cannot have any tags on them. So I had to save all the tags on the container app resource level.

Can you please review it again? It is ready to merge now.

Issue #1651, opened 2022-06-05
Use source generated LoggerMessage


woksin:
Is it time to start gradually moving over to using [Microsoft's source generated LoggerMessage](https://andrewlock.net/exploring-dotnet-6-part-8-improving-logging-performance-with-source-generators/)? With the source generated logger message you get multiple things for free. Better performance and less allocations, no need to check if log level is enabled and hiding the actual log messages behind method-names (i.e LogThatXFailed(error)). 

Technically I don't think that there is a problem introducing given the target frameworks of these projects.

woksin:
This is obviously not something that needs to be done in one big-bang. In my experience it is smart to just decide on starting somewhere and sticking to it, then it's also easier to continue using it for new log-statements.

In terms of performance the best place to start would be where logging happens frequently, if you have a suggestion I would be happy to start this journey

alexeyzimarev:
What I did in Eventuous is to use the EventSource instead, and add an extension for the application builder, which adds a listener that dumps messages produced by the EventSource to `ILogger`. EventSource is very fast, although limited in functionality, and, if allows a more structured approach for logging like 
```csharp
            catch (Exception e) {
                Log.UnableToLoadAggregate<T>(streamName, e);
                throw;
            }
```

I see that the `ILogger` source generator uses the same principle by defining a "DSL" for logging and use attributes for the message template and level. With EventSource it looks like this

```csharp
    [Event(TypeMapRegisteredId, Message = "Type {0} registered as {1}", Level = EventLevel.Verbose)]
    public void TypeMapRegistered(string type, string typeName) {
        if (IsEnabled(EventLevel.Verbose, EventKeywords.All)) WriteEvent(TypeMapRegisteredId, type, typeName);
    }
```

rogeralsing:
I'm all for giving the new source-generated logging a try.
It would be interesting to see it in an isolated case first. just to get some idea of the look and feel of it

rogeralsing:
Closing due to inactivity. Reopen if still relevant.

PR #1907, opened 2023-01-19
Use LoggerMessage for logging in ActorContext

woksin:
## Description

Start introducing LoggerMessage slowly by using it in ActorContext. Refering to this issue #1651 where we talked about introducing it and to see how it would look. I use it in some projects and IMO I think it provides good readability, seperation of concerns at the same time of improving performance :)

## Purpose
This pull request is a:

## Checklist


woksin:
Weird with these failing tests as they should not really be affected by these changes :)

PR #2105, opened 2024-03-14
Fix actor fails to recive batching messages if batchsize is less than…

wuuer:
Fix actor fails to recive batching messages if batchsize is less than the number of messages in the batchingMailbox because of dead loop in the RunAsync method

## Purpose
This pull request is a:

- [✔ ] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [ ✔] I have added necessary documentation (if appropriate)


rogeralsing:
Nice catch!

PR #2107, opened 2024-03-18
Added support for propagating OpenTelemetry baggage

mhelleborg:
## Description

Adds support for propagating Baggage via OTEL, by populating Baggage.Current if received in PropagationContext.  Added tests to cover baggage propagation

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/src/Proto.OpenTelemetry/OpenTelemetryDecorators.cs
+++ b/src/Proto.OpenTelemetry/OpenTelemetryDecorators.cs
@@ -3,6 +3,7 @@ using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
+using OpenTelemetry;
 using OpenTelemetry.Trace;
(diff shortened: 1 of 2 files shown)


src/Proto.OpenTelemetry/OpenTelemetryDecorators.cs
```diff
@@ -312,6 +313,11 @@ public static void Send(string source, PID target, object message, ActivitySetup
 
         var propagationContext = envelope.Header.ExtractPropagationContext();
 
+        if (propagationContext.Baggage.Count > 0)
+        {
+            Baggage.Current = propagationContext.Baggage;
```
rogeralsing, line 319: TLDR on this?
It extracts baggage via header values.
then set the Baggage.Current to an actual baggage.
and then create a new activity via BuildStartedActivity,
and BuildStartedActivity extracts data from Baggage.Current into the new activity?

PR #2146, opened 2025-01-23
After an actor is stopped, block continuations from running

benbenwilde:
## Description

If there are any pending continuations after an actor is stopped, a Continuation system message will still be sent when its wait completes, and the Continuation will still run even though the actor is in a fully stopped state. This can result in some unexpected behavior including data corruption, since the actor could be running elsewhere at the same time (e.g. if it was moved to another node).

This PR adds checks before sending and handling a Continuation so this can no longer happen. A test was added to confirm the issue and the fix.

## Purpose
This pull request is a:

## Checklist


rogeralsing:
Very nice addition! many thanks

PR #2190, opened 2025-08-03
Add GrpcNet WithTLS helper

rogeralsing:
## Summary
- add `WithTLS` extension to GrpcNetRemoteConfig
- update RemoteTLS examples to use `WithTLS`
- move `WithTLS` into existing `GrpcNetExtensions` class

## Testing
- `dotnet test` *(fails: command not found)*
- `wget https://packages.microsoft.com/config/ubuntu/24.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb` *(fails: Proxy tunneling failed)*

------
https://chatgpt.com/codex/tasks/task_e_688ef6741d9c8328a0cf924bf4ec231b

### DIFF
--- a/src/Proto.Remote/Proto.Remote.csproj
+++ b/src/Proto.Remote/Proto.Remote.csproj
@@ -16,6 +16,9 @@
         <PackageReference Include="System.Text.Json" />
     </ItemGroup>
     <ItemGroup>
+        <FrameworkReference Include="Microsoft.AspNetCore.App" />
+    </ItemGroup>
+    <ItemGroup>
         <ProjectReference Include="..\Proto.Actor\Proto.Actor.csproj" />
     </ItemGroup>
     <ItemGroup>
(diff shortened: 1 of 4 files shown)


src/Proto.Remote/Proto.Remote.csproj
```diff
@@ -15,6 +15,9 @@
         <PackageReference Include="Grpc.Tools" PrivateAssets="All" />
         <PackageReference Include="System.Text.Json" />
     </ItemGroup>
+    <ItemGroup>
+        <FrameworkReference Include="Microsoft.AspNetCore.App" />
```
rogeralsing, line 19: is this an issue? can we get around this dep?

danfma: I believe the primary concern is the direct reliance on ASP.NET Core, which could restrict the library's use in environments like Blazor. While we do depend on the Kestrel infrastructure to initiate gRPC, implementing a solid abstraction could allow us to support other environments without sacrificing performance. 

Do you agree?

PR #2410, opened 2025-09-30
Added optional ClusterIdentity parameter to IMemberStrategy.GetActivator

mhelleborg:
## Description

Added the ability for IMemberStrategy to dispatch on ClusterIdentity as well as the caller address.

This allows for more custom strategies that take the identity into account, but is not used in any of the built-in strategies.

## Purpose
This pull request is a:

### DIFF
--- a/src/Proto.Cluster/Partition/PartitionIdentityActor.cs
+++ b/src/Proto.Cluster/Partition/PartitionIdentityActor.cs
@@ -580,7 +580,7 @@ internal class PartitionIdentityActor : IActor
         }
 
         //Get activator
-        var activatorAddress = _cluster.MemberList.GetActivator(msg.Kind, context.Sender!.Address, msg.ClusterIdentity)?.Address;
+        var activatorAddress = _cluster.MemberList.GetActivator(msg.ClusterIdentity, context.Sender!.Address)?.Address;
 
         if (string.IsNullOrEmpty(activatorAddress))
         {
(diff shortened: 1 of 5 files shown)


src/Proto.Cluster/Partition/PartitionIdentityActor.cs
```diff
@@ -687,7 +687,7 @@ TaskCompletionSource<ActivationResponse> setResponse
                             }
 
                             var currentActivatorAddress =
-                                _cluster.MemberList.GetActivator(msg.Kind, context.Sender!.Address)?.Address;
+                                _cluster.MemberList.GetActivator(msg.Kind, context.Sender!.Address, msg.ClusterIdentity)?.Address;
```
rogeralsing: My only nitpick is that we now get "kind" twice.
once as a string, and once in the cluster identity.

We always have the full identity in all cases, right?
I´m thinking if we could just. tweak the GetActivator with an overload that just take address and identity (?)

and Obsolete warn on the other one.
I can´t imagine there are many custom member strategies out there in the wild anyway.

rogeralsing (commented):
LGTM. see nitpick comment.

src/Proto.Cluster/Partition/PartitionIdentityActor.cs
mhelleborg: Good points, cleaning up

PR #2412, opened 2025-10-24
Add ability to use IDiagnosticsTypeName on IActor for better naming control

Prunkles:
## Description

Currently, actor type names in logs or traces are constructed based on the `IActor` implementation type name. So if someone uses only `FunctionalActor`s or some other single configurable type for all actors, then these logs became effectively not so useful. This PR adds an ability to configure this behavior by using `IDiagnosticsTypeName` interface on `IActor` implementations.

## Purpose
This pull request is a:

## Checklist


Prunkles:
Maybe we can use some other dedicated interface instead of `IDiagnosticsTypeName`, but I don't think it's worth it. Message types and actor types do not intersect in any way, so there won't be any confusion, I guess

